PR #1, opened 2015-12-15
Add Render-To-Texture to BL_Texture

DCubix:
Now we're able to render anything to a BL_Texture :smiley:


panzergame:
I see you implement a easy way to use the framebuffer with the image, but currently you didn't show any example of its usage. ImageRender doesn't do the same thing as your patch but in VideoTexture ? If yes i will avoid duplicating features.

panzergame:
I don't understand why you reverted e65d602 and why you mixed merge in the "branch".
For a proper review use a clean branch only rebased on the trunk not merged.

PR #48, opened 2016-03-04
UPBGE: New lod system.

panzergame:


source/gameengine/Converter/KX_BlenderSceneConverter.cpp
```diff
@@ -142,6 +142,11 @@ KX_BlenderSceneConverter::~KX_BlenderSceneConverter()
 		delete (adtList);
 	}
 
+	for (unsigned int i = 0; i < m_map_blenderobject_to_lodlevels.size(); i++) {
```
DCubix: can we create a typedef for unsigned types somewhere so it's shorter to type? Like this:
```c
typedef unsigned int uint;
typedef unsigned short ushort;
etc...
```

panzergame: I will avoid use uint or ushort because we can easily mixed up with short.

panzergame:
I think i'm wrong with names: KX_LodLevel is a bit redundant since Lod is Level of detail. I propose to rename everything:
KX_LodLevels --> KX_Lod
KX_LodLevel --> KX_Lod::Level

Here we can mixed up about the plurial. Also it's is needed to specify that is a mesh/object lod ?
@DCubix : What do you think ?

DCubix:
Yeah, it's just a mesh lod level

panzergame:
pushed in c0e2193b0ed2ed2ffa11f0f737cfa1c1ef87574a

PR #51, opened 2016-03-06
Properly support multiple monitors on Linux

pqftgs:
Re-adds xinerama code for multi-monitor support on Linux.  It was previously removed from Blender for having little use.  However, it's fairly important for full-screen games.

This makes full-screen windows render on one monitor, rather than halfway down the middle.


panzergame:
@pqftgs : you only revert a commit ? if yes can i see the original commit which removed xinerama ?
Also i don't understand how this affect windows screen if this is written in X11 code.

DCubix:
@panzergame It's linux-only. It was removed because the devs didn't want to include additional dependencies.

pqftgs:
@panzergame That's windows as in _full-screen windows in Linux_, and yeah it's reverting an old commit + tweaking CMakeLists.txt to match the current style

PR #55, opened 2016-03-08
UPBGE: new World python API

youle31:
All is working now

Test files: 
- Normal background Python API: http://www.pasteall.org/blend/40984
- Background animation IPO: http://www.pasteall.org/blend/40985
- ImageRender background Python API: http://www.pasteall.org/blend/40986


panzergame:
Commited in 603eef36a13c23c1362c5e272fdcf2035ffb4432

PR #56, opened 2016-03-12
Rasterizer Cleanup

DCubix:
The purpose of this branch is to move (almost) ALL the OpenGL calls to the Rasterizer to avoid calling them directly.


panzergame:
commited

PR #57, opened 2016-03-13
Expose steering actuator path to Python

pqftgs:
When using a navigation mesh, exposes the steering actuator's path (list of points) to Python.
Usage: cont.actuators['Steering'].path


pqftgs:
Closing this PR as the original branch was nuked.  I will open another.

PR #69, opened 2016-03-25
Ge object timer

### LINKED ISSUE #68: Added object time weirdness
The time argument for scene.addObject is advertised as frames, but then converted to "sort-of" seconds (hard coded at 50fps...) resulting in unexpected behavior.

Solution 1:
Make it count frames instead.  Easy & keeps old stuff compatible.

Solution 2:
Make it accept time (in seconds) as the argument, and decrement based on the time scale.  Superior behavior but breaks compatibility.

pqftgs:
Makes the add object timer count frames as advertised.  Before, it was converting frames to seconds at a hard-coded rate of 50fps, which sometimes resulted in unexpected behavior.
Resolves #68

### DIFF
--- a/source/gameengine/Expressions/EXP_IntValue.h
+++ b/source/gameengine/Expressions/EXP_IntValue.h
@@ -34,6 +34,7 @@ public:
 	virtual int				GetValueType();
 	
 	cInt GetInt();
+	void SetInt(int i);
 	CIntValue();
 	CIntValue(cInt innie);
 	CIntValue(cInt innie,
--- a/source/gameengine/Expressions/intern/IntValue.cpp
+++ b/source/gameengine/Expressions/intern/IntValue.cpp
@@ -282,6 +282,12 @@ CValue* CIntValue::CalcFinal(VALUE_DATA_TYPE dtype, VALUE_OPERATOR op, CValue *v
 }
 
 
+void CIntValue::SetInt(int i)
+{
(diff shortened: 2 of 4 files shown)


panzergame:
When i was saying to use int it was only for the user interface because internally float are the best : Why ? : if the user set a high framerate and that it run at the half, we should decrease time twice per frame, But in your patch it will decrease only per frame.

The goal of this patch is to say explicitly to the user : "you set a life of 10, so it will be 1/yourframerate * 10"

@pqftgs : To do this just replace the 1/50 by 1.0f / KX_GetActiveEngine()->GetTicRate() and in LogicBeginFrame pass the delta time from last frame.

pqftgs:
I disagree.  This behavior is a strange quirk unique to bge and potentially causes other problems.  If a user specifies "frames" - it should be frames.  For counting actual time the user would expect to supply time instead.

**The problem**
Expecting an object to live 5 frames, the user might have an update function intended to run 5 times.  When the framerate changes, it doesn't run the expected 5 times and causes strange behavior.  The same applies to the delay sensor.

panzergame:
frames are here for compatibility, but it mean for  frames a life of  5 * 1 / yourframerate.

pqftgs:
I still think it's bad and confusing, but will yield to your judgement and close this PR.  However, someone needs to update the docs, as I've already had a few bugs caused by this misconception.

PR #71, opened 2016-03-28
recalculating normals patch

Jay-Parker:
After moving a mesh's vertices with python (like when generating terrain) you have to recalculate the mesh's normals with python, which can be painfully slow. This adds recalculateNormals(material_ID) to KX_MeshProxy to speed things up.

### DIFF
--- a/source/gameengine/Ketsji/KX_MeshProxy.cpp
+++ b/source/gameengine/Ketsji/KX_MeshProxy.cpp
@@ -37,6 +37,7 @@
 #include "RAS_DisplayArray.h"
 #include "RAS_MeshObject.h"
 #include "RAS_Polygon.h"
+#include "BLI_math.h"
 
 #include "KX_VertexProxy.h"
 #include "KX_PolyProxy.h"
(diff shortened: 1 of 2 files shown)


source/gameengine/Ketsji/KX_MeshProxy.cpp
```diff
@@ -474,4 +476,79 @@ bool ConvertPythonToMesh(PyObject *value, RAS_MeshObject **object, bool py_none_
...
+			}
+
+			/* normalise summed normals */
+			for (int i = 0; i < m_meshobj->NumVertices(mat); i++)
+			{
+				vertex = m_meshobj->GetVertex(matindex, i);
+				MT_Vector3 normal(vertex->getNormal()[0], vertex->getNormal()[1], vertex->getNormal()[2]);
+				normal.normalize();
+				vertex->SetNormal(normal);
+				this->AppendModifiedFlag(RAS_MeshObject::NORMAL_MODIFIED);
```
panzergame: please avoid the use of this->

panzergame:
Did you test with smoothed vertexes ? Because here i think that the first processed polygon will overwrite the other. Also did it manage with quad RAS_Polygon ?

Jay-Parker:
Could you explain why to avoid the use of this-> please? Also, I wasn't sure about that line, other parts that change verts/normals seem to set that flag, but I couldn't seem to figure out what it does. Works fine without it.

Seemed fine with both solid and smooth verts in my tests. I'll double check.

panzergame:
"this->" is useless, "this->func()" and "func()" are the same from a class.

Jay-Parker:
Cheers. I've made some changes. it's a lot shorter now.

Not sure what you mean by getting one polygon overwriting another. Smooth vertices are pretty much spot on now, there's some slight differences between the normals blender initially creates and the recalculated ones. But since it's meant to be used on a mesh after moving verts around it's probably unnoticeable. 

The only real issue I couldn't figure out was a cylinder with an n-gon cap set to smooth produces odd results (appears fine with triangle fan, or with flat shading). My guess is that the n-gon gets triangulated into some horrible mess that throws the function off. Otherwise it works fine with smooth, solid, quads, triangles and any mix of them.

panzergame:
Please provide a example blend file.

Akira1San:
Dont use n-gon for in game.

panzergame:
@AkiraSan: we will find a solution before

Jay-Parker:
Yeah, most people are not going to be using n-gons in the game.

I'm going to close this pr and submit a new one since I've come up with a neater solution and I'll add a example .blend

panzergame:
@Jay-Parker : ok i stop review for the moment, also use function like normal_tri_v3, normal_quad_v3, or normal_poly_v3 from BLI_math_geom.h instead of rewrite the normal compute.

Jay-Parker:
Cheers. Yeah, that's what I've done, and I've also moved the normal calculation to RAS_Polygon to allow python access.

PR #72, opened 2016-03-28
Expose steering actuator path to Python with CListWrapper

pqftgs:
When using a navigation mesh, exposes the steering actuator's path (list of points) to Python.
Usage: cont.actuators['Steering'].path

Uses CListWrapper as previously requested

### DIFF
--- a/source/gameengine/Ketsji/KX_SteeringActuator.cpp
+++ b/source/gameengine/Ketsji/KX_SteeringActuator.cpp
@@ -37,6 +37,8 @@
 #include "KX_PyMath.h"
 #include "Recast.h"
 
+#include "EXP_ListWrapper.h"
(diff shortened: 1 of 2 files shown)


panzergame:
Th path is recalculated every frame, no ?

panzergame:
The patch is fine, but later you should make a patch to enable set a path:
> actuator.pathCustom = True
> actuator.path = [enemy0.worldPosition, enemy1.worldPosition]

pqftgs:
Yeah, the path is recalculated every frame unless specified in the actuator.  I will consider setting the path in a future patch.

panzergame:
Please avoid merge the patch but rebase it and add the commit prefix "UPBGE: "

PR #73, opened 2016-03-28
Recalculating normals 2

Jay-Parker:
As before, this allows you to recalculate the normals of a mesh after moving the vertices. 

The face normal is calculated in RAS_Polygon instead of KX_MeshProxy. So I was able to add python access to the face normal as well: KX_PolyProxy.normal. Useful if you want to align something with a poly's normal.

I'm still figuring out how you use git and this request seems to include other changes from when I merged upbge with my fork, sorry if this confuses things. (Here's the .diff: https://www.dropbox.com/s/c14182iefw06is6/normals3.diff?dl=0)

examples:
https://www.dropbox.com/s/xwfiam6rdnd2eqs/faceNormalsTest.blend?dl=0
https://www.dropbox.com/s/ytir967mn8tzx7h/recalcNormals2.blend?dl=0


source/gameengine/Ketsji/KX_MeshProxy.cpp
```diff
@@ -474,4 +476,56 @@ bool ConvertPythonToMesh(PyObject *value, RAS_MeshObject **object, bool py_none_
...
+			}
+
+			/* sum face normals for each connected vert*/
+			for (int p = 0; p < m_meshobj->NumPolygons(); p++)
+			{
+				polygon = m_meshobj->GetPolygon(p);
+
+				for (int v = 0; v < polygon->VertexCount(); v++)
+				{
+					MT_Vector3 vertNormal(polygon->GetVertex(v)->getNormal()[0], polygon->GetVertex(v)->getNormal()[1], polygon->GetVertex(v)->getNormal()[2]);
```
panzergame: Why reusing the vertex normal ? it should be written only, no ?

source/gameengine/Rasterizer/RAS_Polygon.cpp
```diff
@@ -120,3 +121,19 @@ RAS_DisplayArray *RAS_Polygon::GetDisplayArray()
...
+
+	for (int v = 0; v < m_numvert; v++)
+	{
+		verts[v][0] = GetVertex(v)->getXYZ()[0];
+		verts[v][1] = GetVertex(v)->getXYZ()[1];
+		verts[v][2] = GetVertex(v)->getXYZ()[2];
+	}
+
+	normal_poly_v3(faceNormal, verts, m_numvert);
+	return faceNormal;
```
panzergame: ohhhh, here you return a pointer to an array which will be freed just after the return.
Use MT_Vector to make a safe copy to return.

panzergame:
If i understand for smoothed vertexes you add all polygons normal to the shared vertex and after normalize the normal ?
It's fine for an entire mesh but smoothed vertex between two materials will fail.

Jay-Parker:
@panzergame Yeah, all the normals of adjacent faces to the vertex are added together then then normalised to produce the vertex. In the case of solid shading the vertices are split, so the normal totals will be different, so it works with both shading types. 

Not sure how to fix the multiple materials issue. It doesn't look like it's possible to get a vert from RAS_Polygon without a material index. If I'm understanding it rightly verts are stored by material not mesh? You can get a polygon without a material, and then get the verts from that. But that would mean verts get processed multiple times unnecessarily, perhaps a check to guard against that?

source/gameengine/Rasterizer/RAS_Polygon.cpp
Jay-Parker: I initially tried that, but it wouldn't compile. Fixed it now.

Jay-Parker:
Right, it should now work with any mix of materials and face shading types. There's still issues with n-gons set to smooth. I believe it stems from the n-gon being triangulated into an uneven mess, so each vertex has varying amounts to uneven faces that contribute to the normal.

Curiously, if you take out the code to filter out solid faces then smooth n-gons display correctly. Most other things will, but if you mix solid and smooth faces then it breaks. Generally I think it's better to have more combinations work and accept that n-gons should be avoided in the BGE.

panzergame:
Please make it working properly before ask a review.

Akira1San:
Game engine should only use triangles!

panzergame:
@Akira1San : We use only triangle as internal render side, but for the user it's sometimes better to use quad and tris.

Akira1San:
Yes thats normal for every game engine.

Jay-Parker:
Ok, I'll close this and get back to work on it. Any ideas for fixing the smooth n-gon problem?

PR #88, opened 2016-04-21
Fix screenshot crash

youle31:
since openGL refactor, crash when screenshot.
I don't understand why, but this patch seems to fix the issue.
task->dumprect, path, im_format freed in another function and save
screenshot_thread_func try to access a NULL pointer with MEM_freeN?

test file: http://www.pasteall.org/blend/41545

### DIFF
--- a/source/gameengine/Rasterizer/RAS_ICanvas.cpp
+++ b/source/gameengine/Rasterizer/RAS_ICanvas.cpp
@@ -92,14 +92,12 @@ void save_screenshot_thread_func(TaskPool *__restrict UNUSED(pool), void *taskda
 	ibuf->rect = task->dumprect;
 	
 	BKE_imbuf_write_as(ibuf, task->path, task->im_format, false);
+	MEM_freeN(task->path);
 	task->path = NULL;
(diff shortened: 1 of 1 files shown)


source/gameengine/Rasterizer/RAS_ICanvas.cpp
```diff
@@ -90,14 +90,16 @@ void save_screenshot_thread_func(TaskPool *__restrict UNUSED(pool), void *taskda
 	/* create and save imbuf */
 	ImBuf *ibuf = IMB_allocImBuf(task->dumpsx, task->dumpsy, 24, 0);
 	ibuf->rect = task->dumprect;
-
+	
 	BKE_imbuf_write_as(ibuf, task->path, task->im_format, false);
-
+	task->path = NULL;
```
panzergame: Indeed path need to be set to NULL. But as you can see later we alloc the name with
`	char *path = (char *)MEM_mallocN(FILE_MAX, "screenshot-path");`

So you have also to free the path:
`
MEM_freeN(task->path);
task->path = NULL;
`

PR #90, opened 2016-04-25
add lodbias for lamp textures

youle31:
I don't know If we make an API to change it in realtime. Seems
complex->texture proxy for lamps? Sounds not very usefull

test file: http://www.pasteall.org/blend/41582


panzergame:
Yeah, no need to refactor the API for lamp textures.

PR #96, opened 2016-05-02
Implementation of KX_Filter proxy to set uniforms

youle31:
This is a prototype:

- Problem to get the name of the filter (we have to find 2D filter
  actuator name). For the moment the name is "0" for all filters.
- setSampler doesn't work neither setUniform1i("tex", texBindcode)... we
  have to investigate on this. A texture can't be bound to the 2D filter
  for the moment.
- I just tested for setUniform1f but I guess it works for 2f, 1i etc...
- I copied all the methods from BL_Shader but this means we have all in
  double. It's not very esthetic. But the methods used in KX_2DFilter use
  glUseProgramARB to update uniforms so this methods are not exactly the
  same as in BL_Shader. But maybe a better way could be found.
- I didn't add setUniformDef for the moment, I don't know if it's needed...
- Neither setAttrib...
- I didn't add #ifdef WITH_PYTHON for the moment so to compile you have to select WITH_PYTHON_INSTALL
- We have to move bge.logic.filters() (KX_PythonInit.cpp) to KX_Scene.filters
- Maybe some cleanups could be done in #includes because of inheritance.

Needs tests and to solve some issues but I'm not experienced enough to
solve it alone...

Test file for setUniform1f: http://www.pasteall.org/blend/41667

### DIFF
--- a/doc/python_api/rst/bge_types/bge.types.BL_Shader.rst
+++ b/doc/python_api/rst/bge_types/bge.types.BL_Shader.rst
@@ -11,9 +11,9 @@ base class --- :class:`PyObjectPlus`
 
    TODO - Description
 
-   .. attribute:: enable
+   .. attribute:: enabled
 
-      Enable/Disable shader.
+      Set shader enabled to use.
 
       :type: boolean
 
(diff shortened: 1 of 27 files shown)


youle31:
So, for textures issue, this is because of DesactivateMaterial(rasty); in RAS_MaterialBucket::RenderMeshSlots.
Unbind the material at this moment of the rendering process makes all the textures unavailable:
- For PostRendering
- For all kind of bgl stuff with GL_TEXTURE_2D

I don't know well this part of the code but I suggest we differ/delay the desactivation/unbinding of GPU material.

PR #99, opened 2016-05-09
New Game Controller implementation. It follows Xbox360 controller

lordloki:
It follows Xbox360 controller standard. The joystick logic brick is adapted to it now.

A Game Controller is composed of:

6 axis availables (in order from 0 to 5):	
AXIS_LEFTSTICK_X, AXIS_LEFTSTICK_Y, AXIS_RIGHTSTICK_X, AXIS_RIGHTSTICK_Y, AXIS_TRIGGERLEFT and AXIS_TRIGGERRIGHT.

and 15 buttons availables (in order from 0 to 14):
BUTTON_A, BUTTON_B, BUTTON_X, BUTTON_Y, BUTTON_BACK, BUTTON_GUIDE, BUTTON_START, BUTTON_LEFTSTICK, BUTTON_RIGHTSTICK, BUTTON_LEFTSHOULDER, BUTTON_RIGHTSHOULDER, BUTTON_DPAD_UP, BUTTON_DPAD_DOWN, BUTTON_DPAD_LEFT and BUTTON_DPAD_RIGHT.


BluePrintRandom:
:+1:

source/blender/editors/space_logic/logic_window.c
```diff
@@ -985,41 +985,34 @@ static void draw_sensor_delay(uiLayout *layout, PointerRNA *ptr)
```
panzergame: Can you explain why do you need to change the UI of the jostick logic brick ?

source/blender/makesdna/DNA_sensor_types.h
```diff
@@ -189,8 +189,8 @@ typedef struct bJoystickSensor {
 	short axis_single;
 	int axisf;
 	int button;
-	int hat;
-	int hatf;
+	int hat; //unused
```
panzergame: Are they still used for compatibility conversion ? Else you can simply remove it.

source/gameengine/GameLogic/Joystick/SCA_Joystick.cpp
```diff
@@ -77,92 +75,74 @@ SCA_Joystick *SCA_Joystick::m_instance[JOYINDEX_MAX];
...
+		return;
+	}
 
-		/* The video subsystem is required for joystick input to work. However,
-		 * when GHOST is running under SDL, video is initialized elsewhere. We
-		 * also need to set the videodriver to dummy, and do it here to avoid
-		 * interfering with addons that may use SDL too.
-		 *
-		 * We also init SDL once only. */
-#  ifdef WITH_GHOST_SDL
```
panzergame: Did you investigated on this ?

source/gameengine/GameLogic/Joystick/SCA_Joystick.cpp
```diff
@@ -77,92 +75,74 @@ SCA_Joystick *SCA_Joystick::m_instance[JOYINDEX_MAX];
...
-#endif /* WITH_SDL */
+	if (m_instance[joyindex]) {
+		m_instance[joyindex]->DestroyJoystickDevice();
+		delete m_instance[joyindex];
+		m_instance[joyindex] = NULL;
 	}
+#endif /* WITH_SDL */
 }
 
 void SCA_Joystick::cSetPrecision(int val)
```
panzergame: Why the prefix "a" and "c" to these functions ?

youle31:
Many Thanks for this @lordloki !

source/blender/editors/space_logic/logic_window.c
lordloki: Basically, with the new system used by SDL2 every joystick/gamepad/etc is configured as a xbox360 controller. Then there is no necessity to expose old elements, i.e.:

- There is no hats, now the digital pads are treated as normal buttons.
- We only expect 15 buttons max
- etc

Then I modified the UI to show the buttons and axis as in the Xbox360 controller to facilitate the configuration.

The issue with that new SDL2 system is that if your joystick is not preconfigured you have to use Steam Big Mode, controllermapping in SDL2 test sources or antimicro application to generate a new valid mapping. But it seems that it is being very well accepted by the community.

source/blender/makesdna/DNA_sensor_types.h
lordloki: I will remove them.

source/gameengine/GameLogic/Joystick/SCA_Joystick.cpp
lordloki: In all net example about SDL2 joystick that i can find it was not necessity to enable SDL video subsystem.

source/gameengine/GameLogic/Joystick/SCA_Joystick.cpp
lordloki: i don't know either

lordloki:
you're welcome!! :-)

panzergame:
@lordloki : Apparently the patch doesn't compile without WITH_SDL, can you fix that ?

lordloki:
In last commits:
- Hat unused variables are removed now.
- build with WITH_SDL off is now fixed
- Partial redesign of Joystick sensor brick ui to include special axis for
shoulder triggers

panzergame:
@lordloki : I was a bit buzy these last days, but the review progress, also i was obligated to rebuild my kernel to support joystick (gentoo default options) but my only joystick is not working in both blender and UPBGE.

lsusb -v for the joystick gives:
`
Bus 003 Device 004: ID 0079:0006 DragonRise Inc. PC TWIN SHOCK Gamepad
Couldn't open device, some information will be missing
Device Descriptor:
  bLength                18
  bDescriptorType         1
  bcdUSB               1.00
  bDeviceClass            0 
  bDeviceSubClass         0 
  bDeviceProtocol         0 
  bMaxPacketSize0         8
  idVendor           0x0079 DragonRise Inc.
  idProduct          0x0006 PC TWIN SHOCK Gamepad
  bcdDevice            1.07
  iManufacturer           1 
  iProduct                2 
  iSerial                 0 
  bNumConfigurations      1
  Configuration Descriptor:
    bLength                 9
    bDescriptorType         2
    wTotalLength           41
    bNumInterfaces          1
    bConfigurationValue     1
    iConfiguration          0 
    bmAttributes         0x80
      (Bus Powered)
    MaxPower              500mA
    Interface Descriptor:
      bLength                 9
      bDescriptorType         4
      bInterfaceNumber        0
      bAlternateSetting       0
      bNumEndpoints           2
      bInterfaceClass         3 Human Interface Device
      bInterfaceSubClass      0 
      bInterfaceProtocol      0 
      iInterface              0 
        HID Device Descriptor:
          bLength                 9
          bDescriptorType        33
          bcdHID               1.10
          bCountryCode           33 US
          bNumDescriptors         1
          bDescriptorType        34 Report
          wDescriptorLength     101
         Report Descriptors: 
           ** UNAVAILABLE **
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x81  EP 1 IN
        bmAttributes            3
          Transfer Type            Interrupt
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0008  1x 8 bytes
        bInterval              10
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x01  EP 1 OUT
        bmAttributes            3
          Transfer Type            Interrupt
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0008  1x 8 bytes
        bInterval              10
`

Any idea ?

lordloki:
Perhaps you should load the hid_dr module. See this link: https://askubuntu.com/questions/302729/cannot-get-usb-gamepad-to-work-in-ubuntu-12-04/521703

panzergame:
Sorry to delay but my joystick still doesn't work :/

panzergame:
i have a good and bad news:
 good: I can now support joystick, my grub was just targeting the wrong kernel.

bad: My joystick joystick doesn't works in your branch contrary to regular blender.
I got an error message:
`SCA_Joystick::HandleEvents, Unknown SDL event (1541), this should not happen`

panzergame:
Also it's pretty vague: the patch is supposed to compile with only SDL2 or SDL1.2 ?

lordloki:
the "SCA_Joystick::HandleEvents, Unknown SDL event (1541), this should not happen" message is due to old SDL_Joysticks events. Now they are ignored.

I have activated the debug to see if the non-working issue is due to an unknow controller.

Could you compile again and test?

panzergame:
I have to compile SDL2 with "haptic" flag, now it works.

lordloki:
Ok. I will modify this night the patch to leave the debug comments by default (as they are not so intrusives).
I also add an additional comment when we detect that the joystick is not a game controller something similar to:

> #!/bin/sh
 export SDL_GAMECONTROLLERCONFIG="[the string you received from controllermap]"
 exec PATH_TO_UPBGE_PLAYER_EXECUTABLE

>

panzergame:
Why the second left and right shoulder are in the "axis" mode ? Contrary to BF BGE that only treated it as a button, also let the sensor manage if the shoulder is pressed or released is very old, we have the tap mode for this (even more if the current mode is inverted : release activated when pressing shoulder). Or script in python, anyway we have to keep consistent with other device sensors: keyboard and mouse.

Also my game controller seems very oddly mapped. I will look at this.

lordloki:
It is because the standard controller is the xbox360. In this controller, second left and right shoulders are not buttons, they are triggers (as in gun triggers). Then, they don't have a behaviour as a button (1 or 0), they have a behaviour similar to an axis (from 0 to 32000?) but only in 1 direction instead of 2 directions.

source/gameengine/GameLogic/Joystick/SCA_JoystickEvents.cpp
```diff
@@ -103,29 +83,17 @@ void SCA_Joystick::HandleEvents(void)
...
+			case SDL_CONTROLLERBUTTONDOWN:
+			case SDL_CONTROLLERBUTTONUP:
+				SCA_Joystick::m_instance[sdl_event.cbutton.which]->OnButtonEvent(&sdl_event);
 				break;
-#endif
-#if SDL_VERSION_ATLEAST(2, 0, 0)
-			case SDL_JOYDEVICEADDED:
-			case SDL_JOYDEVICEREMOVED:
-				/* pass */
+			case SDL_CONTROLLERAXISMOTION:
```
panzergame: Why using all CONTROLLER functions and not JOYSTICK ones ?
My right axis isn't detected here but if you replace by SDL_JOYAXISMOTION it hlaf works.

lordloki:
The issue with old joystick functions is the following:
- The behaviour from sdl 1 to sdl 2 in linux is the same but in windows the behaviour between sdl 1 and sdl2 is different.
- The behaviour between sdl2 in windows and sdl2 in linux is different.

This is due to sdl2 for windows now uses XInput and it ruins the behaviour win/linux/Macos.

Using the new controller functions the controller (once you map correctly) works exactly equal in win/linux/Macos

lordloki:
Just one more test: Could you mapping your joystick using Steam Big Mode, controllermapping in SDL2 test sources or antimicro app. After, I could include in the mapping string to test if it works ok for you

panzergame:
The order of JOYAXIS_[RIGHT/UP/DOWN/LEFT] is totally not the same as SDL_CONTROLLER_AXIS_[LEFT/RIGHT][X/Y].
I don't know for the moment the nicest way to solve this, maybe add a private conversion function in SCA_Joystick used in pGetAxis, also the m_axis_array usage in pAxisTest should be replaced by pGetAxis to use the conversion.
Button locks in the same case.

See also https://wiki.libsdl.org/SDL_GameControllerAxis

panzergame:
I found the reason finally the SDL_CONTROLLER_AXIS_RIGHTX is inverted with SDL_CONTROLLER_AXIS_LEFTY. But i guess i don't have the only controller which is wrongly mapped. @lordloki : Are there a different mapping data bases ?

lordloki:
I dont understand totally what you are saying about order. I think you are comparing complete axis vs directions of each axis.

I have tested my xbox360 controller and my previously mapped PS2 clone controller and both controller works well.

youle31:
I's like to test but my ps3 controller is broken... @lordloki : Do you know a good clone for Xbox wired controller? I'd like to buy one to check if all is ok for me.
@panzergame: I think it's difficult to have a good support for all kind of gamepad. Like in some games, with my old game pad, I had to use another software (XPadder on windows... don't know if it exists on Linux) to map correctly the buttons and axis.

panzergame:
@lordloki : Finally the order is right is just that my mapping is wrong but as the mapping is also false in BF BGE, then it works for this part.

I found something else disappointing, the axis value are accessed with axis*2 + dir, but the shoulder trigger are from the sdl no split in a pair of event (X and Y for stick), so the axis*2+dir failed since the *2 is not needed for the shoulder trigger.
Maybe create a similar function to pGetAxis but without the *2 ?
Also in this case why not in the UI set the left shoulder and the right shoulder a separated axis type and then for the axis flag use simply "shoulder trigger press" "shoulder trigger release" ?

lordloki:
@panzergame I did yesterday some more test related to shoulder and now I understand what you were trying to say me. I will see how to fix it.

panzergame:
Alright, thank you in advance.

youle31:
Hi! So I bought an xbox controller "afterglow". It's really ugly but there was not much choice in the shop haha. I begin my test here and I'll edit each time I find something.

1) I had to click on command line instructions and follow the steps to test. Then I reabsed and I had merge conflicts so I git rebase --skip... I hope it won't cause issues. Would it be possible to make a branch for this pull request panzergame so it would be easier to test and so we don't have to worry about rebase merge conflicts? We'll see if I have issues later.

2) So I compiled with SDL and I have a compilation error in E:\Compilation\Blender\upBGE\source\gameengine\GameLogic\Joystick\SCA_JoystickPrivate.h
related to memset... #include <string.h> fixes the compilation issue. I'm on windows10 64 MSVC 2013 update5.

3)First test: print(bge.logic.joysticks): http://www.pasteall.org/69806

4)Second test: Xbox pad only: http://www.pasteall.org/69807

I'll continue the test when we'll have fixed the crashes on my computer. @Lordloki: This looks awesome :) If you have some time one day and if you're a bit familiar with Visual Studio, I could invite you to take the control of my computer via "team viewer" so we can fix bugs. Or I'll try alone to fix it, or with the help of panzergame if he's ok. I don't know if we can make a 3 persons conference. We'll see. Thanks again! This will be awesome :)

lordloki:
@youle31 Thanks for testing.
I will see if I can compile the code under VS. 

2) I will commit the fix
3) It is related to hotpluggin (not implemented yet). I will see if I left something wrong there.
4) Yes, I think Axis, button, Single Axis order with Axis by default would be better. A threadhold of 5000 improve a lot the experience with sticks.

I hope to have time this week. Sorry for the delay.

youle31:
no hurry! don't worry. Thanks for what you did. And maybe we'll have fixed my issues before next week end :)

youle31:
@lordloki : Hi! I found the possible cause of the crashes I had:
SCA_Joystick::m_instance[sdl_event.caxis.which]->OnAxisEvent(&sdl_event);
in void SCA_Joystick::HandleEvents(void) in E:\Compilation\Blender\upBGE\source\gameengine\GameLogic\Joystick\SCA_JoystickEvents.cpp

m_instance[8] is the array of 8 sdl_joysticks... And sdl_event.caxis.which return the SDL_Joystick ID. 

The first time I launch bge,  which = 0 (this is correct and I have no crash). Then I escape. The second time I launch bge, which = 1 (this is incorrect and cause the crash.)

As I plugged only one stick, if I replace SCA_Joystick::m_instance[sdl_event.caxis.which]->OnAxisEvent(&sdl_event); with SCA_Joystick::m_instance[0]->OnAxisEvent(&sdl_event);
it doesn't crash anymore.

panzergame:
@lordloki : The internal code is very not maintained and badly written, what about ruse a C++ library for joystick ? I was thinking of SFML : http://www.sfml-dev.org/documentation/2.0-fr/classsf_1_1Joystick.php. Of course your work will be not lost because you refactored the UI aspect of the joystick sensor.

lordloki:
@youle31
1) really it is not necessary. I should close correctly when main SDL have closed.
2) I dont know. I have to investigate.
3) In my case they are inverted. It is due to an order issue. I will fix

@panzergame 
I think that controller api is not matured yet but I think this library is the standard now. Using the smfl library methods is like back to early SDL1.2 working way (detecting how many axis, buttons, etc has a joystick and to try to make a standard from this).

For me there is 2 ways:
- SDL2: It defines what it is the standard first and after define the methods to work with this standard. It is very close to how the consoles (Xbox, Playstation, etc) work. The issue is as they cant stablish a specific hardware they need that the people map its hardware  (joysticks, controller, gamepads, etc) to the standard. They try to mitigate the issue supplying a database with almost all game controller hardware already mapped (well not all but it is increasing in good rate).

- SDL1.2, ftml, etc: It tries to detect all axis, hats and buttons and from there it tries to stablish a standard. The issue in this case is that the index of the axis or the buttons is not the same in all controllers and for example, maybe in 1 joystick the button A in other joystick is the button start. 

For me the SDL2 way is THE WAY :-) but if you want to go with the sdl1.2 way I think it is much easier left the code as it is and for windows compile the SDL2 library without Xinput support.

panzergame:
Well, in this case do you have time to fix the bug noticed above ?

lordloki:
Yes :-). I will do an effort to fix it this weekend.

panzergame:
Thanks :)

lordloki:
At last, I have divided the Sensor UI in 4 items (not commited yet):
- Stick Directions (default): 4 direction for each Left/Right Stick. Threshold: 5000 by default
- Stick Axis: 2 axis horz/vert per Left/Right Stick. Threshold: 5000 by default
- Buttons
- Shoulder Triggers: Left and Right trigger. Threshold: 2000 by default

I was checking index issue reported by @youle31 but I cant repeat it myself under linux. I'm trying to setup a windows (MSVC2013) build but without so much luck. (it compiles but it crash before it starts, no related to this issue).

lordloki:
@youle31 following irc talk, really strange as SDL_Init() calls to SDL_SubSystemInit()

> int
SDL_Init(Uint32 flags)
{
    return SDL_InitSubSystem(flags);
}

panzergame:
@lordloki: Thanks for the update. I will review it today.

lordloki:
@panzergame the branch has a merge conflict (i imagine due to your handleEvents() fix/movement to Ketjsi) but I don't know how to update my master branch to stay in syncrony with the upbge master branch.

Do you know how can I update my github master to don't break it (my git knowledge is limited)?

panzergame:
```
git checkout master
git pull --rebase
git rebase master your_branch
# Solve conflict
git add source/
git rebase --continue
git push origin you_branch --force
```

panzergame:
@lordloki : For me "Right Stick Horizontal is inverted with "Left Stick Vertical" in "Stick Axis" event type. Can you check if it's controller independent issue ?

Other than that everything work for shoulders.

lordloki:
@panzergame it is working right for my 2 gamecontrollers (microsoft x360 original and ps2 clone ones).
Could you mapping your joystick to see if the mapping is correct?
You can download this app (compiled by me from SDL2 sources) https://drive.google.com/open?id=0B_aHdCLOBEQSZE5TTFAxeTVJOEU

you should uncompress and execute "mapping_controller 0"
After copy here the mapping.

youle31:
I'm sorry but I still have the same issue... I don't know if it can help but here: http://www.pasteall.org/71184 is the console output when I apply this: http://www.pasteall.org/71185/diff

I opened and closed the ge 3 times with 1 xbox controller on windows 10 64 bits upbge 32 bits...

^$? Weird... I'll try to download directly your master if it's possible because when I rebased your master  on upbge, I had conflicts and I skipped (could it be the cause of this weird behaviour?...). I'll post a message if I find something.

source/gameengine/GameLogic/SCA_PythonJoystick.h
```diff
@@ -34,11 +34,12 @@ class SCA_PythonJoystick : public PyObjectPlus
 	Py_Header
 private:
 	class SCA_Joystick *m_joystick;
+	int m_joyindex;
 #ifdef WITH_PYTHON
 	PyObject* m_event_dict;
 #endif
 public:
-	SCA_PythonJoystick(class SCA_Joystick* joystick);
+	SCA_PythonJoystick(class SCA_Joystick* joystick, int m_joyindex);
```
youle31: int m_joyindex shouldn't be replaced by int joyindex? (just for esthetic...)

panzergame:
@lordloki : I opened an issue in the mapping repository for my joystick : https://github.com/gabomdq/SDL_GameControllerDB/issues/86.

youle31:
This (dirty) patch: http://www.pasteall.org/71214/diff solves crashes and hotplug on my computer. Vibrations are not detected for the moment. I'll investigate on this.

lordloki:
I hope it works now. It is a headache to develop this using a virtualbox environment. I had to setup a visual studio to check in windows but debug build takes an eternity to link (+15min)

BluePrintRandom:
Thank you for all your hard work LordLoki :D

youle31:
@lordloki : I'm very sorry, bad news. I still have issues. Hotplugging works. All is fine the first time I run the game engine. The second time I run the game engine, the good new is that I have no crash anymore, but the gamepad doesn't work, so there is an instance of the gamepad but the axis and the buttons do nothing. If I comment SDL_QuitSubSystem... it half solves the problem if I don't play with plug/unplug...

I saw that you reinitiazed all static variables in Init. That's a good thing I think but something is not reInited at ge start or something is not freed at ge exit... I don't understand. The code is very hard and SDL sucks (on windows with my gamepad at least)

If you want I propose to you to work on my computer via team viewer. We can find a way for you to control my gamepad from your computer. ReCompilation time is ~10 seconds, and you could see the issues. If you don't know how to do, I'll explain to you how to make break points on visual studio.

I think placing SCA_Joystick::Init() in SCA_JoystickManager constructor is a bad idea and all should be done in next frame to create and update m_joystick[] list. The instances should be initialized only in SCA_JoystickEvents in case GAMECONTROLLERADDED and GAMECONTROLLERREMOVED. Like this we won't have to create instances several times. But this supposed to tell to the joystick sensor to not listen till the sdl_event has not added joysticks.

I'm sorry that it is so complex. You can join me on my mail: you.le@live.fr if you want to organize team viewer session (no audio, we can chat by text) or on upbgecoders (I'm often connected)

lordloki:
Hi @youle31. I found a way to follow your recomendations. it seems that SDL_CONTROLLERDEVICEADDED doesn't report about all controllers connected at beginning but SDL_JOYADDED does. Then using SDL own SDL_JoystickID paramenter we can control the id madness.
At this moment, I checked it in linux only.
Thanks for your help offer but my coding time is very erratic (i.e. from 4:30am to 6am before go to work). But if i wouldnt solve this issue before next week end I will ask for your visual studio team support as i will have about 2 days at weekend next week to coding without problems all day (i will be in blender meeting at spain :-)).

youle31:
AND IT WORKS!!!!!!!!!!!! Great job! Thanks. All is ok for me. I tested with 2 gamepads and all seems ok. No crash at all. hot plugging works well. Even my old game pad which doesn't work anymore in pes 2016 works in Blender haha. Thanks very much for your hard work! (When I play with hotplugging with 2 pads I can still cause have one of the 2 pads that doesn't work but no crash and this was just for hard testing the patch. It's not important. I'll try to find a solution for this minor bug). For me, it's ready to commit. That's great!

lordloki:
At last!! That's great. I was testing in windows last night and the issue with hottpluging comes from:
Sometimes xinput pad (ie xbox360) needs to be connected in first place. If not it receive the same index that a previous not xinput connected controller. It is reported as sdl bug. I imagine that it will be fixed in the future.

source/blender/makesrna/intern/rna_sensor.c
```diff
@@ -867,41 +867,67 @@ static void rna_def_joystick_sensor(BlenderRNA *brna)
...
-		{SENS_JOY_HAT_DOWN, "DOWN", 0, "Down", ""},
-		{SENS_JOY_HAT_LEFT, "LEFT", 0, "Left", ""},
-		{SENS_JOY_HAT_RIGHT, "RIGHT", 0, "Right", ""},
-
-		{SENS_JOY_HAT_UP_RIGHT, "UPRIGHT", 0, "Up/Right", ""},
-		{SENS_JOY_HAT_DOWN_LEFT, "DOWNLEFT", 0, "Down/Left", ""},
-		{SENS_JOY_HAT_UP_LEFT, "UPLEFT", 0, "Up/Left", ""},
-		{SENS_JOY_HAT_DOWN_RIGHT, "DOWNRIGHT", 0, "Down/Right", ""},
+	static EnumPropertyItem axis_trigger_items[] = {
+	    {SENS_JOY_LEFT_SHOULDER_TRIGGER, "LEFT_SHOULDER_TRIGGER", 0, "Left Shoulder Trigger", ""},
```
panzergame: I guess you have an issue with the indentation, it was two tabs before and it's a tabs + 4 spaces here. In the code we see nothing as a tab is 4 spaces, but it can show that your editor is using only spaces and that all the code is with spaces indentation. @lordloki: Can you check that, and replace with tabs if it's possible ?

source/blender/makesrna/intern/rna_sensor.c
```diff
@@ -915,53 +941,48 @@ static void rna_def_joystick_sensor(BlenderRNA *brna)
...
 
 	/* Button */
-	prop = RNA_def_property(srna, "button_number", PROP_INT, PROP_NONE);
-	RNA_def_property_int_sdna(prop, NULL, "button");
-	RNA_def_property_ui_text(prop, "Button Number", "Which button to use");
-	RNA_def_property_range(prop, 0, 18);
+	prop = RNA_def_property(srna, "button_number", PROP_ENUM, PROP_NONE);
+	RNA_def_property_enum_sdna(prop, NULL, "button");
+	RNA_def_property_enum_items(prop, button_items);
+	RNA_def_property_ui_text(prop, "Buttons", "Which button to use");
```
panzergame: "Buttons" --> "Button" ? As it's a single button.

source/gameengine/GameLogic/CMakeLists.txt
```diff
@@ -28,6 +28,7 @@ set(INC
 	../Expressions
 	../Rasterizer
 	../SceneGraph
+	../../blender/blenkernel
```
panzergame: Is for what ?

source/gameengine/GameLogic/Joystick/SCA_Joystick.cpp
```diff
@@ -74,95 +72,82 @@ SCA_Joystick::~SCA_Joystick()
...
+		unsigned short i = 0;
+		const char *mapping_string = NULL;
+		mapping_string = controller_mappings[i];
+
+		while (mapping_string) {
+			SDL_GameControllerAddMapping(mapping_string);
+			i++;
+			mapping_string = controller_mappings[i];
+	    }
+#if 0
```
panzergame: This commented code is used for debug ?

source/gameengine/GameLogic/Joystick/SCA_Joystick.cpp
```diff
@@ -200,119 +185,151 @@ bool SCA_Joystick::aAxisIsPositive(int axis_single)
...
+	{
+		return true;
+	}
 #endif
+	return false;
 }
 
-
-bool SCA_Joystick::aHatIsPositive(int hatnum, int dir)
+bool SCA_Joystick::CreateJoystickDevice(void)
```
panzergame: you don't need the void argument for C++

source/gameengine/GameLogic/Joystick/SCA_Joystick.cpp
```diff
@@ -200,119 +185,151 @@ bool SCA_Joystick::aAxisIsPositive(int axis_single)
...
 {
-#ifndef WITH_SDL
-	return false;
-#else
-	bool result;
-	result = !(SDL_CHECK(SDL_JoystickGetButton) && SDL_JoystickGetButton(m_private->m_joystick, button));
-	return result;
+#ifdef WITH_SDL
+	if (!(SDL_CHECK(SDL_GameControllerGetButton) &&
+       SDL_GameControllerGetButton(m_private->m_gamecontroller, (SDL_GameControllerButton)button)))
```
panzergame: For code style it would be nice to indent this line, or align it with the first "(".

source/gameengine/GameLogic/Joystick/SCA_Joystick.cpp
```diff
@@ -74,95 +72,82 @@ SCA_Joystick::~SCA_Joystick()
...
+	else {
+		printf("Error initializing SDL Game Controller: %s\n", SDL_GetError());
 	}
-	return m_instance[joyindex];
-#endif /* WITH_SDL */
+#endif
 }
 
-void SCA_Joystick::ReleaseInstance()
+void SCA_Joystick::Close()
```
panzergame: If this function need to be called only one time and not per scene because there one joystick manager per scene, then move its call in `LA_Launcher::ExitEngine`. Same for `Init` in `LA_Launcher::InitEngine`

source/gameengine/GameLogic/Joystick/SCA_Joystick.cpp
```diff
@@ -74,95 +72,82 @@ SCA_Joystick::~SCA_Joystick()
 }
 
 SCA_Joystick *SCA_Joystick::m_instance[JOYINDEX_MAX];
-int SCA_Joystick::m_joynum = 0;
-int SCA_Joystick::m_refCount = 0;
+bool SCA_Joystick::m_joystickupdatestatus = false;
```
panzergame: Can you explain the use of this static variable ? As i saw it's used only in `SCA_JoystickManager` to update self joystick list. But this list is exactly the same as `SCA_Joystick::m_instance`. I made a patch which remove `m_joystick` in `SCA_JoystickManager` and remove `m_joystickupdatestatus`. : http://www.pasteall.org/71764/diff. I noticed no problems currently.

source/gameengine/GameLogic/Joystick/SCA_Joystick.cpp
```diff
@@ -200,119 +185,151 @@ bool SCA_Joystick::aAxisIsPositive(int axis_single)
...
+			 *						   AXIS_TRIGGERLEFT and AXIS_TRIGGERRIGHT.
+			 *
+			 * 15 buttons availables:  BUTTON_A, BUTTON_B, BUTTON_X, BUTTON_Y,
+			 * (in order from 0 to 14) BUTTON_BACK, BUTTON_GUIDE, BUTTON_START,
+			 *						   BUTTON_LEFTSTICK, BUTTON_RIGHTSTICK,
+			 *						   BUTTON_LEFTSHOULDER, BUTTON_RIGHTSHOULDER,
+			 *						   BUTTON_DPAD_UP, BUTTON_DPAD_DOWN,
+			 *						   BUTTON_DPAD_LEFT and BUTTON_DPAD_RIGHT.
+			 */
+			m_axismax = SDL_CONTROLLER_AXIS_MAX;
```
panzergame: `m_axismax` and `m_buttonmax` are constants, no ? I saw later that you set then to 0. These values are used to know if a joystick is valid, or they can be replaced by static function in `SCA_Joystick` ?

source/gameengine/GameLogic/Joystick/SCA_Joystick.cpp
```diff
@@ -200,119 +185,151 @@ bool SCA_Joystick::aAxisIsPositive(int axis_single)
...
-			JOYSTICK_ECHO("Closing-joystick " << m_joyindex);
-			SDL_JoystickClose(m_private->m_joystick);
-		}
+
+			if (m_private->m_haptic && SDL_CHECK(SDL_HapticClose)) {
+				SDL_HapticClose(m_private->m_haptic);
+				m_private->m_haptic = NULL;
+			}
+
+			if (m_private->m_gamecontroller && SDL_CHECK(SDL_GameControllerClose)) {
```
panzergame: If these conditions are exclusive, you can use `else if` i guess.

source/gameengine/GameLogic/Joystick/SCA_JoystickEvents.cpp
```diff
@@ -37,45 +37,29 @@
 #endif
 
 #ifdef WITH_SDL
-void SCA_Joystick::OnAxisMotion(SDL_Event* sdl_event)
+void SCA_Joystick::OnAxisEvent(SDL_Event* sdl_event)
 {
-	if (sdl_event->jaxis.axis >= JOYAXIS_MAX)
+	if (!this || sdl_event->caxis.axis >= JOYAXIS_MAX)
```
panzergame: Can you explain the `!this` condition ? I guess is to test if the joystick is non NULL but it should be done before call the `OnAxisEvent` function.

source/gameengine/GameLogic/Joystick/SCA_JoystickEvents.cpp
```diff
@@ -37,45 +37,29 @@
...
+	m_axis_array[sdl_event->caxis.axis] = sdl_event->caxis.value;
 	m_istrig_axis = 1;
 }
 
 /* See notes below in the event loop */
-void SCA_Joystick::OnHatMotion(SDL_Event* sdl_event)
+void SCA_Joystick::OnButtonEvent(SDL_Event* sdl_event)
 {
-	if (sdl_event->jhat.hat >= JOYHAT_MAX)
+	if (!this)
```
panzergame: Same here

source/gameengine/GameLogic/Joystick/SCA_JoystickEvents.cpp
```diff
@@ -37,45 +37,29 @@
...
-	// sdl_event->jbutton.button;
-	
 	m_istrig_button = 1;
 }
 
 
 void SCA_Joystick::OnNothing(SDL_Event* sdl_event)
 {
-	m_istrig_axis = m_istrig_button = m_istrig_hat = 0;
+	if (!this)
```
panzergame: And here.

source/gameengine/GameLogic/Joystick/SCA_JoystickEvents.cpp
```diff
@@ -86,48 +70,73 @@ void SCA_Joystick::HandleEvents(void)
...
-				SCA_Joystick::m_instance[sdl_event.jaxis.which]->OnAxisMotion(&sdl_event);
-				break;
-			case SDL_JOYHATMOTION:
-				SCA_Joystick::m_instance[sdl_event.jhat.which]->OnHatMotion(&sdl_event);
-				break;
-			case SDL_JOYBUTTONUP:
-				SCA_Joystick::m_instance[sdl_event.jbutton.which]->OnButtonUp(&sdl_event);
+			case SDL_JOYDEVICEADDED:
+				if (sdl_event.jdevice.which < JOYINDEX_MAX) {
+					if (!SCA_Joystick::m_instance[sdl_event.jdevice.which]) {
```
panzergame: I think it would be interesting to print a message if an added joystick is added on an existing slot.

source/gameengine/GameLogic/SCA_JoystickSensor.cpp
```diff
@@ -374,48 +352,42 @@ PyObject *SCA_JoystickSensor::pyattr_get_axis_single(void *self_v, const KX_PYAT
 	return PyLong_FromLong(joy ? joy->GetAxisPosition(self->m_axis - 1) : 0);
 }
 
+PyObject *SCA_JoystickSensor::pyattr_check_hat(void *self_v, const KX_PYATTRIBUTE_DEF *attrdef)
+{
+	ShowDeprecationWarning("SCA_JoystickSensor.hat", "SCA_JoystickSensor.button");
+	return NULL;
```
panzergame: I don't know about throwing an error for compatibility issues. Can you do a kind of conversion from buttons ? if not i prefer to return a dumb value like 0 every times instead of raise an error.

source/gameengine/GameLogic/SCA_JoystickSensor.cpp
```diff
@@ -374,48 +352,42 @@ PyObject *SCA_JoystickSensor::pyattr_get_axis_single(void *self_v, const KX_PYAT
...
-	
-	int hat_index = (joy ? joy->GetNumberOfHats() : 0);
-	PyObject *list = PyList_New(hat_index);
-	
-	while (hat_index--) {
-		PyList_SET_ITEM(list, hat_index, PyLong_FromLong(joy->GetHat(hat_index)));
-	}
-	
-	return list;
+	ShowDeprecationWarning("SCA_JoystickSensor.hat", "SCA_JoystickSensor.button");
```
panzergame: Same here.

source/gameengine/GameLogic/SCA_JoystickSensor.cpp
```diff
@@ -374,48 +352,42 @@ PyObject *SCA_JoystickSensor::pyattr_get_axis_single(void *self_v, const KX_PYAT
...
-	return PyLong_FromLong( joy ? joy->GetNumberOfButtons() : 0 );
+	return PyLong_FromLong( joy ? JOYBUT_MAX : 0 );
 }
 
 PyObject *SCA_JoystickSensor::pyattr_get_num_hats(void *self_v, const KX_PYATTRIBUTE_DEF *attrdef)
 {
-	SCA_JoystickSensor* self = static_cast<SCA_JoystickSensor*>(self_v);
-	SCA_Joystick *joy = ((SCA_JoystickManager *)self->m_eventmgr)->GetJoystickDevice(self->m_joyindex);
-	return PyLong_FromLong( joy ? joy->GetNumberOfHats() : 0 );
+	ShowDeprecationWarning("SCA_JoystickSensor.numHats", "SCA_JoystickSensor.numButtons");
```
panzergame: And here too.

source/gameengine/GameLogic/CMakeLists.txt
lordloki: part of first sdlgamecontrollerdb.txt reader. Deleted now.

source/blender/makesrna/intern/rna_sensor.c
lordloki: Done

source/blender/makesrna/intern/rna_sensor.c
lordloki: Done

source/gameengine/GameLogic/Joystick/SCA_Joystick.cpp
lordloki: Yes. Removed

source/gameengine/GameLogic/Joystick/SCA_Joystick.cpp
lordloki: aligned now

source/gameengine/GameLogic/Joystick/SCA_Joystick.cpp
lordloki: Moved to LA_Launcher

source/gameengine/GameLogic/Joystick/SCA_Joystick.cpp
lordloki: it was for hotpluggin but I have tested your patch and it works. Then i removed joystickupdatestatus code

source/gameengine/GameLogic/Joystick/SCA_Joystick.cpp
lordloki: They have differents values depending of SDL or not SDL.

source/gameengine/GameLogic/Joystick/SCA_Joystick.cpp
lordloki: they are not exclusives

source/gameengine/GameLogic/Joystick/SCA_JoystickEvents.cpp
lordloki: they were for debug purposes. Now removed

source/gameengine/GameLogic/Joystick/SCA_JoystickEvents.cpp
lordloki: added

source/gameengine/GameLogic/SCA_JoystickSensor.cpp
lordloki: I think it is preferible to give a clue about how to procede now instead of leave to the users asking the reason of that 0 value.

lordloki:
Thanks for reviewing @panzergame.

lordloki:
I see that we need to release joysticks. Tomorrow i will update it

panzergame:
Thanks for the modifications.

lordloki:
I think it is ready for merge now

source/gameengine/Launcher/LA_Launcher.cpp
```diff
@@ -276,6 +280,7 @@ void LA_Launcher::InitEngine()
 
 void LA_Launcher::ExitEngine()
 {
+	SCA_Joystick::Close();
```
panzergame: It assumes that the joystick sensor destructor doesn't call any functions in SCA_Joystick.

source/gameengine/GameLogic/SCA_PythonJoystick.h
panzergame: @lordloki : I think you should replace m_joyindex by joyindex as noticed youle.

lordloki: i forgot it. Done now

lordloki:
Hi @panzergame, to include it in the commit log:

**Joystick Refactor**

This Joystick refactor comes from the issues resulted from the migration to SDL2 library.

SDL2 uses Xinput library for Windows and it works different from old mm and dinput Windows libraries and additionally the Windows behaviour is not consistent with linux/MacOS behaviour.

To mitigate this issue SDL2 implements a new Game Controller API. This Game Controller API changes totally the philosophy of Joystick support on PC comparing it to Joystick support on Consoles (PS4, Xbox, etc). That is: It defines that the standard it is the XBox 360 controller and It needs that all the Joysticks are mapped against it. For this purpose, this new API uses a intermediate file (SCA_JoystickMappingdb.h) where the joysticks are mapped to follow a similar configuration than Xbox 360 game controller. This file is a slightly different version of file from https://github.com/gabomdq/SDL_GameControllerDB/blob/master/gamecontrollerdb.txt (which it is the community effort to map every Joystick against XBox360 one).

Additionally, If a Joystick is not mapped there are several tools availables (Steam Big Mode, controllermapping in SDL2 test sources and antimicro) to perform this mapping.

Then, To clarify a little how it works now regarding our unintuitive Joystick sensor:
A Game Controller is composed of:

6 axis availables (in order from 0 to 5): 

- AXIS_LEFTSTICK_X, AXIS_LEFTSTICK_Y, AXIS_RIGHTSTICK_X, AXIS_RIGHTSTICK_Y, AXIS_TRIGGERLEFT and AXIS_TRIGGERRIGHT.

15 buttons availables (in order from 0 to 14): 

-  BUTTON_A, BUTTON_B, BUTTON_X, BUTTON_Y, BUTTON_BACK, BUTTON_GUIDE, BUTTON_START,                      BUTTON_LEFTSTICK, BUTTON_RIGHTSTICK, BUTTON_LEFTSHOULDER, BUTTON_RIGHTSHOULDER,                      BUTTON_DPAD_UP, BUTTON_DPAD_DOWN, BUTTON_DPAD_LEFT and BUTTON_DPAD_RIGHT.

The Joystick sensor UI has been reworked to follow these new axis and buttons set-up.

Also, to exploit new SDL2 hotpluggin capabilities the Joystick code has been refactored to create or destroy a Game Controller when it receives a SDL_CONTROLLERDEVICEADDED or SDL_CONTROLLERDEVICEREMOVED event only, instead of to create all joysticks at start. This way the UPBGE support Joystick hotpluggin now.

I have tested this patch in Linux/Windows with a Xbox 360 controller and a PS2 clone. Xbox 360 and PS2 clone worked out the box and the behaviour was consistent between linux and windows (not as now).

WARNING:
This patch is NOT BACKWARD COMPATIBLE due to new API. Between other things the new API has not "hat" concept and it processes "hat/dpad" control as any other button (BUTTON_DPAD_UP, BUTTON_DPAD_DOWN, BUTTON_DPAD_LEFT and BUTTON_DPAD_RIGHT). Also, it implement Shoulder trigger buttons (left/right) as independent axis.

Moreover, to implement a "do_version" it is very hard because we would have to differentiate between MacOS&Linux and Windows, and between Joysticks and Game Controllers. The only way to make it compatible with previous versions is to compile SDL2 disabling XINPUT and enabling MM/DINPUT but this way we would loose hottpluggin, haptic (vibration) and standatization (at last) features.

youle31:
On my computer vibrations doesn't work with SDL 2.0.3 and my fake xbox controller "afterglow". Here is a quick hack to make it work. I applied this patch just to test: http://www.pasteall.org/72021/diff .
Then I replaced SDL2.dll in my build folder with one I found here: https://www.libsdl.org/download-2.0.php ("SDL2-2.0.4-win32-x86.zip" for windows 32 bits) and it works :) Maybe if someone knows Juicyfruit he could ask him to update SDL libs for Windows to SDL 2.0.4, otherwise there is this hack.

lordloki:
I think that SDL2 library from Blender is built without vibration support (my original xbox 360 pad is not recognized as a haptic device either). The hack of replace sdl2.dll by the 2,0.4 one, works.
Anyway, i want to leave the vibration part for later, after current code will be merged with master. The current vibration code that it is in place only detects if the device supports vibration or not.

lordloki:
@youle, if you want to implement a vibration actuator (it's the following logical step) you can inspire with testhaptic.c file from sdl2 source

panzergame:
Commited : 57b5c01336c6d53118e9fe481ea3991166c3acb2

PR #101, opened 2016-05-09
In ge_filter, replace setTexture(m_textures index, bindcode) with setTexture(name, bindcode)

youle31:
test file: http://www.pasteall.org/blend/41793


youle31:
@DCubix : Hey! Could I ask what do you think of the last commit? I leaved explainations... Panzergame would like to have your opinion too. He was not convinced that we should setTexture for 2D filter in a single step instead of two. Thanks!

panzergame:
The issue with setTexture is that is it based on previous non-existing bindcode (=0), but if the texture bindcode changed e.g VideoTexture, how the user will be able to overwrite the previous invalid bindcode ?

It's why i prefer the setTexture(index, bindcode) because it explicitly overwrite the emplacement "index" with "bindcode".

youle31:
Yes. This Is why i propose to add a third optionnal arg to choose to force m-texture index

youle31:
will replace with another method

PR #112, opened 2016-05-18
UPBGE: Add a new method to setUniformTexture in KX_2DFilter

youle31:
I think the system with setTexture then setSampler is too complex to
understand for the user. But as setTexture method could be used to assign
a bindcode to m_texture[index] without having to setSampler, I propose to
add another method in KX_2DFilter to set uniform as we did usually in
BL_Shader methods: setUniform...(name, value). I think this is more
understandable for the end user.
The setUniformTexture(textureName, textureBindCode, index) method sets
uniform sampler2D texturename with the bindcode of the texture we want to
bind in the 2D Filter. Index argument is optional: It allows the user to
choose at which emplacement of m_textures list he wants to bind his
texture. Like that, if all the emplacement in m_textures list are already
filled, the user can choose to overwrite an existing emplacement with his
texture bindcode.
If all emplacements in m_textures list are already filled, the user can't
bind additional texture in his 2D filter. A warning message inform him
that all the available m_textures slots are already filled with a
bindcode, so that he has to use optional index argument to overwrite
m_textures[index] with his bindcode.

test file: http://www.pasteall.org/blend/41949

### DIFF
--- a/doc/python_api/rst/bge_types/bge.types.KX_2DFilter.rst
+++ b/doc/python_api/rst/bge_types/bge.types.KX_2DFilter.rst
@@ -18,3 +18,14 @@ base class --- :class:`BL_Shader`
       :type index: integer
       :arg bindCode: The texture bind code/Id.
       :type bindCode: integer
+
+   .. method:: setUniformTexture(textureName, textureBindCode, index)
+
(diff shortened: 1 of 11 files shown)


panzergame:
see c4ad61a7e57e82ca7ad182379ea34c2c50027304

PR #113, opened 2016-05-18
UPBGE: Reload GPU textures when were invalidated.

panzergame:
Previosuly any call to GPU_set_anisotropic or GPU_set_gpu_mipmapping was
freeing all gpu textures, but BL_Texture was still using the old pointer
to this freed texture. This was not causing segfault but the bind code
was just invalid.

To solve this we first ensure that the gpu texture we will not be freed
by calling GPu_texture_ref in BL_Texture which has as effect to increment
the user counter in the gpu texture. But it's not enought since even if we
have a valid texture the material is already using the a new copy of the
gpu texture (which is cooresponding to the needs by GPU_set_anisotropic
and GPU_set_gpu_mipmapping). To get a pointer of the used texture in BL_Texture
we use a new function GetGPUTexture.
This function returns always the updated gpu texture. Internally of this
function we can detect that a texture is only owned by us by caling
GPU_texture_ref_count, if it returns 1 then it means that no anymore materials
are using it and that only the BL_Texture keep it alive (by the previous
call to GPU_texture_ref). Then we free the texture by freeing it in the
image and get the current used gpu texture with GPU_texture_from_blender.
The new gpu texture pointer is copied in m_gpuTex.
The GetGPUTexture function is called everytime we need to use the gpu
texture and must replace any direct use of m_gpuTex other than in the
BL_Texture constructor.

Add the BL_Texture destructor if there a gpu texture we release the user
counter by calling GPU_texture_free.

### DIFF
--- a/source/blender/gpu/intern/gpu_texture.c
+++ b/source/blender/gpu/intern/gpu_texture.c
@@ -726,6 +726,11 @@ void GPU_texture_ref(GPUTexture *tex)
 	tex->refcount++;
 }
 
+int GPU_texture_ref_count(GPUTexture *tex)
+{
+	return tex->refcount;
+}
+
 int GPU_texture_target(const GPUTexture *tex)
 {
 	return tex->target;
--- a/source/blender/gpu/GPU_texture.h
+++ b/source/blender/gpu/GPU_texture.h
@@ -81,6 +81,7 @@ void GPU_invalid_tex_free(void);
 void GPU_texture_free(GPUTexture *tex);
 
 void GPU_texture_ref(GPUTexture *tex);
+int GPU_texture_ref_count(GPUTexture *tex);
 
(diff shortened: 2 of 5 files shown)


source/blender/gpu/intern/gpu_texture.c
```diff
@@ -726,6 +726,11 @@ void GPU_texture_ref(GPUTexture *tex)
 	tex->refcount++;
 }
 
+int GPU_texture_ref_count(GPUTexture *tex)
+{
+	return tex->refcount;
+}
```
DCubix: Isn't it easier to just call tex->refcount instead of an entire new method? :question:

panzergame: As GPUMaterial, these classes are defined in the .c and not in the header to obligate the developer to use getter and setter (it make debug easier). See it like a private data of a C++ class.

PR #121, opened 2016-05-28
Refactor KX_Lod to implement python API for lod levels (and eventually lodManager)

youle31:
I renamed KX_LodList to KX_LodManager.
I created a separate class for old struct Level outside of KX_LodManager.

The goal was to have
KX_GameObject.lodManager.lodLevel[index].attributeGet/set

This needs to be reviewed as it's my first little
"refactor"

We could nevertheless need a struct m_savedData to save Blender lodlevel attributes at game engine start and restore it at game engine exit.

Test file: http://www.pasteall.org/blend/42108

New test file for the complete python API: http://www.pasteall.org/blend/42122

### DIFF
--- a/doc/python_api/rst/bge_types/bge.types.KX_Scene.rst
+++ b/doc/python_api/rst/bge_types/bge.types.KX_Scene.rst
@@ -143,6 +143,12 @@ base class --- :class:`PyObjectPlus`
 
       :type: Vector((gx, gy, gz))
 
+   .. attribute:: lodGlobalScale
+
+      All the scene objects with level of detail LOD scale.
+
+      :type: float
+
    .. method:: addObject(object, reference, time=0.0)
 
       Adds an object to the scene like the Add Object Actuator would.
--- a/doc/python_api/rst/bge_types/bge.types.KX_Scene.rst
+++ b/doc/python_api/rst/bge_types/bge.types.KX_Scene.rst
@@ -143,17 +143,6 @@ base class --- :class:`PyObjectPlus`
 
       :type: Vector((gx, gy, gz))
 
-   .. method:: setLodScale(pythonObjectList, scale)
-
-      Set LOD scale for all objects with LOD in the pythonObjectList.
-
-      :arg pythonObjectList: list of KX_GameObject with LOD to set 
-      :type: list
-      :arg scale: factor to scale initial LOD distance. If the initial distance of a
-      KX_GameObject in the list is 25.0 and the scale is 2.0, the LOD distance for
-      this KX_GameObject will become 50.0.
-      :type: float
-
    .. method:: addObject(object, reference, time=0.0)
 
       Adds an object to the scene like the Add Object Actuator would.
--- a/doc/python_api/rst/bge_types/bge.types.KX_Scene.rst
+++ b/doc/python_api/rst/bge_types/bge.types.KX_Scene.rst
@@ -143,10 +143,15 @@ base class --- :class:`PyObjectPlus`
 
       :type: Vector((gx, gy, gz))
 
-   .. attribute:: lodGlobalScale
+   .. method:: setLodScale(pythonObjectList, scale)
 
-      All the scene objects with level of detail LOD scale.
+      Set LOD scale for all objects with LOD in the pythonObjectList.
 
+      :arg pythonObjectList: list of KX_GameObject with LOD to set 
+      :type: list
+      :arg scale: factor to scale initial LOD distance. If the initial distance of a
+      KX_GameObject in the list is 25.0 and the scale is 2.0, the LOD distance for
+      this KX_GameObject will become 50.0.
       :type: float
 
    .. method:: addObject(object, reference, time=0.0)
--- a/source/gameengine/Ketsji/KX_GameObject.cpp
+++ b/source/gameengine/Ketsji/KX_GameObject.cpp
@@ -2055,6 +2055,7 @@ PyAttributeDef KX_GameObject::Attributes[] = {
 	KX_PYATTRIBUTE_RW_FUNCTION("debug",	KX_GameObject, pyattr_get_debug, pyattr_set_debug),
 	KX_PYATTRIBUTE_RO_FUNCTION("components", KX_GameObject, pyattr_get_components),
 	KX_PYATTRIBUTE_RW_FUNCTION("debugRecursive",	KX_GameObject, pyattr_get_debugRecursive, pyattr_set_debugRecursive),
+	KX_PYATTRIBUTE_RO_FUNCTION("lod", KX_GameObject, pyattr_get_lodlist),
 	
 	/* experimental, don't rely on these yet */
 	KX_PYATTRIBUTE_RO_FUNCTION("sensors",		KX_GameObject, pyattr_get_sensors),
@@ -3273,6 +3274,17 @@ int KX_GameObject::pyattr_set_debugRecursive(void *self_v, const KX_PYATTRIBUTE_
 	self->SetUseDebugProperties(param, true);
 
 	return PY_SET_ATTR_SUCCESS;
+}
+
+PyObject *KX_GameObject::pyattr_get_lodlist(void *self_v, const KX_PYATTRIBUTE_DEF *attrdef)
+{
+	KX_GameObject *self = static_cast<KX_GameObject*>(self_v);
+	if (!self->m_lodList) {
+		PyErr_SetString(PyExc_ValueError, "KX_GameObject.lod: there is no lod levels for this gameobject");
+		return NULL;
+	}
+
+	return self->m_lodList->GetProxy();
 }
 
 PyObject *KX_GameObject::PyApplyForce(PyObject *args)
--- a/source/gameengine/Ketsji/KX_GameObject.cpp
+++ b/source/gameengine/Ketsji/KX_GameObject.cpp
@@ -796,15 +796,14 @@ void KX_GameObject::RemoveMeshes()
 	m_meshes.clear();
 }
 
-void KX_GameObject::UpdateLod(const MT_Vector3& cam_pos)
+void KX_GameObject::UpdateLod(const MT_Vector3& cam_pos, float lodfactor)
 {
 	if (!m_lodManager) {
 		return;
 	}
 
 	KX_Scene *scene = GetScene();
-	float distance2 = NodeGetWorldPosition().distance2(cam_pos);
-
+	const float distance2 = NodeGetWorldPosition().distance2(cam_pos) * (lodfactor * lodfactor);
 	KX_LodLevel *lodLevel = m_lodManager->GetLevel(scene, m_previousLodLevel, distance2);
 	const unsigned short level = lodLevel->GetLevel();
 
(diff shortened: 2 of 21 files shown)


source/gameengine/Ketsji/KX_GameObject.cpp
```diff
@@ -2063,7 +2063,7 @@ PyAttributeDef KX_GameObject::Attributes[] = {
 	KX_PYATTRIBUTE_RW_FUNCTION("debug",	KX_GameObject, pyattr_get_debug, pyattr_set_debug),
 	KX_PYATTRIBUTE_RO_FUNCTION("components", KX_GameObject, pyattr_get_components),
 	KX_PYATTRIBUTE_RW_FUNCTION("debugRecursive",	KX_GameObject, pyattr_get_debugRecursive, pyattr_set_debugRecursive),
-	KX_PYATTRIBUTE_RO_FUNCTION("lod", KX_GameObject, pyattr_get_lodlist),
+	KX_PYATTRIBUTE_RO_FUNCTION("lodManager", KX_GameObject, pyattr_get_lodmanager),
```
panzergame: KX_LodManager is a very nice name for the internal source but as for `filterManager`i don't know if the user like it or find too long ? @Maujoe: What's your opinion ?

source/gameengine/Ketsji/KX_GameObject.h
```diff
@@ -89,7 +89,7 @@ class KX_GameObject : public SCA_IObject
 	STR_String							m_text;
 	int									m_layer;
 	std::vector<RAS_MeshObject*>		m_meshes;
-	KX_LodList							*m_lodList;
+	KX_LodManager						*m_lodList;
```
panzergame: Minor issue : `m_lodList` -> `m_lodManager`

source/gameengine/Ketsji/KX_GameObject.h
```diff
@@ -800,12 +800,12 @@ class KX_GameObject : public SCA_IObject
 	/**
 	 * Set library of lod meshes
 	 */
-	void SetLodList(KX_LodList *lodList)
+	void SetLodList(KX_LodManager *lodList)
```
panzergame: `SetLodManager` and `GetLodManager`

source/gameengine/Ketsji/KX_LodLevel.cpp
```diff
@@ -0,0 +1,114 @@
...
+{
+	return GetName();
+}
+
+double KX_LodLevel::GetNumber()
+{
+	return -1.0;
+}
+
+STR_String &KX_LodLevel::GetName()
```
panzergame: Indeed the best way to name a lod is by its mesh name.

source/gameengine/Ketsji/KX_LodLevel.h
```diff
@@ -0,0 +1,65 @@
...
+class KX_LodLevel : public CValue
+{
+	Py_Header
+
+private:
+
+public:
+	KX_LodLevel();
+	virtual ~KX_LodLevel();
+
```
panzergame: @youle: We don't need `m_savedData` struct because here we copied data form blender and after that we modify nothing in the blender data.

source/gameengine/Ketsji/KX_LodManager.cpp
```diff
@@ -52,78 +51,49 @@ KX_LodList::KX_LodList(Object *ob, KX_Scene* scene, KX_BlenderSceneConverter* co
...
 			if (lod->flags & OB_LOD_USE_HYST) {
-				lodLevel.flags |= Level::USE_HYST;
+				lodLevel->m_flags |= KX_LodLevel::USE_HYST;
 			}
 
-			lodLevel.level = level++;
-			lodLevel.meshobj = BL_ConvertMesh(lodmesh, lodmatob, scene, converter, libloading);
-			lodLevel.distance = lod->distance;
-			lodLevel.hysteresis = lod->obhysteresis;
+			lodLevel->m_level = level++;
```
panzergame: I think you can use a constructor since `KX_LodLevel` is not anymore a struct. You just have to define a `int flag` 5 lines above. The constructor will look like:
`KX_LodLevel(level, meshobj, distance, hysteresis, flag)`

source/gameengine/Ketsji/KX_LodManager.cpp
```diff
@@ -52,78 +51,49 @@ KX_LodList::KX_LodList(Object *ob, KX_Scene* scene, KX_BlenderSceneConverter* co
...
-	const Level& lod = m_lodLevelList[level];
-	const Level& lodnext = m_lodLevelList[level + 1];
+	const KX_LodLevel *lod = m_lodLevelList[level];
+	const KX_LodLevel *lodnext = m_lodLevelList[level + 1];
 
 	float hysteresis = 0.0f;
 	// if exists, LoD level hysteresis will override scene hysteresis
-	if (lodnext.flags & Level::USE_HYST) {
-		hysteresis = lodnext.hysteresis;
+	if (lodnext->m_flags & KX_LodLevel::USE_HYST) {
```
panzergame: I prefer use getter here : `GetFlag`, `GetHysteresis`, `GetDistance`, `GetMesh`

source/gameengine/Ketsji/KX_LodManager.h
```diff
@@ -57,49 +44,36 @@ class KX_LodList: public CValue
...
+	KX_LodLevel *GetLevel(KX_Scene *scene, unsigned short previouslod, float distance2);
 
 	/// If it returns true, then the lod is useless then.
 	inline bool Empty() const
 	{
 		return m_lodLevelList.empty();
 	}
 
-	KX_LodList *AddRef()
+	KX_LodManager *AddRef()
```
panzergame: Sorry i forgotten that, indeed CValue have already these function in it. Then we can use CValue as base class instead of PyObjectPlus :/

panzergame:
The last commit is very great, thanks for the work.

source/gameengine/Ketsji/KX_GameObject.cpp
Maujoe: @panzergame I personally think the name is okay

youle31:
Ok, I made the suggested changes, except for the last ("Sorry i forgotten that, indeed CValue have already these function in it. Then we can use CValue as base class instead of PyObjectPlus :/")
I'm not sure that I understood. Inherit from PyObjectPlus and keep KX_LodManager *AddRef() is shorter than restore inheriting from CValue without to write a very small function AddRef, no?

youle31:
New test file for the complete python API:  http://www.pasteall.org/blend/42122

youle31:
@panzergame : When we create a list of objects like I did in the last commit, how do we do if the object is removed from the scene? We have to remove this object from the list I guess. I tested but it cause no crash. But when I do:
obj.endObject()
print(obj.lodManager.lodLevel) I have an error in the console (representation...). I don't know if it comes from the last commit or not.
My test file: http://www.pasteall.org/blend/42140

(just need to add a keyboard sensor and a endobject actuator and uncomment #print(own.lodManager.lodLevel))

This branch begins to grow and grow...

youle31:
So finally we can pass a list of KX_GameObject to a method. I didn't know that and this is very cool! I think we could implement other methods to pass lists of vertices for example in a method to recalculate normals of this vertices as BPR wanted.

I wrote the final commit message but this needs a professional review to merge in the master.

My test file for KX_Scene.setLodScale(pythonObjectList, scale): http://www.pasteall.org/blend/42162

Passing lists to python methods... Oh my god! this is great!

doc/python_api/rst/bge_types/bge.types.KX_Scene.rst
```diff
@@ -143,10 +143,15 @@ base class --- :class:`PyObjectPlus`
 
       :type: Vector((gx, gy, gz))
 
-   .. attribute:: lodGlobalScale
+   .. method:: setLodScale(pythonObjectList, scale)
```
panzergame: I'm not sure to understand the use here. If the user have a list of wanted objects why not iterating on it and modify the lodManager.distanceScale value ?

youle31: Because they did not want to iterate in a list with python. What I can understand. Here we iterate but in c++

panzergame: Why they don't want ?
```
for obj in list:
    obj.lodManager.distanceScale = scale
```

2 lines vs 1.

youle31: Because I guess it is slow. For i in range 1000: listOfTrees[i].obj.lodManager.distanceScale = scale is slow in python. I guess this is faster in c++.

panzergame: Is not too much faster than that in C++ because you still iterate in a python list, yes you avoid time on `lodManager.distanceScale`, but is very insignificant.

youle31: I test...

youle31:
as there is ge_lod_distance_scale now, can I close this branch and feature request?

source/gameengine/Rasterizer/RAS_MeshObject.h
```diff
@@ -39,6 +39,7 @@
```
panzergame: Please avoid add python dependencies in the rasterizer part. Part allowed to use python are Ketsji, GameLogic, VideoTexture, Converter, and Launcher (in some time).

panzergame:
@youle: Can you review the diff please ? : https://github.com/UPBGE/blender/pull/121/files

youle31:
Ok. I didn't understood what you meant with useMesh and useMaterial in KX_LodLevel. With the doc it is ok (I don't know if it will be used but nevermind)
About KX_LodLevel.m_distance RO, I'm not sure but you decide (Users could need to set individual level of detail distance...)
About const float distance2 = NodeGetWorldPosition().distance2(cam_pos) * (lodfactor * lodfactor);, why (lodfactor * lodfactor)? (same for distance2 *= (m_distanceFactor * m_distanceFactor);)
I'm sorry but I had a big confusion in my mind about distance /= lodfactor and I think finally this is better than "*=" (we just need to clamp factor to avoid zero division error). Sorry... I don't know what happened.
About KX_LodLevel.level, it's very weird as you call KX_LodLevel with his level id/number.
 About pyattr_get_mesh in KX_LodLevel, when we make KX_MeshProxy *meshproxy = new KX_MeshProxy(self->GetMesh());, it is automatically freed? (or I didn't see the code to free it)... And what happens if there is no mesh for this level of detail? No error?

I haven't tested yet. I'll do it later. Sorry again about "/=" thing. Big bug in my head...

panzergame:
`I unchecked useMesh for each level of detail. I returns original Cube mesh. So I guess this is right?`

Yes it's right, useMesh is to true only if you check the mesh otpion in lod panel which mean that you will not use the same mesh as the one in the object.

panzergame:
KX_LodLevel.distance is set in RO for the moment, because i don't know how to manage it with hysteresis and avoid to the user to unordered lod levels.

distance /= factor or distance2 /= (factor * factor) is the same as distance * 1 / factor or distance2 * (1 / factor * 1/ factor).

Then to summarize the user just have to pass to KX_LodManager.distanceFactor or KX_Camera.lodDistanceFactor 1 / zoom.

panzergame:
KX_MeshProxy is freed by python. It's why we use the function NewProxy instead of GetProxy. Indeed the function NewProxy take as argument a boolean which decide who will own the instance, if true the instance is owned by python, else by ourself.

PR #124, opened 2016-06-01
Lod distance scale

panzergame:
mplement KX_LodManager.distanceScale

The distance to camera is now mulplied by the the lod distance scale in
KX_LodManager::GetLevel, which by the side obligate us to use non-squared
distances.

The distance scale is currently simply named distanceScale, this name is
to debate, maybe distanceFactor ?.

Example file : http://www.pasteall.org/blend/42171

@youle31 : I was thinking to something like this for the distance scale, can you add your opinion on it, please ?
@Maujoe; you too, please ?

### DIFF
--- a/doc/python_api/rst/bge_types/bge.types.KX_2DFilter.rst
+++ b/doc/python_api/rst/bge_types/bge.types.KX_2DFilter.rst
@@ -9,7 +9,7 @@ base class --- :class:`BL_Shader`
 
    2D filter shader object. Can be alterated with :class:`BL_Shader`'s functions.
 
-   .. method:: setTexture(index, bindCode)
+   .. method:: setTexture(index, bindCode, samplerName)
 
       Set specified texture bind code :data:`bindCode` in specified slot :data:`index`. Any call to :data:`setTexture`
       should be followed by a call to :data:`BL_Shader.setSampler` with the same :data:`index`.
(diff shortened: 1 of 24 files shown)


Maujoe:
@panzergame I don't know how it worked before but scale the camera distance by the factor sounds like a plausible solution and yes distanceFactor is a clearer name.

youle31:
@panzergame : Hi!

- About setTexture, iirc SetUniform(loc, index); didn't work to set uniform in the 2D Filter. This is the reason why I used only SetUniformiv(loc, RAS_Uniform::UNI_INT, &index, (sizeof(int)), 1); (Maybe some changes have been made in the file since I looked and it works now... I'll test.

- About distance /= (lodfactor * lodfactor); or distance /= (lodfactor) in KX_GameObject::UpdateLod(const MT_Vector3& cam_pos, float lodfactor), finally I think we should use a linear function because http://www.pasteall.org/blend/42178
like distance -= lodFactor instead to have a linear zoom....

panzergame:
Sorry wrong manip :/

youle31:
Allright for setTexture, this works. I don't know why but before: #ifdef SORT_UNIFORMS returned false whereas now it is true so SetUniformiv(loc, RAS_Uniform::UNI_INT, &index, (sizeof(int)), 1); is used.

youle31:
I lost the framapad we made about API: Do you still have the link? Sorry.

PR #125, opened 2016-06-01
Ge lodoptimization

youle31:
Can't we avoid a for loop like that?

(there's another for loop in DrawDebug... I don't know if it can also be avoided)

(see last commit only)

edit: with 16000 cubes increases from 5fps (24 -> 29 fps) (20,8%)

### DIFF
--- a/doc/python_api/rst/bge_types/bge.types.KX_Camera.rst
+++ b/doc/python_api/rst/bge_types/bge.types.KX_Camera.rst
@@ -27,6 +27,12 @@ base class --- :class:`KX_GameObject`
 
       :type: float
 
+   .. attribute:: lodFactor
+
+      The camera's level of detail factor value.
+
+      :type: float
+
    .. attribute:: fov
 
       The camera's field of view value.
--- a/doc/python_api/rst/bge_types/bge.types.KX_GameObject.rst
+++ b/doc/python_api/rst/bge_types/bge.types.KX_GameObject.rst
@@ -496,6 +496,15 @@ base class --- :class:`SCA_IObject`
 
       :type: int
 
+   .. attribute:: lodManager
+
+      Return the :class:`KX_LodManager` of this KX_GameObject.
+      We need to access lodManager to set attributes of levels of detail
+      for this KX_GameObject. Example:
+      KX_GameObject.lodManager.lodLevel[0].distance = 25.0.
+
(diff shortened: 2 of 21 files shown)


panzergame:
The cost of the loop itself (without its content) is not so big, maybe 5-10 instructions per iterations. For the moment is not grave. @youle31 : Instead can you investigate on these lines:
```
	// Handle dupligroups
	if (m_pInstanceObjects) {
		for (CListValue::iterator it = m_pInstanceObjects->GetBegin(), end = m_pInstanceObjects->GetEnd(); it != end; ++it) {
			KX_GameObject *instob = (KX_GameObject *)*it;
			instob->UpdateLod(cam_pos, lodfactor);
		}
	}
```
Objects in groups are not in objects scene list ? If it's the case then these lines are useless.
An other way to optimize the schmilblique is to multi task the function KX_GameObject:: UpdateLod but then set a mutex (or spin lock) before changing of mesh. But i'm not sure if it's great to use many task e.g 2000 tasks, if not then create 20 task for 100 objects ect…

youle31:
optimization of 20.8% with 16000 cubes on my computer without script

panzergame:
Well, loop cost most that what i though :P

What 20.8 %, how do you get how many ?? :).

youle31:
5 * 100 / 24

panzergame:
I mean how just merging a loop can save too much performance ? I'm surprised, it's nice.

youle31:
I guess a loop threw 16000 objects has a certain cost

panzergame:
I found the reason: we updated lod for every logic frame and not only for every render frame, and as default we run as much as possible logic frame to follow the user FPS (with the logic maximum frame limit sure).
The patch http://www.pasteall.org/70166 fix it.

@youle31 : Thanks for the issue report, nice hit :+1:

youle31:
panzergame: your patch increase from 4 fps, mine of 5 fps, so I have 33/34 fps  now with both patches...

youle31:
Oh no wait... I'll make other tests tomorrow... But the for loop has an impact. It's sure

youle31:
http://www.pasteall.org/pic/103846 Master
http://www.pasteall.org/pic/103847 Master with your commit
http://www.pasteall.org/pic/103848 Master with my commit

panzergame:
I commited the first patch for the moment, I don't very know about merging the loop…

PR #126, opened 2016-06-02
Common launcher and event system refactor.

panzergame:
This branch introduce a new event system which allow the user to detect event raised during a frame in the right order.

The input device for GHOST is located in Ghost/

As a common was needed to reduce code and that a event system was the major barrier this was merged in the same branch.

The launcher files are in a new directory Launcher/

LA_Launcher is the base class inherited to two classes LA_BlenderLauncher and LA_PlayerLauncher.

Compatibility :
if events[KEY] == status --−> if status in events[KEY].status/queue

@youle31, @DCubix : For the moment as the description is not very good can you just test if there's no major compatibility issues or bugs ?


youle31:
in this file: http://www.pasteall.org/blend/42195 modified to fit the new event system (keyboard controls ZSQD), I can move the sun with some keyboard keys whereas we normally move the sun with drag/drop left mouse button.

Also, the drag and drop with mouse doesn't work correctly.

To make many tests with complex files, we really need a system of backward compatibility for events because if there are many scripts in a/some .blend that handle events, it will be too hard to find and modify all this scripts.

And if an user wants to switch to upbge, if he opens his .blend made with BF Blender and if he realize that his file doesn't work directly, there are some chances that he will be discouraged...

panzergame:
About compatibility the "event[KEY] == status" is still supported but deprecated, only the case with "if event" is not anymore supported because i can't override something here.

youle31:
Can't we say that if we write: if keyboard.events[bge.events.ZKEY] this is the same as if we have written if bge.logic.KX_INPUT_ACTIVE in keyboard.events[bge.events.ZKEY].status or keyboard.events[bge.events.ZKEY].queue in (bge.logic.KX_INPUT_JUST_ACTIVATED, bge.logic.KX_INPUT_JUST_RELEASED) ? (to make the new system compatible with the old?

panzergame:
No because:
```
event = None

if condition:
    event = keyboard.events[KEY]
elif condition2:
    event = keyboard.events[KEY2]

if event:
    if KX_INPUT_ACTIVE in event.status:
       move()
```

The `if event:` should react how ? Here the natural behaviour is to return True if is it not None, and in this case is useful. It's disappointed but i think we can't do something for this compatibility case…

youle31:
I don't understand. Why can't we do: if keyboard.events[KEY]: move()
We just have to return true if there is something in keyboard.events[KEY].status or keyboard.events[KEY].queue()...

panzergame:
`keyboard.events[KEY]` return a `SCA_InputEvent` then how to differ a call to test if the event contains a valid status, or that we test that a `SCA_InputEvent` is not None ?

```
if keyboard.events[KEY]:
    move()

# Same has
event = keyboard.eents[KEY]
if event:
  move()

#Then the both test only test if the event for the given case is not None,
# of course this test should not be done as it since every keys enumerations
# in bge.events give a non-None SCA_InputEvent. But for complex code
# were the input event can differ it can be useful to test if it is not None.
```

panzergame:
Mouse bug fixed.

youle31:
In docks level the camera moves in all directions (embedded)...
At the begining, the events system refactor goal was to fix a minor bug (if you had 2 keys pressed in the same frame, only the first was taken into account), no? I'm sorry to insist but I think if we can't have backward compatibility for events, it will be a "kind of disaster" for the upbge as there are many .blend that use python scripts that handle events. Je suis désolé, c'est pas pour te provoquer ou parce que j'aurais une dent contre toi que je dis ça. C'est juste que par exemple, pour le background, je me suis rendu compte que c'était une erreur de casser la compatibilité (c'était chiant de réécrire les scripts pour mettre horizon à la place de background). Je me dis juste que si un nouvel utilisateur veut tester l'upbge, il va le tester avec ses .blend et s'il voit que ça marche pas direct, y'a de fortes chances qu'il laisse tomber. Je pense que ça peut casser l'envie des utilisateurs de se tourner vers l'upbge... Après, ton système est sûrement mieux que l'original, mais on pourrait pas garder les 2 systèmes en parallèle ou trouver une astuce pour la rétrocompatibilité?
Désolé de remettre ça sur le tapis... C'est juste que ça me paraît important... On va trouver une solution!! haha

panzergame:
The `if event` is now solved. But it means that now the user for new code must be explicit and use:
```
if KX_INPUT_ACTIVE in event.status:

or 

if event is not None:
```

Anyway we printed a deprecated warning :
`Method if event: is deprecated, please use if KX_INPUT_ACTIVE in event.status:  or  if event is not None: instead.`

youle31:
Awesome. Thanks very much! Sorry to have insist...

pqftgs:
Awesome!  Input misfires have been a huge problem.
Could I suggest a more user-friendly API?
```
keyboard.pressed(WKEY)  # True when initially pressed
keyboard.released(WKEY) # True when initially released
keyboard.held(WKEY)     # True when the button is down
```

panzergame:
Thanks @pqftgs.

About the API i'm not sure because the user will want to count status events and value. Then we give it a simple way to do a simple test `enum in event.list`.

Sure it's a bit hard API, but we can provide an example to the user which define the function you describe above in python only.

pqftgs:
Seems like you're expecting users to implement their own event-driven framework on top of the Python API.  Check Unity (or any other engine) and they pull it off quite nicely.  This could be a lot simpler without losing any functionality.

Generally if a player manages to press a button twice in one frame (< 0.02 seconds) it's probably a glitchy button and doesn't really matter if only one press is detected.

PR #129, opened 2016-06-04
UPBGE: Implement game object bounding volume box python API.

panzergame:
The user wasn't able to set or just read the AABB previously, this can be useful
for "predefined AABB" for animated objects or at least to know the volume of the
mesh.

The API now exposes a KX_BoundingBox python class accesible in KX_GameObject.bounds.
This class is temporary created when accesing obj.bounds, none logic is proceed in
it for internal usage.
The class contains the following attributes:
- min (Vector3, rw),
- max (Vector3, rw),
- center (Vector3, ro),
- radius (float, ro),
- autoUpdate (bool, rw)

### DIFF
--- a/doc/python_api/rst/bge_types/bge.types.KX_BoundingBox.rst
+++ b/doc/python_api/rst/bge_types/bge.types.KX_BoundingBox.rst
@@ -28,13 +28,13 @@ base class --- :class:`PyObjectPlus`
 
    .. attribute:: min
 
-      The minimal point of the bounding box.
+      The minimal point in x, y and z axis of the bounding box.
 
       :type: :class:`mathutils.Vector`
 
    .. attribute:: max
 
-      The maximal point of the bounding box.
+      The maximal point in x, y and z axis of the bounding box.
 
       :type: :class:`mathutils.Vector`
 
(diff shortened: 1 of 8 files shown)


panzergame:
Test file : http://www.pasteall.org/blend/42255

youle31:
Hi! I made my own test file: http://www.pasteall.org/blend/42270

- I find the controls of the bounding box a bit weird. Proposition: make top/bottom/left/right
- I tested on a static object... when autoUpdate is set to False, the bounding box shouldn't keep the same dimensions? In my test file it is updated with the scale (animation keyframes) of the object.

Other that, sounds ok... It's a little test. I didn't tested collisions.

panzergame:
The bounds here are for culling not for physics.

panzergame:
The user can scale the box by:
```
bounds.max = bounds.max * scale
bounds.min = bounds.min * scale
```

youle31:
Ah yes I forgot...

panzergame:
I can rename `bounds` to `cullingBox` to avoid misunderstood, @youle31, @DCubix, and everyone, What do you prefer ? For the moment i don't have a strong opinion on this.

youle31:
Yes for culling box.
It's weird to split box with box.min, box.max...
Why not box.top bottom left right front back

PR #130, opened 2016-06-08
replace RGBA8 with RGBA

youle31:
I have a gain of performances in this test file: http://www.pasteall.org/blend/42272

Could someone confirm?

We could also use glCopyTexSubImage2D(GL_TEXTURE_2D, 0, 0, 0, textureleft, texturebottom, texturewidth, textureheight); instead of glCopyTexImage2D


panzergame:
How much it decreases render time ?

youle31:
I had ~100 more fps / 1500. But it change each time I open blender...

PR #137, opened 2016-06-13
UPBGE: Import Benoit Bolsee VideoTexture commit into upbge

youle31:
notes: in RAS_OpenglRasterizer.cpp I included KX_Globals to get canvas
(KX_GetActiveEngine->GetCanvas()) for
RAS_OpenGLRasterizer::CreateOffScreen function.

I noticed a performances decrease in water demo file.

I made no test of the new features for the moment.

### DIFF
--- a/intern/moto/include/MT_Matrix4x4.h
+++ b/intern/moto/include/MT_Matrix4x4.h
@@ -131,6 +131,16 @@ public:
     }
 
 	/**
+	 * Scale the rows of this matrix with x, y, z, w respectively.
+	 */
+		void tscale(MT_Scalar x, MT_Scalar y, MT_Scalar z, MT_Scalar w) {
+		m_el[0][0] *= x; m_el[1][0] *= y; m_el[2][0] *= z; m_el[3][0] *= w;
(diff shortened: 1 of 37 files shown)


intern/moto/include/MT_Matrix4x4.h
```diff
@@ -131,6 +131,16 @@ class MT_Matrix4x4 {
     }
 
 	/**
+	 * Scale the rows of this matrix with x, y, z, w respectively.
```
panzergame: To reduce cleanup work you can commit this modification in a independant commit

youle31:
As I have different performance results than the author of the patch, could someone test this files with this branch and the master to see if there is a performance difference (water with fbo works only with this branch). Also, if you want to make a complete test, you can compare the framerate with the one in BF Blender 2.77a.

If you can post your results it would be very nice. Thanks!

Files: http://www.pasteall.org/blend/42429 (without fbo)
http://www.pasteall.org/blend/42430 (with fbo)

panzergame:
Sorry for delay:

Without FBO: 6.7ms per frame
With FBO: 6.7ms per frame

youle31:
and in BF blender 2.77a without fbo?

panzergame:
The benoit's commits was merged, it request now a big cleanup.

PR #138, opened 2016-06-16
Python interpreter

panzergame:
This patch implement a little python interpreter instanced by pressing Ctrl+Shift+Alt+T and terminated in console with Ctrl+D for linux and Ctrl+Z+Enter in windows.

The interpreter gives access to everything, it uses auto-completion, bge module is already imported  and a "scene" variable is already defined to the current scene.

An example video is available to : https://www.youtube.com/watch?v=wibPpotzhmU

Topic:
   Default variables: 
       What "defines" expose to the user ? extra functions ?
       "scene" is half useful because the interpreter is instanced after a logic frame, then the current scene is just the last scene.


panzergame:
merged in 0.0.9

PR #140, opened 2016-06-21
Simplify UpdateFrom function used in softbody and skin update

youle31:
test file: http://www.pasteall.org/blend/42460

### DIFF
--- a/source/gameengine/Rasterizer/RAS_DisplayArray.cpp
+++ b/source/gameengine/Rasterizer/RAS_DisplayArray.cpp
@@ -48,31 +48,24 @@ int RAS_DisplayArray::GetOpenGLPrimitiveType() const
 
 void RAS_DisplayArray::UpdateFrom(RAS_DisplayArray *other, int flag)
 {
-	if (flag & RAS_MeshObject::TANGENT_MODIFIED) {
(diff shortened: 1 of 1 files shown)


youle31:
This improves performances on my computer (10% iirc)

source/gameengine/Rasterizer/RAS_DisplayArray.cpp
```diff
@@ -48,31 +48,24 @@ int RAS_DisplayArray::GetOpenGLPrimitiveType() const
...
-		}
-	}
-	if (flag & RAS_MeshObject::UVS_MODIFIED) {
-		for (unsigned int i = 0; i < other->m_vertex.size(); ++i) {
-			for (unsigned int uv = 0; uv < 8; ++uv) {
-				m_vertex[i].SetUV(uv, MT_Vector2(other->m_vertex[i].getUV(uv)));
+			switch (flag)
+			{
+			case RAS_MeshObject::POSITION_MODIFIED:
+				m_vertex[i].SetXYZ(MT_Vector3(other->m_vertex[i].getXYZ()));
```
panzergame: you should use `break` to avoid doing all the code after. For example a vertex modified in position will be updated in position, normal, tangent and uv here without `break`.

BluePrintRandom: So this will improve moving vertex in mesh speed / gpu update as well as
armature skin and soft body?

panzergame:
Ehhh, why it was closed ? I just delayed to review it.

PR #143, opened 2016-06-29
Realtime cube maps.

youle31:
test file: http://www.pasteall.org/blend/42547

Warning: It's far from ready.


BluePrintRandom:
<3 <3 <3

panzergame:
Updated test file : http://www.pasteall.org/blend/42581.
Screenshot : http://www.pasteall.org/pic/show.php?id=104743.

panzergame:
I know that KX_* is used in RAS_OpenGlRasterizer, but this is deprecated. It's why we used two levels for cube maps : `RAS_CubeMap[Manager]` and `KX_CubeMap[Manager]`. `RAS_CubeMap[Manager]` contains as much as possible all rasterizer calls, e.g `GPU_texture_*` `GPU_frame_buffer_*`, `RAS_IRasterizer::SetProjectionMatrix` and `KX_CubeMap[Manager]` contains all Ketsji calls, e.g `KX_GameObject::NodeGetWorldPosition`, `KX_Scene::RenderBuckets`, `KX_Camera::SetProjectionMatrix`. I know that it make the code more complicated, but it will organize our mind by separating the render part to the scene/camera/object part.

@youle31 : I'm sure you can do what i describe, good luck :)

youle31:
I prefer to let you fix it because I don't understand what you want exactly.... Like I said it's just ideas but my commits can be removed. If you want to do it on my computer, I'll see what you want to avoid.

youle31:
the test file to test last commit: http://www.pasteall.org/blend/42731 the objects with cubemap are inverted wtf?

youle31:
oh my stupidness. I set the face matrix before setting the viewmatrix from facematrix..... Pfffffffff. So I got it.

youle31:
So in the last commit about culling I set the clipend of the KX_CubeMapManager m_camera to 10.0f for testing purpose. I tested with this test file: http://www.pasteall.org/blend/42763

test protocol:

Suzanne origin is at ten Blender units from the sphere. As its nose is slightly at less than 10 blender units from the sphere you can see her.

If you move suzanne far from the sphere of 1 BU you can't see her anymore reflected in the sphere. And this for each face.

To make the test easier, I parented Suzanne to an empty at the position of the sphere (I could have parented Suzanne directly to the sphere but I'm stupid). So you can rotate empty of 90° on 1 axis to test each face culling.

So now we have working normal culling for m_camera each time a face is rendered.

youle31:
Ready to be reviewed and refactored :P https://www.youtube.com/watch?v=OWJKTpg2rb4
(Finally I think the option to render background or not is not needed)
I changed the default size of the generated images to be compatible with cubemaps. I also changed default mode to GRID instead of Blank image. I found it more funny. I set a variable to avoid to execute some functions in the rendering process while rendering cubemaps: for example, no need to process lighting or to update things that have already been updated in the previous render. And I did other things... I don't remember.

youle31:
@panzergame : Hello! I found the cause of this bug: https://drive.google.com/file/d/0B8EITuogwZr6c1ZtOFBHc284OTA/view?usp=sharing with cubemaps: The new materials have default constant texture for material activated (the grid texture is visible on the cube when we render cubemaps).

As TwisterGE says he has slow down in the viewport when some constant options are activated (dynamic uniforms are maybe faster in the viewport) and as it makes sometimes bugs hard to find (for us and maybe also sometimes for the users), if you want we could eventually rediscuss make constant as default for newly added materials or find an alternative.

youle31:
I totally disagree with your last commit. afk

panzergame:
I reverted 00a695a.

The abyss effect will be still here but the user can set a priority order to the cube map object to control the abyss.

E.g:
In a scene with a cube and a sphere. 
If the cube compute its cube map first, then its cube map will contain the reflection of the sphere without the reflection of the cube on the sphere. And the cube map of the sphere will contain the reflection of the cube plus the reflection of the sphere on the cube.

Else if the sphere compute its cube map first, then its cube map will contain the reflection of the cube without the reflection of the sphere on the cube. And the cube map of the cube will contain the reflection of the sphere plus the reflection of the cube on the sphere.

Well it's a pretty complicated case.

youle31:
Yes i understand. I need to take a break. Cu

BluePrintRandom:
Thank you guys, I don't know if I say it enough, but I am very happy with
your progress....

lordloki:
Just a though. Is it possible to not render the cubemap each frame (i.e each 2 or 3 frames)? This way we could save some computational power and I think the eye wouldn't catch the difference.

youle31:
Hi, I think (not tested) that if we update the cubemap each 3 frames in the sources, this can cause framerate increase/decrease that will make performance tests difficult, but this is a good idea to let the user choose when he wants to update the cubemap with a simple python API.

I had better performance with GL_LINEAR filter mode than with GL_NEAREST mode for both min and mag filters. I don't know why.

About lodbias, it could be an option in the UI (replace slider with "factor button" iirc -> button to enable/disable+slider) even if it has a performance impact.

BluePrintRandom:
(quick question - is Cube maps working on objects with a mix of flat and
curved surfaces now?)

Thank you youle!

youle31:
@BluePrintRandom : Hi, no, for the moment, all objects with realtime cubemap option enabled will be rendered smooth: http://antongerdelan.net/opengl/cubemaps.html ("Common mistakes") -> the shader works only when normals are "interpolated"

BluePrintRandom:
that is ok, I don't need flat mode :D

does a cube with 2 simple and 1 smooth subdivision render correctly now?

BluePrintRandom:
[CyclesScratches3.zip](https://github.com/UPBGE/blender/files/398304/CyclesScratches3.zip)

here is a test file for cube map reflections, the texture Refit.png is the texture to be replaced with the cube map reflection shader

youle31:
Waiting for a better solution, to have (more or less) accurate reflections on a cube, you can apply to it a bevel modifier with 0.001 width... This "changes" the mesh normals and the way the shader map the reflection texture. Maybe there is a better way to do this working on mesh normals...

source/gameengine/Ketsji/KX_CubeMap.cpp
```diff
@@ -0,0 +1,248 @@
...
+
+PyMethodDef KX_CubeMap::Methods[] = {
+	KX_PYMETHODTABLE_NOARGS(KX_CubeMap, update),
+	{NULL, NULL} // Sentinel
+};
+
+PyAttributeDef KX_CubeMap::Attributes[] = {
+	KX_PYATTRIBUTE_RW_FUNCTION("viewpointObject", KX_CubeMap, pyattr_get_viewpoint_object, pyattr_set_viewpoint_object),
+	KX_PYATTRIBUTE_BOOL_RW("autoUpdate", KX_CubeMap, m_autoUpdate),
+	KX_PYATTRIBUTE_INT_RW("ignoreLayers", 0, (1 << 20) - 1, true, KX_CubeMap, m_ignoreLayers),
```
youle31: Hi, I think there is an issue with this: http://pasteall.org/blend/index.php?id=43238

source/gameengine/Rasterizer/RAS_CubeMap.cpp
```diff
@@ -0,0 +1,161 @@
...
+	GPU_texture_free(m_cubeMapTexture);
+	GPU_free_image(m_texture->GetImage());
+
+	for (unsigned short i = 0; i < 6; ++i) {
+		GPU_framebuffer_free(m_fbos[i]);
+		GPU_renderbuffer_free(m_rbs[i]);
+	}
+}
+
+void RAS_CubeMap::BeginRender()
```
youle31: Can we remove it now?

source/gameengine/Ketsji/KX_CubeMapManager.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+	/* We hide the gameobject in the case backface culling is disabled -> we can't see through
+	 * the object faces if the camera is inside the gameobject
+	 */
+	gameobj->SetVisible(false, true);
+
+	/* For Culling we need first to set the camera position at the object position */
+	m_camera->NodeSetWorldPosition(pos);
+
+	/* For each cubeMap, environment texture clipend value is not necessarly the same
```
youle31: We can update the comment.

source/gameengine/Rasterizer/RAS_CubeMap.cpp
```diff
@@ -0,0 +1,161 @@
...
+	m_useMipmap = (env->flag & ENVMAP_MIPMAP) != 0;
+
+	m_cubeMapTexture = m_texture->GetGPUTexture();
+	// Increment reference to make sure the gpu texture will not be freed by someone else.
+	GPU_texture_ref(m_cubeMapTexture);
+
+	if (!m_useMipmap) {
+		// Disable mipmaping.
+		GPU_texture_bind(m_cubeMapTexture, 0);
+		GPU_texture_filter_mode(m_cubeMapTexture, false, false);
```
youle31: when you use GPU_texture_filter_mode(m_cubeMapTexture, false, true); the reflection is less aliased when you look near a the cubemap. But I have a little performance decrease compared to GL_LINEAR filter. Weird last time I checked it was the inverse. Linear could be an option?

source/gameengine/Rasterizer/RAS_CubeMap.cpp
panzergame: Why ? it freed the frame buffers and render buffers created by the cube map, also freeing the texture/image just ask to the next user of this texture to reload it.

source/gameengine/Ketsji/KX_CubeMapManager.cpp
panzergame: Feel free to push a correction.

source/gameengine/Rasterizer/RAS_CubeMap.cpp
panzergame: the last argument of GPU_texture_filter_mode is to enable mipmapping, it's why you get nicest result with this argument to true. But if you enable mipmap you have to regenerate mipmaps at every render end, which is done if the user enable the UI option for mipmapping.

youle31: It's not needed to regenerate mipmap each frame when you set the filters min and mag when you set the texture at the beginning (in the contructor). I'll try to find a way to make it as an option.

panzergame:
ge_cubemap is enought advanced to be reviewed. The current bugs to fix are related to blenderplayer.

source/gameengine/Rasterizer/RAS_CubeMap.cpp
```diff
@@ -66,10 +60,16 @@ static const MT_Matrix4x4 backFaceViewMat(
...
+	0.0f, 0.0f, 1.0f, 0.0f,
 	0.0f, -1.0f, 0.0f, 0.0f,
-	-1.0f, 0.0f, 0.0f, 0.0f,
+	0.0f, 0.0f, 0.0f, 1.0f);
+
+static const MT_Matrix4x4 leftFaceViewMat(
+	1.0f, 0.0f, 0.0f, 0.0f,
+	0.0f, 0.0f, -1.0f, 0.0f,
+	0.0f, 1.0f, 0.0f, 0.0f,
 	0.0f, 0.0f, 0.0f, 1.0f);
```
youle31: Hi, for me it doesn't work anymore...

panzergame: The commit message said that render matrix are fixed but not culling matrix. You should not have any issue with world background rendering but culling maybe be wrong.

source/gameengine/Rasterizer/RAS_CubeMap.cpp
```diff
@@ -0,0 +1,197 @@
...
+}
+
+void RAS_CubeMap::AttachTexture()
+{
+	// Increment reference to make sure the gpu texture will not be freed by someone else.
+	GPU_texture_ref(m_gpuTex);
+
+	for (unsigned short i = 0; i < NUM_FACES; ++i) {
+		m_fbos[i] = GPU_framebuffer_create();
+		m_rbs[i] = GPU_renderbuffer_create(GPU_texture_width(m_gpuTex), GPU_texture_height(m_gpuTex),
```
youle31: Hi! Do we really need renderbuffers? In most examples I see on the net for cubemaps there is no renderbuffer.

source/gameengine/Rasterizer/RAS_CubeMap.cpp
```diff
@@ -0,0 +1,197 @@
...
+										   0, GPU_HDR_NONE, GPU_RENDERBUFFER_DEPTH, NULL);
+
+		GPU_framebuffer_texture_attach_target(m_fbos[i], m_gpuTex, cubeMapTargets[i], 0, NULL);
```
youle31: why is this better to have one fbo per face with only one texture attached to each fbo? iirc before we used one fbo and GPU_framebuffer_texture_attach_target(m_fbo, m_gpuTex, cubeMapTargets[i], 0, NULL); when we bound the face.

source/gameengine/Ketsji/KX_CubeMapManager.cpp
```diff
@@ -0,0 +1,157 @@
...
+		const MT_Matrix4x4& proj = rasty->GetFrustumMatrix(-clipstart, clipstart, -clipstart, clipstart, clipstart, clipend, 1.0f, true);
+		cubeMap->SetProjectionMatrix(proj);
+		cubeMap->SetInvalidProjectionMatrix(false);
+	}
+
+	// Else we use the projection matrix stored in the cube map.
+	rasty->SetProjectionMatrix(cubeMap->GetProjectionMatrix());
+	m_camera->SetProjectionMatrix(cubeMap->GetProjectionMatrix());
+
+	cubeMap->BeginRender();
```
youle31: There is not a problem here? Do we really need to CheckValidTexture each frame? When I remove that I have a performance increase.

doc/python_api/rst/bge_types/bge.types.KX_CubeMap.rst
```diff
@@ -0,0 +1,74 @@
...
+
+      Enable the cube map to render the scene.
+
+      :type: boolean
+
+   .. attribute:: ignoreLayers
+
+      The layers to ignore when rendering the cube map.
+
+      :type: bitfield
```
youle31: A code snippet could be fine here I think.

youle31:
For me, according to my code abilities,  all seems right. The comments I made are more questions (except for the code about bits for ignoreLayers)

source/gameengine/Rasterizer/RAS_CubeMap.cpp
panzergame: Render buffers are faster for old GPUs and a least made no difference for very recent GPUs. We use RBOs for the same reasons for the global FBO rendering.

source/gameengine/Rasterizer/RAS_CubeMap.cpp
panzergame: Switching of FBO is faster than detaching and attaching a texture to a FBO.

source/gameengine/Ketsji/KX_CubeMapManager.cpp
panzergame: It's needed to every frame check if the gpu texture used by the material is the same as the gou texture we are rendering to. This case happen when mipmapping was disabled or enabled globally (e.g blenderplayer launch, python api).

doc/python_api/rst/bge_types/bge.types.KX_CubeMap.rst
panzergame: I will modify the code example at the begining of the documentation.

panzergame:
Merged in master.

PR #144, opened 2016-07-05
use glDrawRangeElements for big display array lists

youle31:
in VA mode, I had bottleneck when drawing big display array lists.
Using 2 draw calls instead of 1 fixes the issue. I don't know if the usage
I make of glDrawRangeElements is correct but it seems to work fine. I make
a pull request to join a test file.

test file: http://www.pasteall.org/blend/42607

This concerns only Vertex Array mode.

steps: In Master:
- Run test file, press P (the spheres (composed only of quads) are rendered at 1200 fps on my computer.
- Run bpy script to apply modifiers, then press P. The spheres are rendered at 800 fps on my computer ---> bottleneck

In this branch: Do the same tests and you have the same fps with or without modifiers if all is right (on my computer it works). I didn't noticed performance decrease on other test files in VA.

Source: http://www.songho.ca/opengl/gl_vertexarray.html (I didn't understood range 0, 6 then 1, 7).

### DIFF
--- a/source/gameengine/Rasterizer/RAS_OpenGLRasterizer/RAS_StorageVA.cpp
+++ b/source/gameengine/Rasterizer/RAS_OpenGLRasterizer/RAS_StorageVA.cpp
@@ -30,6 +30,7 @@
 #include "RAS_DisplayArray.h"
 #include "RAS_MaterialBucket.h"
 #include "RAS_MeshUser.h"
+#include <algorithm>    // std::min_element, std::max_element
 
 #include "glew-mx.h"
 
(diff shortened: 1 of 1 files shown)


panzergame:
I have nice speed up as you said:
Normal:
    Total: 29.1ms
    Rasterizer: 0.22ms
    GPU Latency: 28.54ms

RangeElements:
    Total: 17.9ms
    Rasterizer: 0.23ms
    GPU Latency: 17.29ms

@youle, do you find why it's faster ? Also did it works with VBO ?

source/gameengine/Rasterizer/RAS_OpenGLRasterizer/RAS_StorageVA.cpp
```diff
@@ -186,8 +187,43 @@ void RAS_StorageVA::IndexPrimitives(RAS_MeshSlot *ms)
 	else
 		glColor4f(0.0f, 0.0f, 0.0f, 1.0f);
 
-	// here the actual drawing takes places
-	glDrawElements(array->GetOpenGLPrimitiveType(), array->m_index.size(), GL_UNSIGNED_INT, indexarray);
+	/* For big display lists, using only one draw call with glDrawElements can generate a bottleneck
+	 * in Vertex Array mode, and cause a performance decrease. So we split the array list into
+	 * several parts and make one draw call for each part with glDrawRageElements */
+	if (array->m_index.size() > 50000) {
```
panzergame: Maybe you can add a check if the VA is using DL, if is not the case don't compute parts.

youle31:
@panzergame : you're right. I added if displaylist as condition to use glDrawRangeElements

youle31:
Finally after tests, I realized that the bottlneck only happened when there was displayed (not culled) more than GL_MAX_ELEMENTS_INDICES on the screen. On my computer, GL_MAX_ELEMENTS_INDICES is more than 1 000 000 indices. So the bottlneck is very rare. If we wanted to use glDrawRange elements, each frame we would have to compute total of all objetcs displayed indices (avoiding those which are not culled). This computation could be heavy so this is a bad idea.

If we ALWAYS use glDrawRangeElements instead of glDrawElements, we have a slight performance decrease in some files.

So better to avoid completely glDrawRangeElements in bge (unless one day we have an exceptionnal case) or we could use glDrawRangeElements only for meshes with indices > GL_MAX_ELEMENTS_INDICES (but this almost never happens I guess)

So I close the pull request and the branch.

PR #151, opened 2016-07-11
Render text object as mesh.

panzergame:
Rendering text object as mesh is only to reuse the sort of alpha objects.

For the moment all text objects use a common global material which is an instance of KX_TextMaterial.

@youle31 , @DCubix, @lordloki: Can you please review this patch a bit ? I'm very uncertain about the architecture.


source/gameengine/Converter/BL_BlenderDataConversion.cpp
```diff
@@ -1141,9 +1141,7 @@ static KX_GameObject *gameobject_from_blenderobject(
 		/* font objects have no bounding box */
 		gameobj = new KX_FontObject(kxscene,KX_Scene::m_callbacks, rendertools, ob, do_color_management);
 
-		/* add to the list only the visible fonts */
-		if ((ob->lay & kxscene->GetBlenderScene()->lay) != 0)
-			kxscene->AddFont(static_cast<KX_FontObject*>(gameobj));
+		kxscene->AddFont(static_cast<KX_FontObject*>(gameobj));
```
panzergame: As we render texts as mesh, only texts from active object are rendered, then this workaround which was broken for added object is removed.

source/gameengine/Ketsji/KX_TextMaterial.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+/** \file gameengine/Ketsji/KX_TextMaterial.cpp
+ *  \ingroup ketsji
+ */
+
+#include "KX_TextMaterial.h"
+
+#include "DNA_material_types.h"
+
+static KX_TextMaterial textMaterial = KX_TextMaterial();
```
panzergame: Using a static global material is to avoid allocate and delete the text material and then avoid bugs with libloading.

source/gameengine/Rasterizer/RAS_BucketManager.cpp
```diff
@@ -332,6 +332,10 @@ RAS_MaterialBucket *RAS_BucketManager::FindBucket(RAS_IPolyMaterial *material, b
 		else
 			m_buckets[useinstancing ? SOLID_SHADOW_INSTANCING_BUCKET : SOLID_SHADOW_BUCKET].push_back(bucket);
 	}
+	if (material->IsText()) {
+		m_buckets[TEXT_BUCKET].push_back(bucket);
```
panzergame: In case of libload there will be two material bucket for text material, this is not dangerous.

source/gameengine/Rasterizer/RAS_IRasterizer.h
```diff
@@ -379,6 +379,9 @@ class RAS_IRasterizer
 	 * IndexPrimitives_3DText will render text into the polygons.
 	 */
 	virtual void IndexPrimitives_3DText(class RAS_MeshSlot *ms, class RAS_IPolyMaterial *polymat) = 0;
+
+	/// Render text mesh slot using BLF functions.
+	virtual void IndexPrimitivesText(RAS_MeshSlot *ms) = 0;
```
panzergame: I'm very not sure about the name.

DCubix:
Looks fine to me :D

source/gameengine/Ketsji/KX_FontObject.cpp
```diff
@@ -113,6 +117,85 @@ void KX_FontObject::ProcessReplica()
...
+			GetOpenGLMatrix();
+		}
+
+		// Allow for some logic brick control
+		if (GetProperty("Text")) {
+			m_text = split_string(GetProperty("Text")->GetText());
+		}
+
+		// update the animated color
+		GetObjectColor().getValue(m_color);
```
youle31: The patch is not finished? A normal material will be added in blender data conversion ConvertMesh to have diffuse, specular etc? Like for other meshes? Or is it impossible?

youle31:
I've not a global view on the render mesh process, but if it follows what is done for all other meshes, I guess it is fine... Do we need so much KX_Font related only code? or could we re_use the existing functions in the rendering process to include KX_Font specific code?

source/gameengine/Ketsji/KX_FontObject.cpp
panzergame: The issue with text object is that they use multitexturing to use a font texture without material. I'm fully agree with you that a material would be the nicest case, but currently it's pretty complicated as we have to implicitly create a specific material for mesh which is able to use a font texture and that is not reused by other meshes. This is a similar case to how we manage GI materials and i have a lot of issues with it.

@youle31 : I will investigate deeper on the question, but i think that in any case we will need a fake material as KX_TextMaterial to make compatibility with shadeless text as before.

panzergame:
@youle31, @lordloki, @DCubix: I'm (and you too) are very uncertain with this patch, i propose you two choice: 1) merge it on 0.0.9 if everyone agree and that there's no ameliorations possible 2) wait 0.1.0 if someone has a amelioration possible.

Thanks for aswers.

youle31:
That's cool to fix issues with text transparency in 0.0.9. If we see other possible ameliorations later (I know nothing about fonts), we can do it later.

PR #157, opened 2016-07-20
Textured Area Lights

DCubix:
This re-applies the old patch for textured area lights. It was updated to work in the latest sources of UPBGE.
More info in the commit message :smiley: 
Screenshot


BluePrintRandom:
<3 <3 <3

youle31:
:+1:  It works better than the last time we applied it. I don't know why but it works better :)

adriansnetlis:
Does it work fine when you apply normalmap to ground and press play? The candy branch version resulted in balck floor.
Does this also not override a specific specular channel with square speculars but instead check for light type and always use the square speculars for area lights?

If so, than :+1:

DCubix:
I think this feature is ready to be merged. I made a lot of tests and it appears to work fine.

youle31:
Awesome! Thanks!

source/blender/gpu/GPU_material.h
```diff
@@ -178,6 +178,7 @@ typedef enum GPUDynamicType {
 	GPU_DYNAMIC_LAMP_COEFFCONST      = 13 | GPU_DYNAMIC_GROUP_LAMP,
 	GPU_DYNAMIC_LAMP_COEFFLIN        = 14 | GPU_DYNAMIC_GROUP_LAMP,
 	GPU_DYNAMIC_LAMP_COEFFQUAD       = 15 | GPU_DYNAMIC_GROUP_LAMP,
+	GPU_DYNAMIC_LAMP_DYNAREAMAT      = 16 | GPU_DYNAMIC_GROUP_LAMP,
```
panzergame: Can you move this line before under GPU_DYNAMIC_LAMP_DYNSPOTSCALE to order DYN and non DYN uniform (DYN means that the uniform can't be make constant).

source/blender/gpu/intern/gpu_material.c
```diff
@@ -152,8 +153,9 @@ struct GPULamp {
 	float obmat[4][4];
 	float imat[4][4];
 	float dynimat[4][4];
+	float dynareamat[4][4];
 
-	float spotsi, spotbl, k;
+	float spotsi, spotbl, k, area_size, area_sizey;
```
panzergame: I think you can cut the line.

source/blender/gpu/intern/gpu_material.c
```diff
@@ -877,18 +882,85 @@ static void shade_light_textures(GPUMaterial *mat, GPULamp *lamp, GPUNodeLink **
...
+					 GPU_builtin(GPU_VIEW_POSITION),
+					 GPU_image(mtex->tex->ima, &mtex->tex->iuser, false), GPU_uniform(mtex->size),
 					 GPU_select_uniform(&mtex->lodbias, GPU_DYNAMIC_TEX_LODBIAS, NULL, mat->ma),
-			         GPU_dynamic_uniform((float *)lamp->dynpersmat, GPU_DYNAMIC_LAMP_DYNPERSMAT, lamp->ob),
-			         &tex_rgb);
-			texture_rgb_blend(mat, tex_rgb, *rgb, GPU_uniform(&one), GPU_uniform(&mtex->colfac), mtex->blendtype, rgb);
+					 GPU_dynamic_uniform((float *)lamp->dynpersmat, GPU_DYNAMIC_LAMP_DYNPERSMAT, lamp->ob), &tex_rgb);
+
+			if (lamp->type == LA_AREA) {
+				one = 0.95f;
```
panzergame: don't assign `one` to everything else than 1. This value is like a constant in this function as `zero`. All code after can be affected, i know that is not the case here, but later if we add something after we will spend time to looking on a possible rare bug. In your case you can create a different value e.g `float factor = 0.95` and call `GPU_uniform(&factor)`.

Also can you explain why `texture_rgb_blend` must be different for area lamps ?

source/blender/gpu/intern/gpu_material.c
```diff
@@ -877,18 +882,85 @@ static void shade_light_textures(GPUMaterial *mat, GPULamp *lamp, GPUNodeLink **
...
+				texture_rgb_blend(mat, tex_rgb, *rgb, GPU_uniform(&one), GPU_uniform(&mtex->colfac), MTEX_MUL, rgb); 
+			}
+			else {
+				texture_rgb_blend(mat, tex_rgb, *rgb, GPU_uniform(&one), GPU_uniform(&mtex->colfac), mtex->blendtype, rgb);
+			}
+		}
+	}
+}
+
+static void shade_area_diff_texture(GPUMaterial *mat, GPULamp *lamp, GPUNodeLink **rgb, GPUShadeInput *shi)
```
panzergame: Why this function is unused ?

source/blender/gpu/intern/gpu_material.c
```diff
@@ -877,18 +882,85 @@ static void shade_light_textures(GPUMaterial *mat, GPULamp *lamp, GPUNodeLink **
...
+}
+
+static void shade_area_spec_texture(GPUMaterial *mat, GPULamp *lamp, GPUNodeLink **rgb, GPUShadeInput *shi)
+{
+	GPUNodeLink *tex_rgb;
+	MTex *mtex = NULL;
+	int i;
+	float one = 1.0f;
+       
+	for (i = 0; i<MAX_MTEX; ++i) {
```
panzergame: spaces around `<`

source/blender/gpu/intern/gpu_material.c
```diff
@@ -877,18 +882,85 @@ static void shade_light_textures(GPUMaterial *mat, GPULamp *lamp, GPUNodeLink **
...
+}
+
+static void shade_area_diff_texture(GPUMaterial *mat, GPULamp *lamp, GPUNodeLink **rgb, GPUShadeInput *shi)
+{
+	GPUNodeLink * tex_rgb;
+	MTex *mtex = NULL;
+	int i;
+	float one = 1.0f;
+       
+	for (i = 0; i<MAX_MTEX; ++i) {
```
panzergame: spaces around `<`

source/blender/gpu/intern/gpu_material.c
DCubix: I don't know, the patch is just like that. And I can't contact the original owners

panzergame:
It seems that the lamp scale is not taken in account.

youle31:
Hi, so to scale specular this is in shade area spec: for example:
vec2 halfSize = vec2(areaSizeX * 20.0, areaSizeY) / 2.0;
and for diffuse this is in shade inp area: example:
vec2 asize = vec2(areaSizeX * 20.0, areaSizeY);

so I think you just have to replace uniforms areaSizeX and areaSizeY in gpu_material.c with dynamic uniforms (or select uniform) and make the python api if needed.

test file: http://www.pasteall.org/blend/43010

youle31:
I'm not sure about the last commit (deletion of the UI)...

source/blender/gpu/intern/gpu_material.c
```diff
@@ -877,22 +883,81 @@ static void shade_light_textures(GPUMaterial *mat, GPULamp *lamp, GPUNodeLink **
 		MTex *mtex = lamp->la->mtex[i];
 
 		if (mtex && mtex->tex->type & TEX_IMAGE && mtex->tex->ima) {
-			mat->dynproperty |= DYN_LAMP_PERSMAT;
+			if (lamp->type == LA_AREA) {
+				mat->dynproperty = DYN_LAMP_VEC | DYN_LAMP_CO | DYN_LAMP_AREAMAT;
```
panzergame: |=

youle31:
Sometimes when we play with textures the specular becomes incorrect...: http://www.pasteall.org/blend/43132 I don't know how exactly we can reproduce the bug but it seems that there is a problem with the area matrix I guess (dynareamat iirc)

youle31:
I was in orthographic mode! Stupid me. So this is not a bug... all seems right... I make other tests

source/blender/gpu/intern/gpu_material.c
youle31: So we do this modif?

source/blender/gpu/intern/gpu_material.c
```diff
@@ -877,22 +883,81 @@ static void shade_light_textures(GPUMaterial *mat, GPULamp *lamp, GPUNodeLink **
...
+		if (mtex && mtex->tex->type & TEX_IMAGE && mtex->tex->ima) {
+			mat->dynproperty |= DYN_LAMP_CO | DYN_LAMP_AREAMAT;
+			GPU_link(mat, "area_diff_texture",
+					 GPU_builtin(GPU_VIEW_POSITION),
+					 GPU_image(mtex->tex->ima, &mtex->tex->iuser, false),
+					 GPU_select_uniform(&mtex->lodbias, GPU_DYNAMIC_TEX_LODBIAS, NULL, mat->ma),
+					 GPU_dynamic_uniform(lamp->dynco, GPU_DYNAMIC_LAMP_DYNCO, lamp->ob),
+					 shi->vn,
+					 GPU_dynamic_uniform((float *)lamp->dynareamat, GPU_DYNAMIC_LAMP_DYNAREAMAT, lamp->ob),
+					 GPU_dynamic_uniform((float *)lamp->area_size, GPU_DYNAMIC_LAMP_DYNAREASIZE, lamp->ob),
```
youle31: dynamic_uniform(lamp->areasize * obmat_scale...?

source/blender/gpu/intern/gpu_material.c
```diff
@@ -877,22 +883,81 @@ static void shade_light_textures(GPUMaterial *mat, GPULamp *lamp, GPUNodeLink **
...
+		if (mtex && mtex->tex->type & TEX_IMAGE && mtex->tex->ima) {
+			mat->dynproperty |= DYN_LAMP_CO | DYN_LAMP_AREAMAT;
+			GPU_link(mat, "area_spec_texture",
+					 GPU_builtin(GPU_VIEW_POSITION),
+					 GPU_image(mtex->tex->ima, &mtex->tex->iuser, false),
+					 GPU_select_uniform(&mtex->lodbias, GPU_DYNAMIC_TEX_LODBIAS, NULL, mat->ma),
+					 GPU_dynamic_uniform(lamp->dynco, GPU_DYNAMIC_LAMP_DYNCO, lamp->ob),
+					 shi->vn,
+					 GPU_dynamic_uniform((float *)lamp->dynareamat, GPU_DYNAMIC_LAMP_DYNAREAMAT, lamp->ob),
+					 GPU_dynamic_uniform((float *)lamp->area_size, GPU_DYNAMIC_LAMP_DYNAREASIZE, lamp->ob),
```
youle31: lamp->areasize * obmat_scale?

youle31: Sometimes when we play with render to texture, the image is not centered correctly. An option that could be cool to have is the offset buttons (beside texture size). I'll do some tests... Only as GPU_uniforms

youle31:
I propose this patch: http://pasteall.org/79162/diff to solve a problem related to superposition of diffuse + specular when we render specular only.

Test file: http://pasteall.org/blend/index.php?id=43533 (to test at runtime)

There is a bug related to rendering specular of area light texture when there is no other light in the scene. It was here before the patch I propose. I'm not really sure about the patch but this feature is really difficult to understand.

Side node: We always have to ensure that spotlight is rendered because I don't know why exactly there's a kind of conflict beetween area lights and spots.

youle31:
So, what the fuck we do with this :) ? @panzergame , @lordloki , @DCubix ? Please tell what's your mind about this. Can we commit something that we don't fully understand? Can we trust Martinsh? Can we commit something that is not perfect like we did with parallax?

panzergame:
@youle31 : I think we can't commit something not working for the most trivial case, or something that we don't understand it.

We have the main code, we need next to find better shader functions.

youle31:
Ok... It's a bit shame (dommage) because i think features like That (or parallax which is far from perfect and easy to use or cubemap That doesn't work on flat surfaces...) That allow to do shiny and impressive effects and make the user interested about upbge. And i don't think That There is a real bug in This feature... There are Just some cases where we are surprised by the shader behaviour... But it Is your décision and i respect (even if i'm frustrated haha)

adriansnetlis:
Hi... How comes that cubemaps don't work on flat surfaces?
I thought it worked as simple as
`reflectionVector = reflect(view, normal);
reflectionColor = textureCube(cubemapSampler, reflectionVector)`
If you use this method than it shall work flawlessly.
Um... I guess I should have found where you mention this at cubemap threads. :D

youle31:
I made a branch uptodate with last master: ge_textured_area_light2

BluePrintRandom:
Nice work youle!

panzergame:
Some shader compute are now shared, but there's still some issues:
- When the specularity is blurred, the area edges in the specularity are not blurred.
- Diffuse light projection is odd.

panzergame:
I think i done all possible to have a "correct" approximation of area light. @youle31, @DCubix, @lordloki, @Akira_San, @BluePrintRandom: Could you test this branch with the files in:
http://dropmefiles.com/VH9aZ.

panzergame:
The specular seems a bit odd:
capture datant de 2016-12-11-16h24min56s

Left a point specular and right an are light specular. The point specular is on the surface contrary to the area specular which seems under the surface.
@martinsh : Do you remember this part of the shader ? And know more or less the issue/solution ?

lordloki:
@panzergame windows defender says that there is a trojan in the files.rar. Have you checked it?

panzergame:
@lordloki : I rechecked, the file is right.

youle31:
test:
- Bug with mipmap while we play video.
- The specular intensity decrease when angle beetween lightvec and viewvec increase (This is the same behaviour as for spotlights, but I don't know if this is a good behaviour for area lights)
- Problem with texts and render to texture in the file I shared above (But I don't think this is related to this patch)
- The far the area light is, the more its texture seems blurry (I don't know if this is good?)

The mix with other lights sounds good to me. I saw what you meant when you said: "specular which seems under the surface" but I don't know if this is a bug.

panzergame:
Thanks for the tests youle. For mipmap issue we can step aside this bug as it is not only related to area lights. I will continue checking the specularity, and for the texture it's a side effect of the use of lod bias… i don't know if there's an alternative to blur a texture inside the shader.

DCubix:
@panzergame The specular reflection is not wrong. That's exactly how area lights work.
You can make a test by putting a matrix of point lights in the scene:

Since area lights is basically a matrix of point lights, you can see the specular is accurate!

Akira1San:
Yey, Area lights for 0.1.7. :D

DCubix:
This is currently broken 😢 
@martinsh @panzergame help me!

DCubix:
Ok so, apparently Eevee has a way better implementation for Area Lights. It works even in orthographic view!
So when we finally merge with 2.8, this will be there.
I will (finally) close this!

UnidayStudio:
Let's re-open this! :)

PR #164, opened 2016-07-24
Implement geometry shader

panzergame:
This branch implement geometry shader for custom shaders. As we use GPUShader we just have to pass the string to the GPU_shader_create function.

The major issue of this feature is that the geometry shader can't be passed to the python function `setSource` as it will look like : `setSource(vert, frag, apply, geom)` and latter if we add tesselation shader it will became more complicated. To solve this issue a second function was added : `setSourceDict(sources, apply)`. The `sources` argument is a dictionnary containing all shader script used it will look at a full usage like :
```
sources = {"vertex" : vert,
                  "fragment" : frag,
                  "geometry" : geom
                 }
```

The only think changed in `setSource` is that we set the geometry shader to an empty string.

Questions:
1) What name for `setSourceDict` ? `setSources`, `setSourceList` ?

### DIFF
--- a/doc/python_api/rst/bge_types/bge.types.BL_Shader.rst
+++ b/doc/python_api/rst/bge_types/bge.types.BL_Shader.rst
@@ -67,7 +67,7 @@ base class --- :class:`PyObjectPlus`
       :arg index: Texture sample index.
       :type index: integer
 
-   .. method:: setSource(vertexProgram, fragmentProgram)
+   .. method:: setSource(vertexProgram, fragmentProgram, apply)
 
       Set the vertex and fragment programs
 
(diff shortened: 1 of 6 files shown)


panzergame:
Example file : http://www.pasteall.org/blend/42843

youle31:
Very cool :) It works for me with this simplificatrion: http://www.pasteall.org/72405/diff

lordloki:
I think that setSourceList would be better. Additionally, If we are defining a new (and future-proof) setSourceList function I would deprecate the old (and non guture-proof) setSource function.

adriansnetlis:
Wow! Great to see those amazing news - so now we can make tessellation in custom shaders without pyOpenGL or such library help? That sounds brilliant!:)

panzergame:
@lordloki, @youle31, @DCubix : Can you review a bit the patch before i merge it in master ?

lordloki:
It seems ok to me. Good work!!

PR #165, opened 2016-07-25
UPBGE: Add lodbias option for world textures

youle31:
Waiting for the debate about constant uniforms in the viewport, I added
GPU_uniform(&mtex->lodbias) in do_world_tex and the corresponding UI

test file (from alexander romanov): http://www.pasteall.org/blend/42853


panzergame:
The `GPU_uniform` is ok cause the world texture parameter are not animable e.g : blend factor, zenith up factor, zenith down factor.

@DCubix, @lordloki : Is all fine for you ?

lordloki:
Yes, it seems fine

PR #169, opened 2016-07-29
UPBGE: Add a vibration actuator for joystick

youle31:
This is a simple version of vibration actuator where you can set:

- joystick index (indicates which joystick will rumble)
- strength (the force of the vibration)
- duration (the duration of the vibration)

For the moment we can't set vibration "type" but it is ready to use.

The python API is included (not tested)

For some types of joystick devices (mine for example (xbox controller)),
SDL 2.0.4 is needed. The current version is SDL 2.0.3.

SDL 2.0.4 can be found here: https://www.libsdl.org/download-2.0.php

Test file: http://www.pasteall.org/blend/42908

Some improvements can be made for:

- UI
- Vibration "types" handling
- Some changes in KX_VibrationActuator::Update function to improve actuator sensor functionality (return true until the vibration is finished, otherwise return false). We can have a look on sound actuator to see how this is done. (int SDL_HapticGetEffectStatus(SDL_Haptic* haptic, int effect)


lordloki:
Good work @youle31 :+1:  ,

some comments (I know it is preliminary work):
- Move KX_VibrationActuator to SCA_VibrationActuator in GameLogic
- Also, we should develop a new DEV_JoystickVibration where we handle the play modes, etc. Basically, to avoid to call SDL functions directly from SCA_VibrationActuator file.

youle31:
Thanks for review and comments :) . I'll continue the stuff a bit later. Yeah it's a preliminary work, I did it this night

source/gameengine/GameLogic/SCA_VibrationActuator.cpp
```diff
@@ -0,0 +1,143 @@
...
+	if (!instance) {
+		return false;
+	}
+	bool bNegativeEvent = IsNegativeEvent();
+
+	if (bNegativeEvent) {
+		return false;
+	}
+	
+	SDL_Haptic *haptic;
```
panzergame: To rejoin lordloki I think you can create a `DEV_Joystick::RumblePlay(strength, duration)` to avoid include SDL here.

source/gameengine/GameLogic/SCA_VibrationActuator.h
```diff
@@ -0,0 +1,59 @@
+/*
```
panzergame: The GPL license block should be without the "The Original Code is Copyright (C) 2001-2002 by NaN Holding BV.". To avoid this frequent error I made a wiki page : https://github.com/UPBGE/blender/wiki/GPL-license-block.

panzergame:
The patch looks pretty good.

youle31:
Thanks panzergame for review and comments :) I think I did what you suggested. I also fixed the value returned by update function in vibrationactuator.cpp.

I think it's better to keep a simple UI with only strength and duration, but we could make a python API to handle more complex vibration effects... What do you think @panzergame and @lordloki ?

youle31:
I need average framerate for this actuator. Can I do like this: http://www.pasteall.org/73557/diff ? (#include "../Ketsji/KX_KetsjiEngine.h" then engine->GetAverageFrameRate()) or is it better to write custom function to compute framerate with #include ctime ?

panzergame:
The actuator should work with time or frames ?

source/gameengine/GameLogic/SCA_VibrationActuator.cpp
```diff
@@ -0,0 +1,124 @@
...
+	SCA_VibrationActuator* replica = new SCA_VibrationActuator(*this);
+	replica->ProcessReplica();
+	return replica;
+}
+
+bool SCA_VibrationActuator::Update()
+{
+	/* Can't init instance in constructor because m_joystick list is not available yet */
+	SCA_JoystickManager *mgr = (SCA_JoystickManager *)GetLogicManager();
+	DEV_Joystick *instance = mgr->GetJoystickDevice(m_joyindex) ? mgr->GetJoystickDevice(m_joyindex) : NULL;
```
panzergame: ehh, this doesn't work ?:
```
DEV_Joystick *instance = mgr->GetJoystickDevice(m_joyindex);
```

youle31: surely I fix it hehe

source/gameengine/Device/DEV_Joystick.cpp
```diff
@@ -40,6 +40,8 @@
 #include "DEV_JoystickPrivate.h"
 #include "DEV_JoystickMappingdb.h"
 
+#include <iostream> //std::cout
```
panzergame: I think you can remove this include.

PR #173, opened 2016-08-12
UPBGE: Apply modifiers at game start and restore it at game exit

youle31:
This saves the mesh as it is before applying modifiers in
KX_BlenderSceneConverter constructor. At ge exit we restore the saved mesh
and delete the mesh created when we apply modifiers.

This is done with a bpy script. I don't know anything about bpy so the
script needs to be improved.

Before using this method I used apply_modifer method and then I used undo
to restore the mesh but it was unstable. This seems to be more stable.

To do: save and restore selected objects when we start the game engine
make a list of modifiers that cause problems in ge runtime and apply only
modifiers that cause problems.
The UI can be changed and designed to be per mesh and not per scene.
Crash in docks level at exit.
optionnal: replace python code with c++ code (for pros)

when we apply modifiers on very big meshes it can freeze the game engine.

Performance tests can also be done.

test file: - https://developer.blender.org/T48458

Oh, and I forgot, the option to enable the feature is in render panel beside show mouse cursor

I have very nice performance results with some files!! :)

http://www.pasteall.org/blend/43097
http://www.pasteall.org/pic/show.php?id=105900
http://www.pasteall.org/pic/show.php?id=105901

### DIFF
--- a/release/scripts/startup/bl_ui/properties_data_mesh.py
+++ b/release/scripts/startup/bl_ui/properties_data_mesh.py
@@ -383,6 +383,7 @@ class DATA_PT_customdata(MeshButtonsPanel, Panel):
         col.prop(me, "use_customdata_vertex_bevel")
         col.prop(me, "use_customdata_edge_bevel")
         col.prop(me, "use_customdata_edge_crease")
+        col.prop(me, "apply_modifiers_at_ge_start")
 
 
 class DATA_PT_custom_props_mesh(MeshButtonsPanel, PropertyPanel, Panel):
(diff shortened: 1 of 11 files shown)


youle31:
I need help on this... I don't know how to avoid crash in any case: for example if you try to apply the modifier on a mesh that has 1000000 polygons.

I have no idea to improve it anymore... We can choose to leave it like this and consider it as an experimental feature and display warnings or find another idea to try to improve it.

lordloki:
I dont find the reason that crash blender because the method that you use is the correct (it is the used in the export addons).

youle31:
Yes, in fact when I apply modifiers on a cube with a subsurf modifier with 6 subdivisions, it takes ~3GBytes of RAM (the mesh is saved in the RAM) so if I duplicate this cube 3 times, it will take ~9GBytes of RAM and I have only 8GBytes so there is a crash... So maybe I could use import_ply.py/export_ply.py to save the mesh to a text file and maybe the RAM wouldn't be fill like this... I'll give it a try.

lordloki:
I think that it is preferred to show a warning about hardware limits that to export the mesh to an external format. I think it will be very slow when we have 100s of meshes

youle31:
Finally panzergame found a better solution (ge_nomodifier) :) So I close this for the moment.

PR #174, opened 2016-08-12
UPBGE: Add an API to check if a gameobject is inside frustum

youle31:
or more exactly if an object is culled or not. Why this wasn't available
before?? Unless it was available and I didn't see it?

This can be very handy in python scripts for example to not update shader
uniforms when the gameobject with shader is culled. Can save a lot of
performances in many cases I guess.

test file: http://www.pasteall.org/blend/43105

### DIFF
--- a/doc/python_api/rst/bge_types/bge.types.KX_GameObject.rst
+++ b/doc/python_api/rst/bge_types/bge.types.KX_GameObject.rst
@@ -274,7 +274,11 @@ base class --- :class:`SCA_IObject`
 
    .. attribute:: culled
 
-      Returns True if the object is culled, False if not.
+      Returns True if the object is culled, else False.
+
+      .. warning::
+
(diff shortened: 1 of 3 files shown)


panzergame:
` Why this wasn't available before??` Because object culling is a valid data after processing cullings, but all logic is proceed before the culling stage. The only part where it's useful is for `scene.post_draw` or `scene.pre_draw` (not `scene.pre_draw_setup`). Other than this limitation I'm agree to expose the culling value.

panzergame:
About name, what about `culled` simply ? Also in the python documentation you should expose the limitations i described before.

youle31:
panzergame: the value returned seems correct no? http://www.pasteall.org/blend/43107

panzergame:
The value can be correct if the camera doesn't move and if there's no multiple viewports or stereo.

youle31:
maybe we could add such a sample code:

def isCulled():

    own["culled"] = own.culled

scene.pre_draw = [isCulled]

PR #175, opened 2016-08-13
Per material storage settings.

panzergame:
Previously in UPBGE and BFBGE the storage used to render was limited to all the material/mesh in the scene. This is a problem because some mesh are rendered faster in special storage. e.g geometry instancing meshes with VBO, big meshes with VA+DL and animated meshes with VBO or VA only.

To solve all of these issues a per material storage type is exposed. This storage expose four modes:
- auto
- va
- vbo
- scene

The only mode added compared to the scene storage is the mode `scene`. This mode only reuse the storage setting from the scene, it's the mode used for compatibility and default. 

From the UI side:
[img]http://i.imgur.com/HZ6V3o9.png[/img]

From the code side:
- The functions [Index/Bind/Unbind]Primitives now take as first argument the storage type. The storage type is taken from the enumeration RAS_IRasterizer::StorageType.
- RAS_IPolyMaterial contains now two new getter for the storage type and the display lists support.
- KX_BlenderMaterial made the conversion for the storage type and the display lists support.

@youle31, @DCubix, @lordloki : Can you please review this patch ?

### DIFF
--- a/source/blender/makesrna/intern/rna_material.c
+++ b/source/blender/makesrna/intern/rna_material.c
@@ -890,7 +890,6 @@ static void rna_def_material_gamesettings(BlenderRNA *brna)
 	};
 
 	static EnumPropertyItem storage_items[] = {
-		{RAS_STORE_AUTO, "AUTO", 0, "Auto Select", "Choose the best supported mode"},
 		{RAS_STORE_VA, "VERTEX_ARRAY", 0, "Vertex Arrays", "Usually the best choice (good performance with display lists)"},
 		{RAS_STORE_VBO, "VERTEX_BUFFER_OBJECT", 0, "Vertex Buffer Objects",
 		                "Typically slower than vertex arrays with display lists, requires at least OpenGL 1.4"},
--- a/source/blender/makesrna/intern/rna_material.c
+++ b/source/blender/makesrna/intern/rna_material.c
@@ -890,10 +890,10 @@ static void rna_def_material_gamesettings(BlenderRNA *brna)
 	};
 
 	static EnumPropertyItem storage_items[] = {
-		{RAS_STORE_VA, "VERTEX_ARRAY", 0, "Vertex Arrays", "Usually the best choice (good performance with display lists)"},
-		{RAS_STORE_VBO, "VERTEX_BUFFER_OBJECT", 0, "Vertex Buffer Objects",
+		{GAME_STORAGE_VA, "VERTEX_ARRAY", 0, "Vertex Arrays", "Usually the best choice (good performance with display lists)"},
+		{GAME_STORAGE_VBO, "VERTEX_BUFFER_OBJECT", 0, "Vertex Buffer Objects",
 		                "Typically slower than vertex arrays with display lists, requires at least OpenGL 1.4"},
-		{RAS_STORE_SCENE, "SCENE", 0, "Scene Default", "Use scene storage mode"},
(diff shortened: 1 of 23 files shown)


source/blender/makesrna/intern/rna_material.c
```diff
@@ -888,6 +888,15 @@ static void rna_def_material_gamesettings(BlenderRNA *brna)
...
 	};
+
+	static EnumPropertyItem storage_items[] = {
+		{RAS_STORE_AUTO, "AUTO", 0, "Auto Select", "Choose the best supported mode"},
+		{RAS_STORE_VA, "VERTEX_ARRAY", 0, "Vertex Arrays", "Usually the best choice (good performance with display lists)"},
+		{RAS_STORE_VBO, "VERTEX_BUFFER_OBJECT", 0, "Vertex Buffer Objects",
+		                "Typically slower than vertex arrays with display lists, requires at least OpenGL 1.4"},
+		{RAS_STORE_SCENE, "SCENE", 0, "Scene Default", "Use scene storage mode"},
+		{0, NULL, 0, NULL, NULL}
+	};
```
lordloki: Could it be possible to add this info ("This is a problem because some mesh are rendered faster in special storage. e.g geometry instancing meshes with VBO, big meshes with VA+DL and animated meshes with VBO or VA only.") in each text explanation?

youle31:
I agree with Lordloki. #define RAS_STORE_IMMEDIATE	1 could be removed. I'd remove also display list (enable it aways for VA) and why not remove it also in Render Tab? (when I don't use display lists my framerate is /100...). Finally Scene Default is a good name I think (after test). Sounds a cool patch!

source/gameengine/Ketsji/KX_BlenderMaterial.cpp
```diff
@@ -84,6 +85,32 @@ KX_BlenderMaterial::KX_BlenderMaterial(
...
+		storage = blenderScene->gm.raster_storage;
+		m_flag |= (blenderScene->gm.flag & GAME_DISPLAY_LISTS) ? RAS_DISPLAYLISTS : 0;
+	}
+	else {
+		m_flag |= (game->storage_flag & GEMAT_DISPLAY_LISTS) ? RAS_DISPLAYLISTS : 0;
+	}
+
+	switch (storage) {
+		// Currently the AUTO mode is equivalent to VA storage.
+		case RAS_STORE_AUTO:
```
lordloki: The auto mode, can it be erased? (As it is the same as VA)

panzergame:
DL settings is let accessible to the user because the user can modify a vertex through the python API and we can't determine in the code if it will be frequently.

DL can't be compared to VAO in this point because VAO are created for a VBO and the contents of this VBO is independent. For DL the content is copied in the DL so we need to recreate is for each modifications.

youle31:
OK... I never tested if when we set vertex attribute it is faster without DL than with DL but I doubt.... But let the option don't cause major issue so I stop debating :)

panzergame:
If the user modify a mesh with VA + DL it will cost the DL recompilation but will win on the render.
Without DL the user will cost the slower render of a VA.

But DL cost like 10/30 more than the performance gap between DL+VA and VA only. The user should take this in account only if it modify the mesh frequently, else he can spend the recompilation time.

youle31:
Ok Thanks for the explainations :)

PR #178, opened 2016-08-17
Use multiple vertex formats.

panzergame:
Previously the vertex class (`RAS_TexVert`) was using 8 channel for UV (64bytes). But the user use 8 channel very rarely.

The issue is the memory used for a vertex which is to0 high and will be more for multiple color channels.

To reduce the memory size we have to make a vertex class with a UV channel number as template argument to size the uv array to the uv channels existing.
`RAS_TexVert<unsigned int UVSize>`

This class is subclass of the base class `RAS_ITexVert` for multiple reasons:
- Keep using vertex class in deformers and avoid using indirect virtual RAS_DisplayArray function.
- Allow base functions for getting/setting position/normal/tangent to be inlined.

In this inheritance as all base functions are inlined only the following functions are virtual:
- virtual const unsigned short getUVSize() const
- virtual const float *getUV(const int unit) const
- virtual void SetUV(const int index, const MT_Vector2& uv)
- virtual void SetUV(const int index, const float uv[2])

The only disadvantage of an inheritance is that the vptr take 8 bits more in the vertex class.
To counter this side effect, some vertex data are split in a new class `RAS_TexVertInfo`. The data contained are:
- int m_origindex
- short m_softBodyIndex
- short m_flag

The reason to doing this is that these data are not needed for the render and are very rarely accessed (only for deformers).

About `RAS_DisplayArray` a similar change to `RAS_TexVert` was applied. `RAS_DisplayArray` take as template parameter the vertex class. It contains the `std::vector` of vertexes and indices and it is based upon `RAS_IDisplayArray`.

This class is now obligated to give access to vertexes/indices through virtual functions:
- virtual RAS_ITexVert *GetVertex(unsigned int index) const
- virtual unsigned int GetIndex(unsigned int index) const

As the vertex type is **dynamic** we have to pass the template paramter to the functions/classes which instantiate the `RAS_DisplayArray`. A new class `RAS_TexVertFormat`, which is currently only containing the uv size, is passed to the functions:
- RAS_MeshObject::AddMaterial
- RAS_MaterialBucket::AddMesh
- RAS_MeshSlot::init

The uv size value is choose in BL_ConvertMesh with the expression : 
max_ii(1, validLayers)

About the rasterizer storage the modification are pretty minors. Only VBO need to get offset in an other way because computing the offset with a virtual function (getUV) with a NULL instance raise a segfault. So RAS_DisplayArray gives offset with functions:
- virtual unsigned int GetVertexMemorySize() const
- virtual void *GetVertexXYZOffset() const
- virtual void *GetVertexNormalOffset() const
- virtual void *GetVertexTangentOffset() const
- virtual void *GetVertexUVOffset() const
- virtual void *GetVertexColorOffset() const

Performance test are made with three cases: (1) without ge_vertex, (2) with ge_vertex without `RAS_TexVertInfo`, (3) with ge_vertex and `RAS_TexVertInfo`.

Armature 1 UV:

| Category | 1 | 2 | 3 |
| -------------| -----------| -----------| -----------|
| Animation | 5.62ms | 4.19ms | 4.09ms |
| Rasterizer | 3.72ms | 2.00ms | 1.64ms |

Armature 8 UV:

| Category | 1 | 2 | 3 |
| -------------| -----------| -----------| -----------|
| Animation | 5.61ms | 6.67ms | 4.52ms |
| Rasterizer | 3.62ms | 3.54ms | 3.38ms |

Soft body 1 UV:

| Category | 1 | 2 | 3 |
| -------------| -----------| -----------| -----------|
| Rasterizer | 0.22ms | 0.20ms | 0.20ms |

Soft body 8 UV:

| Category | 1 | 2 | 3 |
| -------------| -----------| -----------| -----------|
| Rasterizer | 0.56ms | 0.58ms | 0.55ms |

About memory the size are:

| UVs |  1  |  2  |  3  |  4 |  5  |  6    |   7   |  8   |  
| -------------| --- | --- | ----| --- | --- | ----- | ----- | ---- |
| Size (bits) | 64 | 72 | 80 | 88 | 96 | 104 | 112 | 120 |
| Compiler pad | 4 | 4 | 4 | 4 | 4 | 4 | 4 | 4 |
| Cache lines | 1 | 2 | 2 | 2 | 2 | 2 | 2 | 2 |

@youle31, @lordloki, @DCubix : I will enjoy you review this patch these days.


DCubix:
I made the review, but I didn't really understand why you did it. Seems good to me but i'm a little confused

panzergame:
A cache vertex pointer system was introduced. Indeed calling RAS_DisplayArray::GetVertex from python mesh.getVertex was costing more than before. To fix this issue a list named m_vertexPtrs is added, this list contain the vertex pointer at the end of the conversion. The advantage of this list is to replace a virtual call by a inline function.

By doing this the speed down is avoided.

PR #183, opened 2016-08-19
UPBGE: Fix crash when calling setUniform when setSource has not been

youle31:
called

Previously, when you called setUniform... methods, m_error variable was
used to return NULL/None when setUniform cannot be called. But m_error
variable is set only when you previously called setSource. The default
value of m_error is false (in constructor) because if you set it to true
in constructor, getShader doesn't work anymore.

So m_error was false when you haven't called setSource previously and this
caused a crash when you tried to call setUniform...

So instead of using m_error (that is useful for getShader method), I
choosed to use if (!m_ok) return NULL instead to fix the crash when
shder.setSource is not called before because m_ok has the right value in
constructor and is correctly set in setSource...

test file: http://www.pasteall.org/blend/43213

### DIFF
--- a/source/gameengine/Ketsji/BL_Shader.cpp
+++ b/source/gameengine/Ketsji/BL_Shader.cpp
@@ -222,7 +222,7 @@ KX_PYMETHODDEF_DOC(BL_Shader, getFragmentProg, "getFragmentProg( )")
 
 KX_PYMETHODDEF_DOC(BL_Shader, validate, "validate()")
 {
-	if (m_error) {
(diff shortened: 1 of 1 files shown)


panzergame:
There's also function `RAS_Shader::Ok()` but it wil assume that if the user disable a shader calling in python `RAS_Shader.enabled = True`, then it will be impossible to set an uniform.

youle31:
So my commit is correct, no? http://www.pasteall.org/blend/43224

PR #184, opened 2016-08-19
Soft particles / Depth transparency.

panzergame:
Soft particles or depth transparency is a technique used to blend transparent meshes (planes in general) with the other meshes depending of the depth between the two.

For the user it's a very simple interface: An option in the UI to enable depth transparency and a factor parameter to know how much blending the meshes with the others.

To implement this feature we first need to access to the depth texture, it's why this branch is upon ge_fbo. After we have to pass the depth texture to the shader. It wasn't possible because the depth texture will change for each windows resizing and per game start. To do it we need to use a pointer to this depth texture (GPUTexture **). The function GPU_dynamic_texture_ptr does this job.
At the end the shader need to read the depth and compare it by interpolating the current vertex position and the vertex position scale, with the depth value as factor.


source/gameengine/Rasterizer/RAS_IRasterizer.h
```diff
@@ -663,6 +664,8 @@ class RAS_IRasterizer
 
 	virtual void RemoveLight(RAS_ILightObject *lightobject) = 0;
 
+	virtual void UpdateGlobalDepthTexture() = 0;
```
panzergame: I'm not sure about this function name. Do you guys have a proposition ?

youle31:
I like this name, It's clear enough for me. Just looking at the name (not at the code), I'll say that this function updates the depth texture in the fbo rendered with the active_camera. Tell me if I'm wrong, if not, I think this means that's a good name. And it would be very cool if we can access this depth texture for shaders...

panzergame:
Can this branch be reviewed for Thursday ?

panzergame:
Test file : http://pasteall.org/blend/index.php?id=43272

PR #185, opened 2016-08-19
Fix full framerate timing issue.

panzergame:
Previously the full framerate mode incremented the clock time by the frame time step (= scale / fps) to be sure to proceed a frame with the algorithm:
```
frames = int(deltatime * m_ticrate / m_timescale + 1e-6);
```

But by doing this the clock time and the frame time was totally wrong and not constant.

To fix this issue in case of non-fixed framerate an other way was used. First the clock time is incremented only by the time from the previous frame as for fixed FPS. Second the number of frame to proceed is always set to 1. Third the frame time step is set to the increment value of the clock time.

This patch also remove the ambiguity about "fixed time" and "fixed framerate". The both therms are opposite in the code.

It was tested with physics simulation and time scaling.

@youle31, @lordloki, @DCubix : Can you test this patch with every blend and play around options vsync and "use framerate" ? Thanks in advance.

### DIFF
--- a/source/gameengine/Ketsji/KX_KetsjiEngine.cpp
+++ b/source/gameengine/Ketsji/KX_KetsjiEngine.cpp
@@ -121,7 +121,7 @@ KX_KetsjiEngine::KX_KetsjiEngine(KX_ISystem *system)
 	m_inputDevice(NULL),
 	m_bInitialized(false),
 	m_activecam(0),
-	m_bFixedTime(false),
+	m_fixedFramerate(false),
 	m_useExternalClock(false),
 	m_firstframe(true),
 	m_frameTime(0.0f),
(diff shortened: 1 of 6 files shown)


youle31:
I don't know how it should work, but for example, when you uncheck use framerate and disable vsync, timer property goes very fast, so the shader based on timer goes very fast too.

In water demo 0.99, the mouse look use linear velocity, when you uncheck use frame rate disable vsync, the mouselook is very fast whereas the objects displayed in the scene go to a normal speed (except the objects with shader based on timer property)

panzergame:
Thanks for the report.

The bug about shader time is fixed, but the mouse look is time independent and then impossible to slow down to the the frame time step.

pqftgs:
Sweet.  This has been a huge problem.  It's great to have physics and animations playing at the expected speed.

I tested 2 games that use framerate-independent movement scripts, with both vsync and "use frame rate" disabled.

One issue:  Any dynamic/rigid-body object with collision bounds starting on a surface will go flying.  This can be replicated with a cube & plane.
http://pasteall.org/blend/index.php?id=43499

panzergame:
Apparently it's related only to the margin collision solving. The issue is not here without margin, i will look if i can solve it...

panzergame:
@pqftgs : I didn't find a solution for the margin. The bug appears only with static triangle mesh (bvhtrianglemesh) and margin.

PR #188, opened 2016-08-21
Ge suspendphysics

youle31:
test file: http://www.pasteall.org/blend/43239

### DIFF
--- a/source/gameengine/Ketsji/KX_GameObject.cpp
+++ b/source/gameengine/Ketsji/KX_GameObject.cpp
@@ -70,9 +70,6 @@
 #include "BKE_object.h"
 #include "BKE_python_component.h"
 
-#include "CcdPhysicsController.h"
-#include "CcdPhysicsEnvironment.h"
-
 #include "BL_ActionManager.h"
 #include "BL_Action.h"
 
@@ -239,10 +236,7 @@ void KX_GameObject::SetName(const char *name)
 
 PHY_IPhysicsController* KX_GameObject::GetPhysicsController()
 {
-	if (m_pPhysicsController && !m_suspendedPhysics) {
-		return m_pPhysicsController;
-	}
-	return NULL;
+	return m_pPhysicsController;
 }
 
 KX_GameObject* KX_GameObject::GetDupliGroupObject()
@@ -341,7 +335,7 @@ void KX_GameObject::SetParent(KX_GameObject* obj, bool addToCompound, bool ghost
 		RemoveParent();
 		obj->GetSGNode()->AddChild(GetSGNode());
 
-		if (GetPhysicsController())
+		if (m_pPhysicsController)
 		{
 			m_pPhysicsController->SuspendDynamics(ghost);
 		}
@@ -364,7 +358,7 @@ void KX_GameObject::SetParent(KX_GameObject* obj, bool addToCompound, bool ghost
 			Release();
 		// if the new parent is a compound object, add this object shape to the compound shape.
 		// step 0: verify this object has physical controller
-		if (GetPhysicsController() && addToCompound)
+		if (m_pPhysicsController && addToCompound)
 		{
 			// step 1: find the top parent (not necessarily obj)
 			KX_GameObject* rootobj = (KX_GameObject*)obj->GetSGNode()->GetRootSGParent()->GetSGClientObject();
@@ -402,7 +396,7 @@ void KX_GameObject::RemoveParent()
 		if (!rootlist->SearchValue(this))
 			// object was not in root list, add it now and increment ref count
 			rootlist->Add(AddRef());
-		if (GetPhysicsController())
+		if (m_pPhysicsController)
 		{
 			// in case this controller was added as a child shape to the parent
 			if (rootobj != NULL && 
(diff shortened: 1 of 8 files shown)


youle31:
This file: http://pasteall.org/blend/index.php?id=43227 shows that there is no culling for physics ?!

This is an Enormous performances loose. I think we should find a way to implement physics culling just by deleting physics controller on some conditions. The process could be partly controlled by the user. I add this to the framapad.

youle31:
@panzergame : Do you think we have a risk with this patch? With children objects or other things?

BluePrintRandom:
Is this patch ready?

I would love to see how much less CPU it uses then my current scheme :D 
https://www.youtube.com/watch?v=hOsDv3P1BA4

panzergame:
@youle31 : I would like that you test to modify physics attributes when the object is disabled.
Like reinstancing a physics shape, change the mass, apply a velocity ect...

youle31:
@BluePrintRandom : could you test this build: https://drive.google.com/file/d/0B3GouQIyoCmrZXBfTU01djNxM3M/view?usp=sharing

to see if this works for you and report us if you have bugs, performances gain etc...? Thanks!

BluePrintRandom:
sure!

BluePrintRandom:
object.suspendPhysics() and object.restorePhysics() right?

what is the api?

BluePrintRandom:
added.suspendPhysics = True

got it I think

BluePrintRandom:
Ok! tested - it's a little faster then adding/removing objects each frame,
and also it removes the complexity of needing a physics mesh and a seperate
graphics mesh for my kdtree example and also decreases the bloat to
scene.objects

so
benefits -
*less objects* = faster item lookup
*suspend is faster than *- add/remove
*reduced complexity - *Now it's easier to 'ghost' things as well


*http://pasteall.org/blend/index.php?id=43423
<http://pasteall.org/blend/index.php?id=43423>*

source/gameengine/Ketsji/KX_GameObject.cpp
```diff
@@ -236,7 +240,10 @@ void KX_GameObject::SetName(const char *name)
 
 PHY_IPhysicsController* KX_GameObject::GetPhysicsController()
 {
-	return m_pPhysicsController;
+	if (m_pPhysicsController && !m_suspendedPhysics) {
```
panzergame: These modifications are needed ?

source/gameengine/Ketsji/KX_GameObject.cpp
```diff
@@ -120,7 +123,8 @@ KX_GameObject::KX_GameObject(
       m_components(NULL),
       m_pInstanceObjects(NULL),
       m_pDupliGroupObject(NULL),
-      m_actionManager(NULL)
+      m_actionManager(NULL),
+	  m_pSavedPhysicsController(NULL)
```
panzergame: What's the goal of `m_pSavedPhysicsController` ? Please avoid the type in name convention : replace `m_p` by `m_`.

panzergame:
Using the physics controller while it is suspended should not be a problem. The physics controller is still valid (it's rigid/soft body too) but just not used in the physics simulation world.

You just have to check that all the cases calling CcdPhysicsEnvironment::AddCcdPhysicsController are impossible when the physics controller is considerated suspended. I'm thinking about calling from python restoreDynamics.

source/gameengine/Ketsji/KX_GameObject.cpp
youle31: When we try to set an attribute if physics controller has been removed with : 
CcdPhysicsEnvironment *env = (CcdPhysicsEnvironment *)GetScene()->GetPhysicsEnvironment();
env->RemoveCcdPhysicsController((CcdPhysicsController *)GetPhysicsController());

it causes a crash.

source/gameengine/Ketsji/KX_GameObject.cpp
youle31: m_pSavedPhysicsController is not needed indeed

source/gameengine/Ketsji/KX_GameObject.cpp
panzergame: It happens for every physics attributes ?

youle31:
When you call RemoveCcdPhysicsController you can't use anymore functions to set attributes like mass...

e.g: http://pasteall.org/blend/index.php?id=43454

source/gameengine/Ketsji/KX_GameObject.cpp
youle31: It happens at least for mass and as almost all attributes are set like mass, I guess this is the same

BluePrintRandom:
Will this branch be merged soon?

(I like it :D)

panzergame:
@youle31 : I propose you the patch : http://pasteall.org/82228/diff

I use `restorePhysics`and `disablePhysics` to be consistent with the functions `restoreDynamics` and `disableDynamics`.
The issue of changing mass with a suspended physics controller is resolved by checking if the physics controller is contained in the physics environment.
The same fix is applied for any functions using `UpdateCcdPhysicsController`in`CcdPhysicsController`.

youle31:
ok cool :)

panzergame:
@youle31 : Do you agree to apply my patch in the branch ?

@DCubix, @lordloki : Are you agree with the current works ? Can you test with the example and export it to bigger files for tests ?

youle31:
Yes i agree. Or You Can do ot if You want

youle31:
@panzergame : I can't apply your patch in ge_suspendphysics neither in master...

panzergame:
Merged in master.

PR #189, opened 2016-08-22
UPBGE: Fix not culled texts in sintel files

youle31:


source/gameengine/Rasterizer/RAS_OpenGLRasterizer/RAS_OpenGLRasterizer.cpp
```diff
@@ -2134,6 +2134,9 @@ void RAS_OpenGLRasterizer::RenderText3D(
 	/* gl prepping */
 	DisableForText();
 
+	Enable(RAS_CULL_FACE);
+	glCullFace(GL_BACK);
```
panzergame: This call to different things than SetFrontFace(true) ?

panzergame:
I think the call to Enable(RAS_CULL_FACE) can be moved in DisableForText if no issues are encountered.
Thanks for the fix.

youle31:
ok, done :)

youle31:
I think it's ok to merge. The only think is that backface culling is always enabled. Maybe later we could link the material backface culling option in material tab to have the choice to enable/disable it, but for now I think it's ok...

PR #190, opened 2016-08-24
Bug fixes for off screen render.

panzergame:
This branch introduced fixes around the previous version of gd_fbo.

It fixes the culling quad draw for the water filter, stereo gray clear and glitch on blenderplayer resize.

@DCubix, @youle31 : Can you test this branch with the files bugged ? thanks.


youle31:
Could you share the bugged files?

PR #191, opened 2016-08-24
Geometry instancing inverse function for GLSL <=1.30

panzergame:
This patch implement  the manual function to inverse mat4 in GLSl for version under 1.40. 

@DCubix : Can you test it ? There's major speed down ?


DCubix:
Still doesn't work for me :disappointed: 
The console says I don't have enough varyings and the version check doesn't work. I think it must be 150 not 140.

panzergame:
@DCubix : Finally it worked or not ?

panzergame:
I have to rewrite this patch to let the handy `inverse` function in a kind of header added at shader generation. In the same way the version will be passed in a macro.

youle31:
@DCubix : Hi! Can you make tests with this file to see if some syntax is supported by your computer? http://pasteall.org/blend/index.php?id=44197

(For me both builtin inverse and panzergame inverse give the same result)

DCubix:
I have OpenGL 3.0 it will not work.

youle31:
@DCubix : Sorry, I didn't wanted to make you angry. I'm trying to learn something about versions... When I say I don't understand, this is really because I'm noob... I don't want to annoy you.

PR #192, opened 2016-08-24
UPBGE: Fix crash with render to texture

youle31:
test file: http://pasteall.org/blend/index.php?id=43203


source/gameengine/VideoTexture/Texture.cpp
```diff
@@ -289,9 +293,8 @@ PyObject *Texture_close(Texture * self)
 		self->m_orgSaved = false;
 		// restore original texture code
 		if (self->m_useMatTexture) {
-			self->m_matTexture->SetBindCode(self->m_orgTex);
 			if (self->m_imgTexture) {
-				// This is requierd for texture used in blender material.
+				// This is requiered for texture used in blender material.
```
DCubix: required :smile:

PR #193, opened 2016-08-24
UPBGE: Add vibration actuator for joysticks

youle31:
Update with the last master

Don't forget you need SDL 2.0.4 to test: https://www.libsdl.org/download-2.0.php

test file: http://pasteall.org/blend/index.php?id=43204


panzergame:
@youle31 : If you rebase this branch upon ge_framerate i think you can use the following patch:
http://pasteall.org/76707/diff

This patch allow to use the common time to avoid multiple time management and including the time scale.
Also the negative event now stop the rumble, i'm very not sure if it's a proper behaviour.

youle31:
@panzergame : If you're sure that curtime is not frame dependant in all circumstances (is always real time), why not.

youle31:
All of this for a std::clock.....

youle31:
If you display messages each time someone plays a game with a pad without vibrations (other can play the same game with a pad with vibrations), there will be a problem. There is already a message written by lordloki at game engine start that tells if haptic is not supported.

youle31:
I don't understand also the purpose of RumbleStop: the code I applied is directly taken from SDL API: https://wiki.libsdl.org/CategoryForceFeedback

In my patch if you activate again the actuator, the rumble effect restart for m_duration milliseconds.

PR #198, opened 2016-08-29
UPBGE: Python API to access VertexProxy from a PolyProxy

youle31:
test file: http://pasteall.org/blend/index.php?id=43268

I thought that this was missing from the API. When you had a poly you could have vertex index but not VertexProxy. I think this is more pratical to have directly vertex proxy.

### DIFF
--- a/doc/python_api/rst/bge_types/bge.types.KX_MeshProxy.rst
+++ b/doc/python_api/rst/bge_types/bge.types.KX_MeshProxy.rst
@@ -61,6 +61,12 @@ base class --- :class:`CValue`
 
       :type: integer
 
+   .. attribute:: polygons
+
+      Returns the list of polygons of this mesh.
+
+      :type: class:`KX_PolyProxy` list (read only)
+
    .. method:: getMaterialName(matid)
 
       Gets the name of the specified material.
--- a/doc/python_api/rst/bge_types/bge.types.KX_PolyProxy.rst
+++ b/doc/python_api/rst/bge_types/bge.types.KX_PolyProxy.rst
@@ -137,3 +137,10 @@ base class --- :class:`SCA_IObject`
       :return: mesh proxy
       :rtype: :class:`MeshProxy`
 
+   .. method:: getVertices()
+
+      Returns the list of vertices of this polygon.
+
+      :return: vertex proxy list
+      :rtype: :class:`KX_VertexProxy` list
+
--- a/doc/python_api/rst/bge_types/bge.types.KX_PolyProxy.rst
+++ b/doc/python_api/rst/bge_types/bge.types.KX_PolyProxy.rst
@@ -74,6 +74,12 @@ base class --- :class:`SCA_IObject`
 
       :type: integer
 
+   .. attribute:: vertices
+
+      Returns the list of vertices of this polygon.
+
+      :type: :class:`KX_VertexProxy` list (read only)
+
    .. method:: getMaterialName()
 
       Returns the polygon material name with MA prefix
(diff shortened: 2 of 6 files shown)


BluePrintRandom:
Great!

it's a bit silly having to use the index etc to look up each vert.

is there any way to store a verts that share a point with a vert? would
this be expensive CPU wise?
I do this at the moment using KDtree**

what about vertex that share an edge?

if we know all vertex that share a point, and all edges those vertex
effect, we can then calculate the normals only for vertex effected by
moving another vertex**

panzergame:
My opinion is pretty vague at the moment. I will just replace the function `getVertices` by an attribute `vertices`. And if you estimate that the list can be very frequently accessed, then use a `CListWrapper`.

youle31:
Ok for attribute instead of method. But for clist wrapper i don't know what would be the advantages

youle31:
@panzergame : I don't remember exactly in which situation I found it useful, but I think this was with rayCast which can return a polygon.

For object, we have this inheritance tree already: 

own.meshes[0].materials[0].textures[0]

I though it would be nice to have also this for geometry:

own.meshes[0].polygon[0].vertex[0] where vertex is a vertex proxy with all the method accessibles...

panzergame:
I agree for `mesh.polygons`, nice idea :) But here it will request a CListWrapper, i will show you an example and maybe write a documentation.

youle31:
@panzergame: Now we can do own.meshes[0].polygons[0].vertices[0].setXYZ(...)

But there is just a minor issue: The names displayed don't correspond to the names set in
kx_poly_proxy_get_vertices_item_name_cb and
kx_mesh_proxy_get_polygons_item_name_cb. I don't know how to fix that.

Despite this minor issue, I think this is ready to land.

panzergame:
@youle31 : Polygons can't be searched by names. You can imply replace the name callback by NULL to disallow name research in CListWrapper.

youle31:
@panzergame: La prochaine fois fait tes changements dans ma branche pour éviter que je me casse le cul à résoudre des conflits.

DCubix:
I have Google Translator. Please not this again.

youle31:
I didn't insulted You.... I was talking to panzergame. I'm in a bas mood. This Will be better when i'll have slept... I was Just upset because This branch is here since a long time ans was mergeable. And panzergame saw i worked on it and he pushed a commit That made the branch conflicting with master. So as i'm angry today, i have bad words. But it has nothing to do with you

youle31:
Désolé pour hier. J'aurai pas dû parler comme ça. Mais ça me porte sur les nerfs de voir mon travail piétiné parfois. Comme pour ge_refraction, je suis pas du tout d'accord avec le design (le fait de mettre ior et mixFactor (d'ailleurs pourquoi tu l'as renommé ratio... enfin bon... moi je m'en fous perso de ça parce que j'ai un minimum de tolérance vis à vis du travail et de la façon de faire des autres). Mes arguments pour que ces attributs appartiennent à EnvMap et soient dans le panneau qui concerne uniquement les environment map étaient solides, mais j'ai l'impression que tu cherches à tout prix à tout contrôler, tout réécrire. Le fait qu'on ne puisse pas régler l'ior et le mix factor pour les static cubemaps est un peu bizarre. Si on fait une statique cubemap, on fait ses réglages dans le viewport, sinon on utilise les realtime cubemaps (qui permettent aussi de faire des static cubemaps...). Bref, voilà les raisons de mon agacement. Mais je m'excuse des mots que j'ai eu. Si tu préfères travailler tout seul, dis le moi franchement, et je m'arrête là.

youle31:
Après c'était une bonne idée de dire:
si mixFactor == 0 -> color = reflection
si mix factor == 1-> color = refraction
sinon color = mix reflection/refraction

Ca économise un peu de calcul si jamais on est en full refraction, on évite de calculer la réflection, donc pour ça j'étais 100% d'accord

PR #202, opened 2016-08-30
UPBGE: Implement PCF shadow and sharpness variance shadow.

panzergame:
Commit message:

PCF is a shadow mapping technique used to create smooth shadow
edges by sampling multiple time in a small zone and mix the results.

Variance sharpness is a value representing the distance between samples
when blurring the color texture of the variance shadow.

PCF depends as for simple shadow of a depth texture with coparaison
enabled. The only difference are in the UI options showed and the shader
functions used.
For the UI the user can interact with the option "samples" for the root
squared number of sampler and the option soft for the distance between
samples. For the shader two functions are used for PCF: test_shadowbuf_pcf
and test_shadowbuf_pcf_early_bail.

The first do every samples and divide the summary. The second
functions test first if the fourth furthest samples, if the values are
the same (all 0.0 or 1.0), then we discard all other samples and use these
four values. It economize a lot of process but can create artifacts for
closest shadow edges.

As PCF introduced two new long functions, "shadows_only" is simplified to
reuse a shadow value computed from an other function described before
instead of calling itself these functions.

Variance sharpness is showed in the UI under the name of "Sharpness".
if it's value is close to -1 then the shadow edges are smooth else they are
hard.

### DIFF
--- a/release/scripts/startup/bl_ui/properties_game.py
+++ b/release/scripts/startup/bl_ui/properties_game.py
@@ -831,7 +831,7 @@ class DATA_PT_shadow_game(DataButtonsPanel, Panel):
         if lamp.ge_shadow_buffer_type == "VARIANCE":
             col.prop(lamp, "shadow_buffer_bleed_bias", text="Bleed Bias")
             col.prop(lamp, "shadow_buffer_sharp", text="Sharpness")
-        elif lamp.shadow_filter == "PCF":
+        elif lamp.shadow_filter in ("PCF", "PCF_BAIL"):
             col.prop(lamp, "shadow_buffer_samples", text="Samples")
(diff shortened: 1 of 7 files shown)


panzergame:
Test file : http://pasteall.org/blend/index.php?id=43296

@youle31, @lordloki, @DCubix : Can you review the patch please ?
@Akira_San, @BPR: Can you please with the test file or with your files by just setting PCF and usable values ?

DCubix:
Tested and it's 100% working.
I'm very happy
screenshot_2

youle31:
Yeah me too. Playing with the seettings we can have very cool effects. Thanks! J - ... Before the release

PR #203, opened 2016-08-31
UPBGE: Fix bug with ImageMirror

youle31:
I can't explain mathematically why this is working but what I can say is
that each time frustum is computed, frustum.camnear takes into account
camera->GetCameraNear(). We have a bug with the file I share in pull
request. Adding camera->GetCameraNear() to frustum.camnear
solves the bug.

test file: http://pasteall.org/blend/index.php?id=43311

remark: frustum.camfar takes into account clipend value (m_clip set by default to 100.0f)

### DIFF
--- a/source/gameengine/VideoTexture/ImageRender.cpp
+++ b/source/gameengine/VideoTexture/ImageRender.cpp
@@ -288,7 +288,7 @@ bool ImageRender::Render()
 		frustum.x2 = mirrorOffset[0]+width;
 		frustum.y1 = mirrorOffset[1]-height;
 		frustum.y2 = mirrorOffset[1]+height;
-		frustum.camnear = -mirrorOffset[2];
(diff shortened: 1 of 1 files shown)


youle31:
@panzergame : Do you think we can we commit this before 0.1.0 release?

panzergame:
@youle31 : I had some render problem with the line added, but now i failed reproduce the bug in current master...

youle31:
@panzergame: HumanSheeple told me last time that he had a "flashing" bug with his mirrors so I think this bug can be considered with attention

panzergame:
The fix works, but i would like understand it.

youle31:
We should implement options to draw camera frustum (and boxes) in a clean way like in the viewport to help to debug and understand better what is obstructing the rendering

PR #209, opened 2016-09-05
Lordloki fix for python joysticks

youle31:
Lordloki fix for SCA_Python joysticks.

test file: http://pasteall.org/blend/index.php?id=43387

(When I begun my tests I had crashes because sometimes addrem[i] had not the right value after handleevents function (1 instead of -1)... I don't know why. Now I can't reproduce the bug but I don't know what was changed since I have not the bug anymore...)


lordloki:
I have tested deeper using 2 joysticks and it works ok.

BluePrintRandom:
<3 <3 <3

PR #213, opened 2016-09-10
Enable WITH_PLAYER by default.

Erstus:
Since UPBGE is focused on game engine, it makes sense to compile with Player by default.


panzergame:
done.

PR #214, opened 2016-09-11
Fix modifiers and geometry instancing.

panzergame:
This branch fix modifier rendering with geometry instancing by disabling the instancing in the shader.

To disable the instancing an uniform named `unfinstmode` is instroduced. This uniform is either 0 or 1.
Then in shader we use uniforms passed to the shader or attributs from vertex arrays. Passing the uniform was an issue because material generation look at used uniform names in nodes, but the in our case the uniform are only assigned to a varying. To solve this we force the usage of the needed uniforms:
GPU_OBJECT_MATRIX
GPU_INVERSE_OBJECT_MATRIX
GPU_LOC_TO_VIEW_MATRIX
GPU_INVERSE_LOC_TO_VIEW_MATRIX
GPU_OBCOLOR

To setup the value of `unfinstmode` the function `GPU_material_enable_instancing` need to be called through `BL_BlenderMaterial::EnableInstancing` and `KX_BlenderMaterial::EnableInstancing`.

Instancing can be disabled in materials' shaders, but the similar case is reproduced for override shaders (shadow instancing, variance shadow instancing). Here we use a functions named `RAS_IRasterizer::EnableOverrideShaderInstancing` to disable the instancing of the current bound override shader.

The both functions `KX_BlenderMaterial::EnableInstancing` and `RAS_IRasterizer::EnableOverrideShaderInstancing` are called in `RAS_IRasterizer::DrawDerivedMesh` to disable and in `RAS_DisplayArrayBucket::RenderMeshSlotsInstancing` to enable instancing.

@lordloki, @DCubix, @youle31 : Could you review this branch please ?

Example file : http://pasteall.org/blend/index.php?id=43509

### DIFF
--- a/source/blender/gpu/GPU_material.h
+++ b/source/blender/gpu/GPU_material.h
@@ -106,6 +106,7 @@ typedef enum GPUBuiltin {
 	GPU_INSTANCING_COLOR_ATTRIB    = (1 << 20),
 	GPU_INSTANCING_MATRIX_ATTRIB   = (1 << 21),
 	GPU_INSTANCING_POSITION_ATTRIB = (1 << 22),
+	GPU_INSTANCING_MODE            = (1 << 23),
 } GPUBuiltin;
 
 typedef enum GPUOpenGLBuiltin {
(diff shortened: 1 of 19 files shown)


youle31:
Hi, I haven't reviewed the code but just tested. So there is no crash so it's a good point. But performances (in my test file) are less interesting with geometry instancing enabled for objects with modifiers than normal VA or VBO. So can't the objects with modifiers use exactly the same process than non instancied objects?

Test file: http://pasteall.org/blend/index.php?id=43523

panzergame:
The major issue with modifier is that they share the same material as other meshes, using a different  material only for modifiers will assume to use a different KX_BlenderMaterial. Currently it's pretty hard to set.

youle31:
I don't understand... When I uncheck geometry instancing I have better performances. So why not forbid geometry instancing for objects with modifiers (or disable geometry instancing flag)? Sorry if I don't understand.

panzergame:
i will try an other way to solve the issue.

PR #217, opened 2016-09-13
UPBGE: Fix crash when we call Texture_Init whereas the object has no texture

youle31:
test file: http://pasteall.org/blend/index.php?id=43536

I'm not sure about the error message...

### DIFF
--- a/source/gameengine/VideoTexture/Texture.cpp
+++ b/source/gameengine/VideoTexture/Texture.cpp
@@ -209,6 +209,9 @@ static void Texture_dealloc(Texture *self)
 ExceptionID MaterialNotAvail;
 ExpDesc MaterialNotAvailDesc(MaterialNotAvail, "Texture material is not available");
 
+ExceptionID TextureNotAvail;
+ExpDesc TextureNotAvailDesc(TextureNotAvail, "Texture is not available");
+
 // Texture object initialization
 static int Texture_init(Texture *self, PyObject *args, PyObject *kwds)
 {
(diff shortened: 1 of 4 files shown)


panzergame:
The crash happen only when texID is negative or bigger than 7 ?

source/gameengine/VideoTexture/Texture.cpp
```diff
@@ -242,6 +242,10 @@ static int Texture_init(Texture *self, PyObject *args, PyObject *kwds)
 			{
 				// get blender material texture
 				self->m_matTexture = mat->GetTexture(texID);
+				if (!self->m_matTexture) {
```
panzergame: Apparently a

youle31:
A debug message could be usefull if we forgot to put a texture in the slot, but it is as you want.

panzergame:
Yeah, maybe we can reuse the try catch but with an other message error. You can look at the THRWEXCP(MaterialNotAvail, S_OK);.
You can define an error named like TextureNotAvail.

panzergame:
ge_keyboard_sensor http://download.upbge.fr/download/buildbot/branch/

youle31:
This: http://pasteall.org/79288/diff doesn't seem to work. There is no crash but no message in the console.

panzergame:
I think you have to some lines in Exception.h/cpp.
e.g:
```
extern ExpDesc TextureNotAvailDesc;
```
```
registerAllExceptions()
…
	TextureNotAvailDesc.registerDesc();
…

youle31:
Ah yes sorry I forgot it. So I tried but I still get the same error message:
Python script error - object 'Plane', controller 'Python':
  File "Text", line 10, in <module>
RuntimeError

No decription is printed... Weird. Or I missed something somewhere.

youle31:
Ok I understand it is not printed in debug mode but it is printed in release mode... So you can merge if you want

PR #219, opened 2016-09-14
UPBGE: fix call to bindprimitive when no vertex array

youle31:
fix for: https://github.com/UPBGE/blender/issues/218


panzergame:
ehh, sorry i pushed 09e36e0f77a4dfecbfa9eaed28951c2ab95215fb some hours ago which is a very similar fix, anyway thanks for the research time.

PR #221, opened 2016-09-15
UPBGE: Add an option to enable/disable casting shadow per object

youle31:
This could be used to perform kind of LOD system for shadows casting.
Performance tests could be done to see if this method saves performances
or not.

test file: http://pasteall.org/blend/index.php?id=43407
video: https://youtu.be/gGaER6iELQM

### DIFF
--- a/doc/python_api/rst/bge_types/bge.types.KX_GameObject.rst
+++ b/doc/python_api/rst/bge_types/bge.types.KX_GameObject.rst
@@ -272,6 +272,12 @@ base class --- :class:`SCA_IObject`
 
       :type: :class:`KX_BoundingBox`
 
+   .. attribute:: castShadows
+
+      This object will cast shadows or not.
+
+      :type: boolean
+
    .. attribute:: culled
 
       Returns True if the object is culled, else False.
--- a/source/gameengine/Ketsji/KX_GameObject.cpp
+++ b/source/gameengine/Ketsji/KX_GameObject.cpp
@@ -120,6 +120,7 @@ KX_GameObject::KX_GameObject(
       m_components(NULL),
       m_pInstanceObjects(NULL),
       m_pDupliGroupObject(NULL),
+      m_castShadows(true),
       m_actionManager(NULL)
 #ifdef WITH_PYTHON
     , m_attr_dict(NULL),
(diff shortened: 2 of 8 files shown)


panzergame:
youle: Do you think a per object layer system `KX_GameObject.layer` will be simpler ?

panzergame:
For example use the `KX_GameObject::[Get/Set]Layer`, normally it should work as it.

youle31:
Kxgameobject::layer? I don't understand... I Just imitated the cast shadow option as it Is in the viewport - i mean don't rendre object when we are in shadow rendering pass. Except That i simplifier it because i don't know why in viewport the option is a material option instead of an object option... I Just answered very simply to akira-san request. M'y code is very simple and i guess efficient because the geometry mightt not been rendered because the object is culled.

youle31:
About layers system, do you mean that the object would cast shadows from lights in a layer and don't cast shadows from lights in other layers?

If it's that, this is not the same as my option and could be another option eventually.

The patch I implemented disables shadows casted by this object for all lights in all layers... I think it can be merged as it is.

youle31:
I think this simple commit doesn't prevent to make another API with layers if you want (I didn't understood exactly how you wanted to design that) so I think it's ready to land.

youle31:
And this?

panzergame:
I prefered a system with object render layer:

object.layer & light.layer ---> shadow

This is planned for 0.1.3 currently but if you have time you can start working on.

youle31:
This is not the same option. Your option with layers doesn't exclude the possibility to have another option to castShadows as in the viewport. And castShadows is easy to use even if it is not per light. The system with layers would be more for advanced users

panzergame:
Imagine you have two shadows buffer, one low resolution and the other high resolution, you will want to select which buffer will be used. With a layer it's possible as the it is not global to all the shadows buffer.

youle31:
Yes it's fine... But what i say is That both options can ne added

panzergame:
What about move the castShadow in KX_BlenderMaterial as for the UI and the layer only in object ?

youle31:
There is no ui in m'y patch. I don't understand why cast shadows option was in material tab . It seems a bit weird to me to Say material.castshadow... This Is not the material That cast shadows... Object.castshadows seems to me much intuitive and avoid a for each material loop

youle31:
As it is not an important feature and as it is difficult to find overall agreement on the subject, I close this pull request.

PR #222, opened 2016-09-15
UPBGE: Add Python API for complex vibration effects

youle31:
For the moment it doesn't work well with my pad or I don't know how to use
it correctly. The API is incomplete for the moment. There are tons of
effects and settings.

test file: http://pasteall.org/blend/index.php?id=43575

@lordloki : Hi, with your pad it works fine? My pad doesn't support most of effects and conditions. Finally it acts just like with vibration actuator...

And I don't know how to use Blender threading modules but it could be fine because "thread" module doesn't work well with blender apparently.

(I made the pull request only to makes review and comments easier but it is far from ready :/)

### DIFF
--- a/doc/python_api/rst/bge_types/bge.types.KX_VibrationEffectProxy.rst
+++ b/doc/python_api/rst/bge_types/bge.types.KX_VibrationEffectProxy.rst
@@ -1,0 +1,120 @@
+KX_VibrationEffectProxy(CValue)
+===============================
+
+.. module:: bge.types
+
+base class --- :class:`CValue`
+
+.. class:: KX_VibrationEffectProxy(CValue)
+
+   Python API for complex vibration effects (based on SDL 2.0.4). Note:
+   With SDL, attributes related to durations are set in milliseconds.
+
+   .. attribute:: type
+
+      Effect type.
+
+     SDL_HAPTIC_CONSTANT     (1 << 0)
+     brief Constant effect supported.
+     Constant haptic effect.
+
+     SDL_HAPTIC_SINE         (1 << 1)
+     brief Sine wave effect supported.
+
+     SDL_HAPTIC_LEFTRIGHT    (1 << 2)
+     brief Left/Right effect supported.
+     Haptic effect for direct control over high/low frequency motors.
+ 
+     SDL_HAPTIC_TRIANGLE     (1 << 3)
+     brief Triangle wave effect supported.
+     Periodic haptic effect that simulates triangular waves.
+
+     SDL_HAPTIC_SAWTOOTHUP   (1 << 4)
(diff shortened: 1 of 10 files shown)


lordloki:
When i tested the xbox 360 pad with the sdl vibration example I can only make it to work with half of effects

lordloki:
I will try to test it later

youle31:
Thanks for tests :) I think we have to use pthread module (generally used in blender) instead of thread module. Or the functions implemented in BLI_thread... I googled pthread multiple arguments earlier and it seems that there are some example but I didn't tested yet. Maybe tomorrow.

lordloki:
I think it is better to use BLI_threads to align with blender and better compatibility with other OSs

youle31:
I tried ptread_detach (BLI_Thread seems to use pthread module) and SDL_thread_detach but I had issues in both cases... I'll give it a try again later.

panzergame:
The branch was pushed in the bug fixing cycle stage, it can't be targeted for 0.1.0, i remove the milestone.

youle31:
@lordloki  && @TwisterGE: It should work on linux now. You just have to enable WITH_CXX11 flag in cmake

youle31:
Hi guys, for this I need your opinion an help @lordloki , @panzergame and @DCubix . My tests with the API I already implemented are not very good with my gamepad. Most of the effects are not supported. We can sometimes have things interestings playing with the settings but this seems difficult to use, the API contains many functions and I don't know which are interesting, I didn't found pre_made effects to see how to have a good setup... So maybe it's better to abandon unless someone wants to continue. For my part I don't really want to continue to work on this as I haven't awesome simple default effects. So anyone wants to continue or we abandon and we consider that we can still use mini vibration actuator API to make simple effects...

lordloki:
It is the same feeling that when a tried the SDL vibration example that comes with SDL source. the most of the effects are not supported.

I think we can supply, at least, a little interface to give the duration and strenght (as we have now) but separate by each motor (LH &  RH). This way the user that wants to do more complicated things will have a basic tool to do it. 

For example, If I want to do a SINUS effect I will use a sinus function applying the LH or RH motor during a very small deltatime to simulate the SINUS effect.

youle31:
Yeah, good idea @lordloki. I don't know how to activate left or right motor but if you want to make the patch feel free to commit on this branch or elsewhere :)

youle31:
I add this to trash for now. We could still reopen it later. @lordloki : For info, BF blender official SDL version is now 2.0.4 on windows :)

lordloki:
I have the patch almost ready. Maybe i will commit this night ;)

youle31:
@lordloki : ok :)

PR #226, opened 2016-09-25
UPBGE: Add an option to enable/disable material program

youle31:
This could be used to do kind of custom material culling even if it seems there is already a culling system.
This could also be used to disable custom shaders.

test file: http://pasteall.org/blend/index.php?id=43707

performances test: http://pasteall.org/blend/index.php?id=43708 (500 to 1200 fps for 191 materials)


panzergame:
Avoiding binding a material is not very safe for the color used to render. Indeed an empty shader will depends on all the buildin shader variable: `gl_Color`, `gl_Front*`, `gl_Back*`…

@youle31 : There's a significant difference rendering without shader than with a material with the `shadless` option enabled ?

Also you should avoid calling the unbind of the material.

youle31:
This was more for custom shaders but I tested and it made no performance difference when we desactive material shader so I close the request

PR #233, opened 2016-09-28
UPBGE: Add refraction option for Realtime Cubemaps rendering

youle31:
init... I'm not sure this is the best design (for example, I don't know if
ior factor must be set in BL_Texture or in KX_CubeMap and I don't know if
it's better to have only one shader for reflections and refractions or 2
shaders... 

This needs to be Re_thought but it works :)   (edited: see below)

EDIT: Ok I moved the API for IOR factor to KX_CubeMap but I didn't tested yet.

EDIT2: About the shader, I didn't noticed any performances regressions using only one shader with cubemap reflections only without refraction (2 more uniforms:
1) use refraction (type: GPU_uniform)
2) ior factor (type: GPU_select_uniform)

Test file: http://pasteall.org/blend/index.php?id=43748

Test file for python API: http://pasteall.org/blend/index.php?id=43750

Just for eyes: http://pasteall.org/blend/index.php?id=43766


panzergame:
About refraction i was more thinking of a mapping mode like the `REFLECTION` mapping mode.
Then you will use directly a new GLSL function named something like `mtex_cube_map_refr[act]` and pass it the eta value (named eta like in the GLSL description: https://www.opengl.org/sdk/docs/man/html/refract.xhtml).
For the eta value i don't know currently what is the best place to store it, personally i prefer BL_Texture but this will assume that the value will be in texture UI, but this statement is not certain.

For the mix of reflection and refraction, the user can define the second mixed texture under an `Image File` environment map and change the mapping.

youle31:
I don't agree. when we mix the 2 shaders we have better performances when the 2 are activated because we don't have to pass extra uniforms. I made performance tests and this is the better solution.

About UI, it is not possible to make a refraction option because refraction shader needs TEXCO_REFL flag so in my opinion this is the best design. I already made test with other designs and it was not esailly possible.

youle31:
For this I'm wondering if we implement mixFactor API: mix beetween reflection/refraction.

Else after relections and test, I think the design is how it has to be:

- We can have full reflections (we don't check refraction option) and I remarked no performance impact with the modifications I made (3 float/integer uniforms more->no performance impact it seems).
- If we want to make a mix beetween reflection/refraction we have mixFactor option and we use only one texture and one cubemap render.
- If we want to use only refraction, we set mixFactor to 1.0 and the extra operations (that won't be used in the shader) are very minimal (and it's done on GPU so it's very fast and that won't cost performances).

About the UI, I think it's better to leave refraction option only available for cubemaps as we don't have reflection/refraction shaders for other types of textures. (So iorFactor and useReflection are in EnvMap struct)

So I think my patch is good but there can be a discussion about variable names, comments, mixFactor API.

youle31:
updated test file: http://pasteall.org/blend/index.php?id=43838

youle31:
Can we merge this?

PR #239, opened 2016-10-02
Begin to work on planar reflections

youle31:
test file: http://pasteall.org/blend/index.php?id=43796

test file 2: http://pasteall.org/blend/index.php?id=43800 (reflection inverted) -> This is normal, it works like the code in ImageRender, the uv has to be scaled by -1 on X axis.

Test file 3: http://pasteall.org/blend/index.php?id=43807 (To test reflection shader deformed with normal map)

test file4: http://pasteall.org/blend/index.php?id=43809 (to work on planar reflections deformed by normal maps and compare with how the deformation is done for cubemaps)

And we inherit from the same bug (1 frame lag) (EDIT: in fact no...)

Maybe we could make a shader to fix UV and take as input shi->vn (view normal) to take into account normal deformations from normal textures in previous slots -> deform reflections.

We'll also need a shader for refraction with shi->vn and IOR as input.

### DIFF
--- a/release/scripts/startup/bl_ui/properties_texture.py
+++ b/release/scripts/startup/bl_ui/properties_texture.py
@@ -928,6 +928,32 @@ class TEXTURE_PT_ocean(TextureTypePanel, Panel):
         col.prop(ot, "ocean_object")
         col.prop(ot, "output")
 
+class TEXTURE_PT_image_planar(TextureTypePanel, Panel):
+    bl_label = "Planar Refle/ractions"
+    bl_options = {'DEFAULT_CLOSED'}
+    tex_type = 'IMAGE'
+    COMPAT_ENGINES = {'BLENDER_GAME'}
+
+    def draw(self, context):
+
+        layout = self.layout
+
+        idblock = context_tex_datablock(context)
+        tex = context.texture
+
+        split = layout.split()
+
+        row = split.row()
+        row.label(text="Planar Type:")
+        row.prop(tex, "planar_type")
+
+        row = layout.split()
+        row.label(text="Clipping:")
+        row.prop(tex, "planar_clip_start", text="Start")
+        row.prop(tex, "planar_clip_end", text="End")
+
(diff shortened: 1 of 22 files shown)


youle31:
@DCubix : Hi, I didn't understood your comments in the last commit.
I explain what I wanted to make:

- In the UI, for Image or Movies, I added a section: planar refl/ractions. For a single texture you choose either to make a reflection or to make a refraction but you can't do both at the same time. Why? Because as you said we need 2 textures if we want to make reflection AND refraction.
- So if we choose to make a reflection, a KX_Planar(Tex *tex, KX_GameObject *viewpoint, RAS_IPolyMaterial *polymat, int type, int width, int height) will be rendered. "type" will be either reflection OR refraction. For the moment I only copied the code from ImageMirror so it is only for reflections (I didn't implement all changes we have to make to render in the ooposite direction of reflection yet).
- If we choose to make refraction render, this will be in another texture slot.
- If we want to have both reflections and refractions on the same plane, we'll need 2 textures (1 for reflection and 1 for refraction as in Martinsh water demo). The blending beetween the 2 textures will be done with blend types and influence factors.

Why do you say that the planar reflection render is not correct? It is the same as ImageMirror. What I wanted to do is just implement the possibility for the user to use a normal map to deform reflection. So instead of using "mtex_image", I just wanted to modify a bit "mtex_image" in a new shader.

I hope that my explainations are more clear.

DCubix:
I said that since ImageMirror already has correct reflection, my method didn't work correctly because of that.
You can revert this is you want.

I thought you were going to make the same thing you did for real-time cubemaps.

youle31:
It's the same than for cubemaps in the principle. You set frustum, You set rasterizer and camera matrix to be oriented correctly, You do culling, and You render background and scene. All of This with fbo system...

DCubix:
Yeah, you need 2 textures, and render 2 times. 1 up for reflection and 1 down for refraction

PR #243, opened 2016-10-04
Remove UV shift parallax option.

panzergame:
I propose you to remove this option because none valid result are found when setting a value different to 0.0.

@youle31, @lordloki, @DCubix : Do you have a case showing the utility of this variable ? Else do you agree to remove it ?

### DIFF
--- a/release/scripts/startup/bl_ui/properties_texture.py
+++ b/release/scripts/startup/bl_ui/properties_texture.py
@@ -1082,7 +1082,6 @@ class TEXTURE_PT_influence(TextureSlotPanel, Panel):
                 factor_but(col, "use_map_hardness", "hardness_factor", "Hardness")
 
                 col.label(text = "Parallax:")
-                factor_but(col, "use_parallax_uv", "parallax_uv_shift", "Height Scale")
                 factor_but(col, "use_map_parallax", "parallax_steps", "Steps")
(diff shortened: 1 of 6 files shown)


youle31:
I agree

DCubix:
I agree

lordloki:
I'm not agree. The "Height Scale" parameters works in opposite to "Bump" parameter. The Bump parameter gives deep to the valley while Height parameter place the position  where the displacement is applied (move the top part higher and centered). This also prevents that others close textures be affected.

Instead of this change, I would like to remove the check tick at texture panel (use_parallax_uv) and rework the parallax panel (modifying tip texts and labels). As I did here https://developer.blender.org/D1762

youle31:
The problem with that option uv_shift is that we can't control in which direction the "shift" is made. It's in diagonal in only one direction. This is seems a bit weird to me. But I say that because I never had to use this option. If @lordloki says that it can be useful in some cases and that the uv shift is to make the top part more "centered", we could improve tooltips and names because this is not very clear.

Else I agree too with the changes made by lordloki in the patch for BF Blender

panzergame:
@lordloki : I understand, i think the Heigh Scale is borken at the moment. I will try to use your UI and fix the Heigth Scale.

PR #244, opened 2016-10-04
Material switching.

panzergame:
This branch allow the user to replace a material by calling in python:
```
mesh.replaceMaterial(index, material)
```

Then the material in the slot `index`will be replaced by the material `material`.

When the material is replaced the RAS_DisplayArrayBucket are moved to the RAS_BucketMaterial of the new material, all the mesh slot in this DAB reset their pointer to the material bucket and the attribute's layers are recomputed to follow the new material.

@lordloki, @DCubix, @sdfgeoff: Do you agree with the API and the patch ? Could you make some test to find any possible bugs ? Thanks in advance.

### DIFF
--- a/doc/python_api/rst/bge_types/bge.types.KX_MeshProxy.rst
+++ b/doc/python_api/rst/bge_types/bge.types.KX_MeshProxy.rst
@@ -132,3 +132,16 @@ base class --- :class:`CValue`
       :arg uv_index_from: optional uv index to copy from, -1 to transform the current uv.
       :type uv_index_from: integer
 
+   .. method:: replaceMaterial(matid, material)
+
+      Replace the material in slot :data:`matid` by the material :data:`material`.
+
+      :arg matid: The material index.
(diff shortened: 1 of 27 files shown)


youle31:
Congrats panzergame! I just tested with this file: http://pasteall.org/blend/index.php?id=43839 and it looks to work fine :) Would it be easy to add an optional argument to replace texture at the same time than material?

youle31:
A weird effect with this file: http://pasteall.org/blend/index.php?id=43861 (the texture can change if the material we change had a texture)

panzergame:
@youle31 : The texture is containing the textures, if you change it you change the texture. Did i answered correctly and understand the question ?

youle31:
I meant:
- When you change a material that has no texture with a material that has texture, the new material has no texture.
- When you change a material that has a texture with a material that has texture, the new material has the new texture.

youle31:
For me it's good.

PR #245, opened 2016-10-05
UPBGE: Add support of rolling friction.

panzergame:
Rolling friction is a friction value used of rounded shapes
when only one collision contact point exists.
It slowdown the rotation of a sphere rolling in a plane.

This value is exposed in the UI under the name "Rolling Friction"
between 0 and 100.

Test file: http://pasteall.org/blend/index.php?id=43841


youle31:
Hehe, that's very cool :) It works even if sphere collision bounds is not checked.

PR #247, opened 2016-10-05
KX_FontObject text measure

DCubix:
Add the possibility to get the width and the height
	of the current text in a KX_FontObject in Blender Units.

This is useful for text alignment like centering, justifying, etc...
Just do `KX_FontObject.textDimensions` and it will return a Vector of size 2 containing the width and the height.

Demo: http://pasteall.org/blend/index.php?id=43868

### DIFF
--- a/doc/python_api/rst/bge_types/bge.types.KX_FontObject.rst
+++ b/doc/python_api/rst/bge_types/bge.types.KX_FontObject.rst
@@ -50,3 +50,10 @@ base class --- :class:`KX_GameObject`
 
       :type: float (0.0001 to 40.0)
 
+   .. method:: getTextSize()
+
+      Gets the size (width and height) of the current text in Blender Units.
+
+      :return: A tuple containing the width and the height.
+      :rtype: tuple(2)
+
--- a/doc/python_api/rst/bge_types/bge.types.KX_FontObject.rst
+++ b/doc/python_api/rst/bge_types/bge.types.KX_FontObject.rst
@@ -50,10 +50,9 @@ base class --- :class:`KX_GameObject`
 
       :type: float (0.0001 to 40.0)
 
-   .. method:: getTextSize()
+   .. attribute:: textDimensions
 
       Gets the size (width and height) of the current text in Blender Units.
 
-      :return: A tuple containing the width and the height.
-      :rtype: tuple(2)
+      :type: :class:`mathutils.Vector`
 
--- a/source/gameengine/Ketsji/KX_FontObject.cpp
+++ b/source/gameengine/Ketsji/KX_FontObject.cpp
@@ -36,6 +36,7 @@
 #include "KX_Globals.h"
 #include "KX_PythonInit.h"
 #include "KX_TextMaterial.h"
+#include "KX_PyMath.h"
 #include "BLI_math.h"
 #include "EXP_StringValue.h"
 #include "RAS_IRasterizer.h"
@@ -53,7 +54,6 @@
 extern "C" {
 #include "BLF_api.h"
 }
-#include "python_utildefines.h"
 
 #define BGE_FONT_RES 100
 
@@ -106,7 +106,7 @@ KX_FontObject::~KX_FontObject()
 	//it's handled in KX_Scene::NewRemoveObject
 }
 
-MT_Vector2 KX_FontObject::GetTextSize()
+const MT_Vector2 KX_FontObject::GetTextDimensions()
 {
 	float width = 0, height = 0;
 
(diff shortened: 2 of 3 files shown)


doc/python_api/rst/bge_types/bge.types.KX_FontObject.rst
```diff
@@ -50,3 +50,10 @@ base class --- :class:`KX_GameObject`
 
       :type: float (0.0001 to 40.0)
 
+   .. method:: getTextSize()
```
panzergame: I prefer use an attribute instead of function: `font.textSize`. Also we already have an attribute named `size`, maybe you should use something like:
```
font.dimension
font.textDimension
```
If `font.text…`is not redundant.

source/gameengine/Ketsji/KX_FontObject.cpp
```diff
@@ -105,6 +106,32 @@ KX_FontObject::~KX_FontObject()
...
+		width = std::max(width, w);
+		height += h + m_line_spacing;
+	}
+
+	// XXX: Quick hack to convert the size to BU
+	width /= 10;
+	height /= 10;
+
+	// Scale the width and height by the object's scale
+	MT_Vector3 localScale = this->GetSGNode()->GetLocalScale();
```
panzergame: the `this->`is optional

source/gameengine/Ketsji/KX_FontObject.cpp
```diff
@@ -278,6 +306,7 @@ PyTypeObject KX_FontObject::Type = {
 };
 
 PyMethodDef KX_FontObject::Methods[] = {
+	{ "getTextSize", (PyCFunction)KX_FontObject::sPyGetTextSize, METH_NOARGS },
```
panzergame: Remove extra space around  `{}`.

source/gameengine/Ketsji/KX_FontObject.cpp
```diff
@@ -290,6 +319,17 @@ PyAttributeDef KX_FontObject::Attributes[] = {
 	{NULL}    //Sentinel
 };
 
+PyObject *KX_FontObject::PyGetTextSize()
+{
+	MT_Vector2 size = GetTextSize();
+	PyObject *ret;
+	ret = PyTuple_New(2);
+	PyTuple_SET_ITEMS(ret,
```
panzergame: You could return a python vector using `PyObjectFrom(vector2)`.

source/gameengine/Ketsji/KX_FontObject.h
```diff
@@ -45,6 +45,8 @@ class KX_FontObject : public KX_GameObject
 
 	virtual ~KX_FontObject();
 
+	MT_Vector2 GetTextSize();
```
panzergame: I'm a big fan of `const` :)
you could write:
```
const MT_Vector& Get…();
```

The function is not constant because some BLF operations in it are certainly not.

panzergame:
Good idea :)

source/gameengine/Ketsji/KX_FontObject.cpp
```diff
@@ -105,6 +106,32 @@ KX_FontObject::~KX_FontObject()
 	//it's handled in KX_Scene::NewRemoveObject
 }
 
+const MT_Vector2 KX_FontObject::GetTextDimensions()
+{
+	float width = 0, height = 0;
```
panzergame: Follow the blender code style about float write; use 0.0f same think for  / 10 --> / 10.0f

DCubix:
I fixed my code, i think it's ok now. It's weird that MT_Vector2 doesn't accept multiplication by another MT_Vector2...

doc/python_api/rst/bge_types/bge.types.KX_FontObject.rst
```diff
@@ -50,3 +50,9 @@ base class --- :class:`KX_GameObject`
 
       :type: float (0.0001 to 40.0)
 
+   .. attribute:: textDimensions
+
+      Gets the size (width and height) of the current text in Blender Units.
```
youle31: We could maybe add "during runtime"? Oh it's not so important

source/gameengine/Ketsji/KX_FontObject.cpp
```diff
@@ -105,6 +106,32 @@ KX_FontObject::~KX_FontObject()
...
+		dimensions.x() = std::max(dimensions.x(), w);
+		dimensions.y() += h + m_line_spacing;
+	}
+
+	// XXX: Quick hack to convert the size to BU
+	dimensions /= 10.0f;
+
+	// Scale the width and height by the object's scale
+	MT_Vector3 scale = NodeGetLocalScaling();
+	dimensions.x() *= scale.x();
```
youle31: fabs(scale.x()) ? I don't know when the object has a negative scale on X axis for example it returns (-11.123, 2.12)
The dimensions shouldn't always be positive?

youle31:
It's ok to me. However I left 2 comments that are more questions. It's cool to use PyObjectFrom so we can do textDimensions.x and .y .

PR #249, opened 2016-10-06
UPBGE: Forbid setParent to an object in an inactive layer

youle31:
Why this behavior is preferable?

- Because of references to objects in the scene (if you add an object of another layer in the scene it won't be referenced as other objects in the scene and I guess this could lead to issues with reference count and endobject...)

- A proposed alternative was set find an instanced object in the scene and setPrent to this instance instead of parenting to the object in the inactive layer. This is not a good option in my opinion because at which instance of the object will be parented? We can't choose randomly an instance and parent the object to that instance.

So the better alternative in my opinion is to forbid parenting object to another in an inactive layer and to display a warning in the console.

test file: http://pasteall.org/blend/index.php?id=43857

### DIFF
--- a/source/gameengine/Ketsji/KX_GameObject.cpp
+++ b/source/gameengine/Ketsji/KX_GameObject.cpp
@@ -3530,6 +3530,15 @@ PyObject *KX_GameObject::PySetParent(PyObject *args)
 	}
 	if (!ConvertPythonToGameObject(logicmgr, pyobj, &obj, true, "gameOb.setParent(obj): KX_GameObject"))
 		return NULL;
+
+	if (obj->GetScene()->GetInactiveList()->SearchValue((CValue *)obj)) {
+		PyErr_SetString(PyExc_ValueError, "Warning: you are trying to parent an object to another in an inactive layer\n"
+			"Please use the exact reference to the object/instance you want to parent your object to\n"
+			"Example: Instead of using setParent('string'), use setParent(myInstancedObject) where myInstancedObject\n"
+			"is a KX_GameObject in an active layer.");
+		return NULL;
+	}
+
 	if (obj)
 		SetParent(obj, addToCompound, ghost);
 	Py_RETURN_NONE;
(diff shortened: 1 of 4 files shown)


panzergame:
I'm totally agree as the conversion manage it in a pretty similar way: move the children in the layer of the parent object.

@youle31 : About the patch i was more thinking disable inactive to active and inverse, which will still allow to parent an inactive object to an other inactive objects.
In this case you might use:
```
if (parentinactive != childinactive) {
     // error
     return NULL;
}
```

Also you have seen that we use in a lot of different place to find if an object is inactive. Maybe you can work on adding an attribute in `KX_GameObject` accessed by a function named like `GetInactive()` and set by `SetInactive`. This `inactive` value will be set in the conversion for inactive object and set to false (active) when an object is added in a scene.
By this way you can speedup a lot of lookup to know if an object is inactive.

youle31:
@panzergame: I agree with your suggestion for attribute/method object->IsInActiveLayer(). But finally I think it should be done in another patch. And I'd maybe need help for this because I don't know all the cases when we'll have to say isInActiveLayer = true; isInActiveLayer = false;

So I propose to keep only the first commit even if it's not 100% exact.

About parenting object in an inactive layer to another object in the same inactive layer, maybe it would be safe to forbid that too.

source/gameengine/Ketsji/KX_GameObject.cpp
```diff
@@ -3530,6 +3530,14 @@ PyObject *KX_GameObject::PySetParent(PyObject *args)
 	}
 	if (!ConvertPythonToGameObject(logicmgr, pyobj, &obj, true, "gameOb.setParent(obj): KX_GameObject"))
 		return NULL;
+
+	if (obj->GetScene()->GetInactiveList()->SearchValue((CValue *)obj) ||
```
panzergame: I think we can allow to parenting an inactive object to another inactive object. In this case the condition will looks like:
```
if (scene->GetInactiveList()->SearchValue(obj) != scene->GetInactiveList()->SearchValue(this))
```
Which will be false if the parent is inactive and the child active or if the parent is active and the child inactive.
Commonly if they are not in the same state.

Also the condition should be moved in `KX_GameObject::SetParent`, because this function is called from python and from the actuator too.

youle31:
test file: http://pasteall.org/blend/index.php?id=43998

PR #252, opened 2016-10-07
UPBGE: Add unlimited Planar Reflections/Refractions Old branch

youle31:
I make another pull request with this branch to have better visibility of the code

### DIFF
--- a/release/scripts/startup/bl_ui/properties_texture.py
+++ b/release/scripts/startup/bl_ui/properties_texture.py
@@ -958,6 +958,7 @@ class TEXTURE_PT_image_planar(TextureTypePanel, Panel):
 
         row = layout.split()
         row.prop(tex, "use_planar_auto_update")
+        row.prop(tex, "use_planar_reflect_cull")
 
 class TEXTURE_PT_mapping(TextureSlotPanel, Panel):
     bl_label = "Mapping"
--- a/release/scripts/startup/bl_ui/properties_texture.py
+++ b/release/scripts/startup/bl_ui/properties_texture.py
@@ -932,7 +932,7 @@ class TEXTURE_PT_image_planar(TextureTypePanel, Panel):
     bl_label = "Planar Reflection / Refraction"
     bl_options = {'DEFAULT_CLOSED'}
     tex_type = 'IMAGE'
-    COMPAT_ENGINES = {'BLENDER_GAME'}
+    COMPAT_ENGINES = {'BLENDER_RENDER', 'BLENDER_GAME'}
 
     def draw(self, context):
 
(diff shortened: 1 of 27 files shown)


youle31:
@martinsh : I think it's ok now for clip planes for reflections. I didn't checked for refraction yet. Now you can have several planes with reflections.

youle31:
test file for last commit about backface culling issues: http://pasteall.org/blend/index.php?id=43939

parenthesis: I remarked a slight performance decrease (860 fps against 880 before) with the test file above

youle31:
test file: http://pasteall.org/blend/index.php?id=43945 (to test avoid reflections of other mirrors on a mirror)

For me it's stable enough (I fixed many crashes and free errors. I didn't really tested python API) to begin review, clean up and refactor eventually to have a commn base with cubemaps.

Some changes be done about names and tooltips.

panzergame:
I'm thinking about changing the python API to receive the planar.

1) The object rendering to the texture will be accessed at:
```BL_Texture.render
or
BL_Texture.probe```
I don't know which is the best name, probe is fine for what we are doing: only rendering the scene from a position, but later maybe it will not be the main case, then `render`will be a better name…

2) The cube map and planar are using the same base python class:
```KX_TextureRenderer
or
KX_TextureProbe```

```KX_CubeMap(KX_Texture…)
KX_PlanarMap(KX_Texture…)
```

Not that the class rendering the planar reflection is named `KX_PanarMap` and not only `KX_Planar` to avoid confusions. It could be renamed differently like `KX_PlanarRenderer` but this will assume to also rename `KX_CubeMap`.

youle31:
Hey, manu thanks for review.

My point is That sharing code to avoid repetitions could make the things harder to read for developers, debug for developers, and this lead to this kind of situation where it can also make things less readable and understandable for the user. For me, neither probe, neither render are explicit enough. For me the more intuitive for the user is: Texture.cubeMap.attr/method and texture.planarMap.attr/method. Else if your design choice is already made, I'd say: Texture.imageRendered.attr/method. Because the final goal would be to have all features of imageRender in UI. In UI, I put imageRender/mirror because I thought later we could have normal/mirror/viewports/video... If all this things have enough things in common, we could keep texture.imageRendered.attr/method. Else I'd add another sublevel and in this case we can be more explicit. Texture.imagerendered.mirror.attr/mthod. Texture.imagerendered.cubemap.attr/method, Texture.imagerendered.video.attr/method,
Texture.imagerendered.planarmap.attr/method... I think this is an important choice and that the API should more or less follow the design of the Ui (even if it is not totally possible because cubemap is in environment map tab and the others in image/imagerender (according to the UI I did). This needs to be thought with imagerender integration in the UI imho

panzergame:
Moved to #390.

youle31:
I reopen just to do the code comparison easier.

PR #253, opened 2016-10-07
Pass rendered camera to scene.pre_draw/pre_draw_setup callbacks.

panzergame:
Previously the user was able to altera	te rendering datas through
the callbacks scene.pre_draw/post_draw/pre_draw_setup. But he
wasn't able to know which camera was used when this callbacks
was called in case of multiple viewports.
Getting this info can help on setting an UI with a more modular
camera viewport management.

To solve this issue the scenes callbacks, exepted post_draw because
this one is not per camera, take as optional argument the current
rendered camera.

Internally the function KX_Scene::RunDrawingCallbacks takes
as first argument an enumeration coresponding to the callbacks list
instead of the callbacks list previously. This modification allow
to remove the two getter to the python callbacks list from
KX_Scene.

Example file: http://pasteall.org/blend/index.php?id=43884
In this file we can determine which viewport is currently rendered by the camera name and then decide of the world horizon color.

### DIFF
--- a/doc/python_api/rst/bge_types/bge.types.KX_Scene.rst
+++ b/doc/python_api/rst/bge_types/bge.types.KX_Scene.rst
@@ -127,7 +127,7 @@ base class --- :class:`PyObjectPlus`
 
    .. attribute:: pre_draw
 
-      A list of callables to be run before the render step.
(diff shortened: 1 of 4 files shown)


DCubix:
Seems alright to me :smile:

PR #263, opened 2016-10-17
UPBGE: Add objects LOD option while we render Cube Maps

youle31:
This was not easy because I had to replicate all the existing LOD system
(except the viewport part) to create a new one because an object can have
LOD for normal render pass computed according to distance from the active
camera and another LOD while we render cubemaps computed according to the
distance from the cubemap rendering camera.

The patch might be simplified and cleaned up. This was just a test to
answer Maujoe feature request. A long copy/paste session.

I don't know if this patch is a good idea because this needs to replace
mesh when we begin cubemap rendering pass and restore mesh after rendering
pass. This might be costfull in terms of performances. I don't know.

But maybe in some cases it could be useful. The debate is opened.

I make a pull request but this is low priority. It's just to have better visibility of the code but the pull request can be removed if we estimate that this is not a good idea.

test file: http://pasteall.org/blend/index.php?id=43980

Perf test: http://pasteall.org/blend/index.php?id=43981 very weird behaviour (when the cubemap object is at the same position than the sphere)

video: https://www.youtube.com/watch?v=y12cKwfm5w0

### DIFF
--- a/release/scripts/startup/bl_operators/object.py
+++ b/release/scripts/startup/bl_operators/object.py
@@ -1034,3 +1034,159 @@ class LodGenerate(Operator):
         scene.objects.active = ob
 
         return {'FINISHED'}
+
+class CubeMapLodByName(Operator):
+    """Add levels of detail to this object based on object names"""
+    bl_idname = "object.lod_by_name"
+    bl_label = "Setup Levels of Detail By Name"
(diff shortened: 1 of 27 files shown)


panzergame:
Currently ge_cubemap is introducing lod update and lod distance factor for cube map render.

PR #265, opened 2016-10-17
UPBGE: Add culling for cubemap object itself. See comments.

youle31:
test file: http://pasteall.org/blend/index.php?id=43982

video: https://www.youtube.com/watch?v=y12cKwfm5w0 (same video than in the pull request about LOD but we can see the issue about rendering cubemaps even if cubemap object is outside camera frustum)


panzergame:
There's some issues with this patch:
- the culling state tested is the one of the last frame which is generally set to false at the end of the final render.
- if the viewpoint isn't the object using the cube map, then the cube map object can be visible without any render.
- if the viewpoint is an empty object, which is always set to culled the render will be broken.

I think the user can reproduce this optimization by setting the viewpoint object in a layer not used by the cube map at rendering.

youle31:
Yes you're right. I think a bit if i can find a better system. I close This request

PR #266, opened 2016-10-19
UPBGE: Prevent cubemap rendering if the cubemap object is culled at previous frame

youle31:
See comments in KX_CubeMapManager.cpp

Test file: http://pasteall.org/blend/index.php?id=43999

### DIFF
--- a/source/gameengine/Converter/BL_BlenderDataConversion.cpp
+++ b/source/gameengine/Converter/BL_BlenderDataConversion.cpp
@@ -1800,7 +1800,7 @@ void BL_ConvertBlenderObjects(struct Main* maggie,
 							}
 						}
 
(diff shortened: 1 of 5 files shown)


youle31:
Can we commit this? @panzergame : I remember you made a comment about layers but I didn't understand.

youle31:
the issue with this patch is that we assume that 1) only one object is using the cube map 2) this object will keep the same visible state at the next frame

youle31:
Wow in fact there is a more important issue: The cubemap object is considered as culled when only a small part of the object is culled :/

youle31:
I close the request for now waiting we find a batter solution...

youle31:
ok this is caused by CalculateVisibleMeshes in RenderShadowBuffers.

PR #269, opened 2016-10-21
UPBGE: Add an API to access filter pass texture bindcode

youle31:
This allows us to get the render from GLSL 2D Filter pass 0 and add it as
input in the GLSL 2D Filter pass 1 for example etc...

test file: http://pasteall.org/blend/index.php?id=44021

It doesn't work well for now...

### DIFF
--- a/source/blender/gpu/intern/gpu_shader.c
+++ b/source/blender/gpu/intern/gpu_shader.c
@@ -535,7 +535,7 @@ void GPU_shader_free(GPUShader *shader)
 
 int GPU_shader_get_uniform(GPUShader *shader, const char *name)
 {
-	return glGetUniformLocation(shader->program, name);
+	return shader? glGetUniformLocation(shader->program, name) : -1;
 }
 
 void *GPU_shader_get_interface(GPUShader *shader)
--- a/source/gameengine/Rasterizer/RAS_2DFilter.cpp
+++ b/source/gameengine/Rasterizer/RAS_2DFilter.cpp
@@ -47,7 +47,8 @@ static char predefinedUniformsName[RAS_2DFilter::MAX_PREDEFINED_UNIFORM_TYPE][40
 RAS_2DFilter::RAS_2DFilter(RAS_2DFilterData& data)
 	:m_properties(data.propertyNames),
 	m_gameObject(data.gameObject),
-	m_uniformInitialized(false)
+	m_uniformInitialized(false),
+	m_colorBindCode(-1)
 {
 	for(unsigned int i = 0; i < TEXTURE_OFFSETS_SIZE; i++) {
 		m_textureOffsets[i] = 0;
@@ -82,9 +83,18 @@ void RAS_2DFilter::Initialize(RAS_ICanvas *canvas)
 	}
 }
 
+int RAS_2DFilter::GetColorBindCode()
+{
+	return m_colorBindCode;
+}
+
 void RAS_2DFilter::Start(RAS_IRasterizer *rasty, RAS_ICanvas *canvas, unsigned short depthfbo,
 						 unsigned short colorfbo, unsigned short outputfbo)
 {
+	if (m_colorBindCode == -1) {
+		m_colorBindCode = rasty->GetOffscreenColorBindCode(outputfbo);
+	}
+
 	rasty->BindOffScreen(outputfbo);
 
 	if (Ok()) {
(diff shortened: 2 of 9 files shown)


source/gameengine/Rasterizer/RAS_2DFilter.cpp
```diff
@@ -82,9 +83,18 @@ void RAS_2DFilter::Initialize(RAS_ICanvas *canvas)
...
 
+int RAS_2DFilter::GetColorBindCode()
+{
+	return m_colorBindCode;
+}
+
 void RAS_2DFilter::Start(RAS_IRasterizer *rasty, RAS_ICanvas *canvas, unsigned short depthfbo,
 						 unsigned short colorfbo, unsigned short outputfbo)
 {
+	if (m_colorBindCode == -1) {
```
panzergame: As you can see in this function the off screen are switched very often, will the input off screen id fine for the user ?
Also this input off screen can be used as a output of the next filter (ping pong).

youle31: Ah ok. Maybe it's the reason why it doesn't work properly for now. Thanks!

source/blender/gpu/intern/gpu_shader.c
```diff
@@ -535,7 +535,7 @@ void GPU_shader_free(GPUShader *shader)
 
 int GPU_shader_get_uniform(GPUShader *shader, const char *name)
 {
-	return glGetUniformLocation(shader->program, name);
+	return shader? glGetUniformLocation(shader->program, name) : -1;
```
panzergame: ehhh, it's for what ?

panzergame:
So if i understand fine, you want to have three filters: A, B and C. A output to the input B and A + B are used in the filter C as input ?

```
A --> B --> C
 |                 ^
 |--------------|
```

source/blender/gpu/intern/gpu_shader.c
youle31: It's a temp fix for filter.setTexture(0, bindcode, name)... I already had this error but I though it was fixed. The shader can apparently be NULL so when we call GetUniformLocation in setTexture we have a crash. Iirc you had linked glsl shader at the first frame to fix this bug but maybe it has been refactored and the fix disappeared...

youle31:
Yes or just 2 filters. The output of A could be used in B. This works more or less. It seems there is a problem with bgl_RenderedTexture -> bgl_RenderedTexture doesn't return always the texture of the main fbo render non modified by GLSL filters. But this texture in in the texture bound. In my test file, its bindcode is 11 but I don't know how to assign it to bgl_RenderedTexture. test file: http://pasteall.org/blend/index.php?id=44057 : rtex should be bgl_RenderedTexture. butt bgl_RenderedTexture is wrong. The file shows the expected result

DCubix:
You can have a new uniform called bgl_PreviousPass or bgl_PreviousTexture
Which can be black for the first pass.

youle31:
In fact, normally we don't really need this because we can access the fbo texture color bindcode of the filter so we can make filter2.setTexture(1, filter1.colorBindCode, "tex")... The real problem it seems is bgl_RenderedTexture that change from one filter to another....

panzergame:
The content of `bgl_RenderedTexture` changed or it's bind code ?

youle31:
I'm not sure. I thought that was the content of bgl_RenderedTexture that changed. But in fact I don't know how the mix is done currently with previous and next filter... I'll do more tests to see.

panzergame:
bgl_RenderedTexture is the output of what was before the filter : the frame of an other filter. So you can't have two bgl_RenderedTexture identique.

youle31:
ah ok. But the initial Rendered Texture is still in the textures bound. bgl_RenderedTexture shouldn't be always the initial texture?

panzergame:
No, to allow accumulate filter on other.

youle31:
I don't know exactly how to design it, but my goal was to have several filters, render this filters to textures but don't display it at the screen and take into account only the last pass (in the last pass take as input all the filters rendered to textures before)

panzergame:
ohhh, i see :) A kind of render layers. You can design for each filter an input/output layer mask. I don't have a very clear idea currently.

youle31:
yeah neither me... I need to think about

youle31:
I close this pull request waiting to have a better idea on how to design it

PR #271, opened 2016-10-22
Ge vibration leftright

lordloki:
Vibration Actuator improvement

Now the vibration actuator has 2 methods (start and stop). Start method starts the vibration and also it can update the parameters of the vibration on fly. Stop method stops the vibration.

Now there is 2 strenght parameters the "left" one sets the strenght of the Low frequency joystick's motor and the "right" one sets the strenght of the High frequency joystick's motor.

Additionaly, 2 new bool attributes have been included (isVibrating and hasVibration) along the new python methods.

I will upload a test blend later


source/gameengine/GameLogic/SCA_VibrationActuator.cpp
```diff
@@ -65,13 +68,35 @@ bool SCA_VibrationActuator::Update()
 	bool bPositiveEvent = IsPositiveEvent();
 
 	if (bPositiveEvent) {
-		instance->RumblePlay(m_strength, m_duration);
-		m_endtime = PIL_check_seconds_timer() * 1000.0f + m_duration;
+		switch (m_mode) {
+			case KX_ACT_VIBRATION_PLAY:
+			{
+				float strength[2];
```
panzergame: Is it an advantage to use an array as argument in `RumblePlay` if we have to pack the array in 3 lines before ?

youle31:
Hi, it looks good to me, however there is 2 specific cases where I'm wondering if the actual behaviour is the right behaviour.

1) When we call cont.deactivate("VibrationPlay"), should it stop vibration ?
2) When we call cont.activate("VibrationStop"), the value returned by SCA_VibrationActuator::Update shouldn't be false? ->e.g : http://pasteall.org/blend/index.php?id=44049 you can test the value returned by Update function with an "actuator" Sensor.

I don't know if it's feasible or not. But for a normal usage, this is perfect :)

lordloki:
@youle31, I'm gonna compare it against the sound actuator (this is the more similar actuator by concept) to mimic its behaviour.

Thanks for testing.

youle31:
in Update for the 1) 
if (IsNegativeEvent()) {
		instance->RumbleStop();
		return false;
	}

lordloki:
@youle31 
For 1) I prefer left the vibration without stop (until it finishes) when the actuator is deactivated. It follows the same convention that the sound actuator.

source/gameengine/GameLogic/SCA_VibrationActuator.cpp
```diff
@@ -62,16 +64,32 @@ bool SCA_VibrationActuator::Update()
...
 
 	RemoveAllEvents();
 
-	return PIL_check_seconds_timer() * 1000.0f < m_endtime;
+	if (!(PIL_check_seconds_timer() * 1000.0f < m_endtime)) {
+		instance->RumbleStop();
+		m_endtime = 0.0f;
+	}
+
+	return instance->GetRumbleStatus();
```
youle31: About that I completed my last comment last time to handle the case we have a .blend with vibration actuator and "Actuator" sensor with a pad that doesn't support vibrations -> I think the .blend should have exactly the same behaviour if the joystick support vibrations or not. Returning only instance->GetRumbleStatus() doesn't handle this case (for joypads that doesn't support vibrations running this .blend with vibration actuator...)

youle31: I don't know why m_endtime switch from 0.0f to pil time + m_duration each frame in my test file...

youle31: Ah I understood. each actuator, play and stop has its own m_endtime. When we activate RumbleStop Actuator we have to say to the RumblePlay actuator to reset its own m_endtime to 0...

youle31: SoI propose this fix to have a variable endtime shared by the 2 actuators: http://pasteall.org/89653/diff (I'm not sure about that but this works for 2 actuators...) Else the other solution would be to have an unique actuator and implement cont.deactivate with rumblestop...

lordloki: Hi,

the m_endtime should not be the same in different actuators. It is an internal count for playing and it fulfil its task.

Regarding to the return behaviour if the joystick supports or not vibration I think that it should be different as it follows the reality. 

BUT you are pointing well out a weak point. To mitigate this weak i have included a new "hasVibration" attribute to can check if the joystick supports or not vibration. 

Thanks.

youle31:
Yes you're right. I'll test tomorrow with my pads :)

youle31:
Ok I tested. So with the same .blend with the same logic bricks: http://pasteall.org/blend/index.php?id=44101 the cube will move with a pad that supports vibrations but won't move with a pad that support vibrations. For me it's not normal because the same game can be runned by several people and everyone has not vibration support.

But as it's a minor bug that concerns a very specific case and as with your attribute hasVibration allows to write a script to workaround this unexpected behaviour (in my opinion), It is good to me. I think it can be landed.

panzergame:
Failed compilation: http://download.upbge.org/download/buildbot/logs/ge_vibration_leftright.txt

lordloki:
I will update to current master

source/gameengine/Device/DEV_JoystickVibration.cpp
```diff
@@ -0,0 +1,190 @@
...
+	if (run_by_effect) {
+		bool new_effect = true;
+
+		if (m_private->m_hapticeffect_status == JOYHAPTIC_UPDATING_EFFECT) {
+			if (SDL_HapticUpdateEffect(m_private->m_haptic, m_private->m_hapticeffect_id, &m_private->m_hapticeffect) == 0) {
+				m_private->m_hapticeffect_status = JOYHAPTIC_PLAYING_EFFECT;
+				new_effect = false;
+			}
+			else {
+				SDL_HapticDestroyEffect(m_private->m_haptic, m_private->m_hapticeffect_id);
```
panzergame: This statement is for an error ? I yes, you can add an error message with `CM_Error`

source/gameengine/Device/DEV_JoystickVibration.cpp
```diff
@@ -0,0 +1,190 @@
...
+ *  \ingroup device
+ */
+
+#include "DEV_Joystick.h"
+#include "DEV_JoystickPrivate.h"
+#include "DEV_JoystickDefines.h"
+
+#include <memory> // We have to include that on Windows to make memset available
+
+bool DEV_Joystick::RumblePlay(float strength_left, float strength_right, unsigned int duration)
```
panzergame: I'm trying to simplify the statements, i will sent you a diff.

panzergame:
@lordloki : Please fix the branch, it is impossible to rebase upon master.

lordloki:
I will launch another uptodate branch with the changes. I need 3-4 days (sorry i'm busy with work)

source/gameengine/Converter/KX_ConvertActuators.cpp
```diff
@@ -869,12 +869,27 @@ void BL_ConvertActuators(const char* maggiename,
...
+				}
+				case ACT_VIBRATION_STOP:
+				{
+					mode = SCA_VibrationActuator::KX_ACT_VIBRATION_STOP;
+					break;
+				}
+			}
 
 			int joyindex = vib_act->joyindex;
 			float strength = vib_act->strength;
```
lordloki: done

source/gameengine/Device/DEV_JoystickPrivate.h
```diff
@@ -45,12 +45,13 @@ class DEV_Joystick::PrivateData
 	SDL_JoystickID m_instance_id;
 	SDL_Haptic *m_haptic;
 	SDL_HapticEffect m_hapticeffect;
+	int m_hapticeffect_id, m_hapticeffect_status;
```
lordloki: I will do a style clean up over DEV_Joysticks* when commited

lordloki:
I have launched a new pull request updated. I close this one

PR #272, opened 2016-10-26
Fix crash when calling setTexture in KX_2DFilter

youle31:
I just added a check in GetUniformLocation to see if there is an error and
return if yes because it caused a crash calling KX_2DFilter::setTexture

test file: http://pasteall.org/blend/index.php?id=44069

If there is an error in the fragment shader it seems m_shader is removed so we can't access it anymore. So we can fix by avoid removing m_shader if there is an error in the shader or do what I choose to do: the simplest way (but not necessarly the better) -> just add a check in the function that caused the crash.

### DIFF
--- a/source/gameengine/Ketsji/KX_2DFilter.cpp
+++ b/source/gameengine/Ketsji/KX_2DFilter.cpp
@@ -90,6 +90,9 @@ KX_PYMETHODDEF_DOC(KX_2DFilter, setTexture, "setTexture(index, bindCode, sampler
 	}
 
 	if (samplerName) {
+		if (GetError()) {
+			Py_RETURN_NONE;
+		}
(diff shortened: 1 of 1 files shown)


source/gameengine/Rasterizer/RAS_Shader.cpp
```diff
@@ -481,6 +481,9 @@ void RAS_Shader::BindAttribute(const char *attr, int loc)
 
 int RAS_Shader::GetUniformLocation(const char *name, bool debug)
 {
+	if (GetError()) {
+		return -1;
+	}
 	BLI_assert(m_shader != NULL);
```
panzergame: Hmm, we have to choose between an assertion or a -1. @youle: What about moving the check in the setTexture function only if the problem appears only in this function.

PR #273, opened 2016-10-26
UPBGE: Cull shadows when light frustum doesn't interstect with camera frustum

youle31:
Works for sun. Needs optimization

test file: http://pasteall.org/blend/index.php?id=44074
perf test file: http://pasteall.org/blend/index.php?id=44075

### DIFF
--- a/source/gameengine/Ketsji/KX_KetsjiEngine.cpp
+++ b/source/gameengine/Ketsji/KX_KetsjiEngine.cpp
@@ -854,22 +854,13 @@ void KX_KetsjiEngine::RenderShadowBuffers(KX_Scene *scene)
 			 * is outside camera frustum, we should see his shadows. What we can do is check if light frustum
 			 * intersects with camera frustum. If this is not the case, we don't render shadows.
 			 */
-			if (raslight->m_type == RAS_ILightObject::LIGHT_SUN) {
-				MT_Vector3 lightfrustum[8];
-				raslight->GetSunShadowBox(lightfrustum);
-				int insideFrustum = scene->GetActiveCamera()->BoxInsideFrustum(lightfrustum); // 0 = inside, 1 = intersect, 2 = outside
-				if (insideFrustum == 2) {
-					return;
-				}
+			MT_Vector3 lightfrustum[8];
+			raslight->GetShadowBox(lightfrustum);
+			int insideFrustum = scene->GetActiveCamera()->BoxInsideFrustum(lightfrustum); // 0 = inside, 1 = intersect, 2 = outside
+			if (insideFrustum == 2) {
+				return;
 			}
-			else if (raslight->m_type == RAS_ILightObject::LIGHT_SPOT) {
-				MT_Vector3 lightfrustum[5];
-				raslight->GetSpotShadowBox(lightfrustum);
-				int insideFrustum = scene->GetActiveCamera()->PyramidInsideFrustum(lightfrustum); // 0 = inside, 1 = intersect, 2 = outside
-				if (insideFrustum == 2) {
-					return;
-				}
-			}
+
 			/* make temporary camera */
 			RAS_CameraData camdata = RAS_CameraData();
 			KX_Camera *cam = new KX_Camera(scene, scene->m_callbacks, camdata, true, true);
--- a/source/gameengine/Ketsji/KX_KetsjiEngine.cpp
+++ b/source/gameengine/Ketsji/KX_KetsjiEngine.cpp
@@ -854,13 +854,11 @@ void KX_KetsjiEngine::RenderShadowBuffers(KX_Scene *scene)
 			 * is outside camera frustum, we should see his shadows. What we can do is check if light frustum
 			 * intersects with camera frustum. If this is not the case, we don't render shadows.
 			 */
-			if (raslight->m_type == RAS_ILightObject::LIGHT_SUN) {
-				MT_Vector3 lightfrustum[8];
-				raslight->GetShadowBox(lightfrustum);
-				int insideFrustum = KX_GetActiveScene()->GetActiveCamera()->BoxInsideFrustum(lightfrustum); // 0 = inside, 1 = intersect, 2 = outside
-				if (insideFrustum == 2) {
-					return;
-				}
+			MT_Vector3 lightfrustum[8];
+			raslight->GetShadowBox(lightfrustum);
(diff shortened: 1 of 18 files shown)


source/gameengine/Ketsji/KX_KetsjiEngine.cpp
```diff
@@ -841,6 +841,17 @@ void KX_KetsjiEngine::RenderShadowBuffers(KX_Scene *scene)
...
 			raslight->HasShadowBuffer() && raslight->NeedShadowUpdate())
 		{
+			/* There was no culling for shadows other than if the object was outside light frustum.
+			 * We can't do a complete culling according to camera frustum because even if an object
+			 * is outside camera frustum, we should see his shadows. What we can do is check if light frustum
+			 * intersects with camera frustum. If this is not the case, we don't render shadows.
+			 */
+			MT_Vector3 lightfrustum[8];
+			raslight->GetShadowBox(lightfrustum);
+			int insideFrustum = KX_GetActiveScene()->GetActiveCamera()->BoxInsideFrustum(lightfrustum); // 0 = inside, 1 = intersect, 2 = outside
```
panzergame: KX_GetActiveScene is not needed here, just replace by `scene`.

youle31:
@lordloki : I can't see your comment anymore. I guess when we push --force the comments on old commits disappear because they are attached to the old commit. You can directly comment on pull request "conversation" or in "files changed" I think, like that the comments don't disappear.

About pyramid, I don't know how to write a function "PyramidInsideFrustum" but if you have an idea this would be better indeed.

youle31:
Done for pyramid. I didn't test so I'll test. 

And panzergame told me something that I didn't understood:  "shadow are rendered in a limited area if we talk about depth, but the shadow is projected in a unlimited depth. So if the camera see an object aligned with the Z axis of the shadow box then the shadow worth update." so we need to discuss it.

panzergame:
http://pasteall.org/pic/index.php?id=108185

The camera will see the shadow of the cube without see the shadow box.

youle31:
The Shadow can be outside the light frustum? How can the Shadow be computed if it Is out light frustum?

youle31:
Ok I understood. So do you think if we take a big value for clipend this would be ok?

panzergame:
There's an issue with multiple cameras: currently the shadow are updated only one time at the begining of the render stage, but it will use only one camera for the culling and not all the cameras.

To solve this i will work on updating the shadow for each cameras but with a flag to avoid redundant update.

youle31:
ah yes... right

panzergame:
I tested the patch this morning and changed some things. I'm very surprised to not see any slowdown when the culling is not needed, when the camera perfectly saw the shadow.

About the pyramide shapes what about add an argument "count" to KX_Camera::BoxInsideFrustum to design the number of point and then in this function iterate of the same number of points.

youle31:
I changed the pyramid and restored a box because even if it Is a bit less accurate we have only 8 points vs 8 plus 5 points if we have a pyramid plus an infinité box

source/gameengine/Rasterizer/RAS_OpenGLRasterizer/RAS_OpenGLLight.cpp
```diff
@@ -200,6 +198,57 @@ MT_Matrix4x4 RAS_OpenGLLight::GetShadowMatrix()
 	return mat;
 }
 
+void RAS_OpenGLLight::GetShadowBox(MT_Vector3 *box)
+{
+	GPULamp *lamp;
+
+	if ((lamp = GetGPULamp())) {
```
panzergame: i perfer avoid assignment in condition, instead do:
```
GPULamp *lamp = GetGPULamp();

if (lamp) {
```

source/gameengine/Rasterizer/RAS_OpenGLRasterizer/RAS_OpenGLLight.cpp
```diff
@@ -200,6 +198,57 @@ MT_Matrix4x4 RAS_OpenGLLight::GetShadowMatrix()
...
+		if (light->GetLightData()->m_type == RAS_ILightObject::LIGHT_SUN) {
+
+			box[0][0] = box[1][0] = box[2][0] = box[3][0] = -GPU_lamp_frustum_size(lamp);
+			box[4][0] = box[5][0] = box[6][0] = box[7][0] = +GPU_lamp_frustum_size(lamp);
+			box[0][1] = box[1][1] = box[4][1] = box[5][1] = -GPU_lamp_frustum_size(lamp);
+			box[2][1] = box[3][1] = box[6][1] = box[7][1] = +GPU_lamp_frustum_size(lamp);
+			box[0][2] = box[3][2] = box[4][2] = box[7][2] = -FLT_MAX;
+			box[1][2] = box[2][2] = box[5][2] = box[6][2] = -GPU_lamp_clipstart(lamp);
+
+			MT_Matrix3x3 orientation = light->NodeGetWorldOrientation();
```
panzergame: You could share the matrix/transform creation and apply to corner for sun light and spot light.

source/gameengine/Ketsji/KX_KetsjiEngine.cpp
```diff
@@ -846,40 +862,56 @@ void KX_KetsjiEngine::RenderShadowBuffers(KX_Scene *scene)
 
 		raslight->Update();
 
-		if (light->GetVisible() && m_rasterizer->GetDrawingMode() == RAS_IRasterizer::RAS_TEXTURED &&
```
panzergame: I will change this to use return statement instead of indentation monster.

youle31: I didn't understood this. I let you fix if you don't mind. Thanks

source/gameengine/Ketsji/KX_KetsjiEngine.cpp
```diff
@@ -898,10 +925,6 @@ void KX_KetsjiEngine::RenderFrame(KX_Scene *scene, KX_Camera *cam, unsigned shor
 
 	KX_SetActiveScene(scene);
 
-#ifdef WITH_PYTHON
```
youle31: This has to be restored.... I don't know how... When I rebase this leads to weird things. This is why I prefer to make new branches hehe

youle31:
Arf it doesn't work as expected. I'll try to draw frustums to see what is wrong... :/

PR #274, opened 2016-10-29
UPBGE: Implement python callbacks for custom shaders.

panzergame:
Previously the user wasn't able to set uniform values per object
using a custom shader and not globally for all the shader.

To solve this lack the easiest way is to give the user the possibility
to use a python callback. This callback is called just before
rendering a mesh with the custom shader.

The callbacks executed receive as arguments the shader to alterate
and the game object to have an identifier.

The callbacks list is accessed under the name of BL_Shader.objectCallback.
###### 

I'm really not sure about the python attribute name, @DCubix, @youle31, @lordloki : Do you have a suggestion ?


panzergame:
Test file:
http://pasteall.org/blend/index.php?id=44110

The both cube use the same custom shader but a different color.

youle31:
I don't understand how it works but it works for me so it's ok. I need to read doc about callbacks...

PR #282, opened 2016-11-02
UPBGE: Add GPU texture coordinates scrolling and transform

youle31:
I used the shader already existing in gpu_shader_material.glsl and
gpu_material.c (mtex_mapping_ofs). It doesn't take into account uv
transformation with a matrix. I don't know if this is possible but I made
as simple as possible. There is already an UI for UV offset in texture
panel (mapping) I was not aware of that, neither that mtex contained
already an attribute float ofs[3].

So what I did is just changing GPU_uniform for GPU_select_uniform and add
an API with Mathutils Vector to set the offset:

own.meshes[0].materials[0].textures[0].uvOffset.x += 0.01

test file/ perf test file: http://pasteall.org/blend/index.php?id=44133


source/blender/gpu/intern/gpu_material.c
```diff
@@ -1375,9 +1375,11 @@ static void do_material_tex(GPUShadeInput *shi)
 				mtex->ofs[1] + 0.5f - 0.5f * mtex->size[1],
 				0.0f
 			};
+			mtex->ofs[0] = ofs[0];
```
youle31: I'm not 100% sure about that. Maybe it would be good to see if mtex->ofs is used elsewhere.

youle31: I made changes to ensure there will be no problem. See comments in gpu_material.c

DCubix:
I think it's kind of weird to have that in the texture, but it's nice :+1: 
We should have uv rotation and scaling too with a simple MT_Matrix4x4

youle31:
Thanks @DCubix :) This is in textures because it concerns texture coordinates and the code is build like that for the moment (all MTex attributes are set in BL_Texture).

About transformation with a matrix4x4, it's feasible (I tested), but I'd do that in another shader called optionaly (contrarly to the shader I used that is always used for all textures) because we have to pass a matrix4x4 as uniform and it is slow... And maybe it would need an UI... I don't really know how to design it for now.

youle31:
@DCubix : I go to sleep now but if you have ideas for uvTransform, you're welcome to commit here or do another patch if you want. mat_rot = mathutils.Matrix.Rotation(radians(45.0), 4, 'X') :P

youle31:
test file for uvTransform: http://pasteall.org/blend/index.php?id=44145

Note: I added an UI in game textures panel->mapping to enable uvTransform to avoid to pass an optional 4x4 matrix to the material shader

youle31:
I have to change center of uv coordinates for transformation so I remove wait reviewer flag

youle31:
👍 I agree with last commit. Makes things more uniform (offset + size + rotation) (not tested yet)

youle31:
I opened a new branch with a slightly different design.

comments: - I think we can't change the size behaviour.
- I don't know if the offset must take into account rotation so I made 2 commits -> in the first, offset doesn't take into account rotation, and in the second, offset takes into account rotation.

It's in ge_uvscroll2

I used this file to test: http://pasteall.org/blend/index.php?id=44226 (but I didn't made enough tests for now)

youle31:
Fot this patch I have a last comment:

I think as for offset, the size shouldn't be influenced by rotation. When in the UI set the size on X axis, I think we expect the size to be set in "global" UV coordinates, not like that: http://pasteall.org/pic/index.php?id=108772

More than when we do S+X in the UV image editor

youle31:
@panzergame : We haven't the same behaviout in your patch and in what I proposed.

outtexco = (texco - vec3(0.5)) * size;
outtexco = ((outtexco + ofs) * mat + vec3(0.5));

In your patch, with outtexco = (texco - vec3(0.5)) * mat * size + vec3(0.5) + ofs; the size and the offste are influenced by rotation... I think when the user use size on X axis even if there is a rotation, he excpects that the scale happens horizontally. Same for the offset... IMHO

lordloki:
Horizontally, in local coordinates or global coordinates?

youle31:
@lordloki: horizontally according to your monitor :P

lordloki:
Hehehe, ;) i was refering that if x axis points 0 degrees, then you turn 90 degrees and after you scale in x, which scale will you see, in vertical or in horizontal (local or global coordinates, respectively)

youle31:
@lordloki: For me this behaviour is not intuitive: https://www.youtube.com/watch?v=61Ci4IWMuJ8&feature=youtu.be

This: https://youtu.be/Y8y5VVPF09E is more intuitive

lordloki:
@youle31 yes, I'm with you. It seems that in @panzergame's algorithm the scale is done in global axis and translate in local axis.

panzergame:
We have three formulas:

**translation/rotation/scale**:
```outtexco = ((texco - vec3(0.5)) * size + ofs * 0.5) * mat + vec3(0.5);```
Scale depends on the rotation and offset independent on the rotation, keeps UV orthogonal, rotate from offset center.

**translation/scale/rotation**:
```outtexco = ((texco - vec3(0.5)) * size + ofs) * mat + vec3(0.5)```
Scale and offset independent from rotation, doesn't keep UV orthogonal, rotate from offset center.

**rotation/scale/translation**:
```outtexco = (texco - vec3(0.5)) * mat * size + vec3(0.5) + ofs```
Scale and offset is dependent of rotation, keeps UV orthogonal, rotate from global center (0, 0).

The rotation/scale/translation is excluded. We have to choice between the two first.

**Comment by youle31** [comment]
I prefer my solution or that:

outtexco = texco - vec3(0.5);
outtexco = outtexco * mat;
outtexco += ofs;
outtexco += vec3(0.5);
outtexco *= size;

to have scale and offset depend on rotation... I don't know.

**Comment by youle31** [comment]
Sorry I didn't see this was your 3rd option. I just tested the 2 first because you said the third was exculded... Why do you exculde the third?

**Comment by panzergame** [comment]
The third is excluded because the offset is dependent of the rotation.

**Comment by panzergame** [comment]
@lordloki : Are you agree to use the formula ? :
```
outtexco = (texco - vec3(0.5)) * mat * size + vec3(0.5) + ofs;
```
Indeed this formula is the only which doesn't deforme the texture but only rotating it.

youle31:
Sorry, I read again the thread and it seems I was very confused, but in fact the last formula proposed by panzergame sounds good to me.

PR #286, opened 2016-11-04
UPBGE: Allow accessing the vehicle speed from python

DCubix:
From the commit message:
> This commit exposes to Python the vehicle speed in meters per
second. You can convert the value to Km/h by multiplying by 3.6
 and to Mph by dividing by 0.44704.
	Given the vehicleID, you can access the speed by simply calling
"vehicleID.speed".

Example: http://pasteall.org/blend/index.php?id=44151

### DIFF
--- a/extern/bullet2/src/BulletDynamics/Vehicle/btRaycastVehicle.h
+++ b/extern/bullet2/src/BulletDynamics/Vehicle/btRaycastVehicle.h
@@ -61,7 +61,7 @@ private:
 	btVehicleRaycaster*	m_vehicleRaycaster;
 	btScalar		m_pitchControl;
 	btScalar	m_steeringValue; 
-	btScalar m_currentVehicleSpeedKmHour;
+	btScalar m_currentVehicleSpeedMetersPerSec;
 
 	btRigidBody* m_chassisBody;
 
@@ -180,9 +180,9 @@ public:
 	}
 
 	///Velocity of vehicle (positive if velocity vector has same direction as foward vector)
-	btScalar	getCurrentSpeedKmHour() const
+	btScalar	getCurrentSpeedMetersSec() const
 	{
-		return m_currentVehicleSpeedKmHour;
+		return m_currentVehicleSpeedMetersPerSec;
 	}
 
 	virtual void	setCoordinateSystem(int rightIndex,int upIndex,int forwardIndex)
--- a/source/gameengine/Ketsji/KX_VehicleWrapper.cpp
+++ b/source/gameengine/Ketsji/KX_VehicleWrapper.cpp
@@ -27,6 +27,7 @@
 #include "KX_VehicleWrapper.h"
 #include "PHY_IPhysicsEnvironment.h"
 #include "PHY_IVehicle.h"
+#include "CcdPhysicsEnvironment.h"
 #include "KX_PyMath.h"
 #include "KX_GameObject.h"
 #include "KX_MotionState.h"
@@ -377,6 +378,7 @@ PyMethodDef KX_VehicleWrapper::Methods[] = {
 
 PyAttributeDef KX_VehicleWrapper::Attributes[] = {
 	KX_PYATTRIBUTE_RW_FUNCTION("rayMask", KX_VehicleWrapper, pyattr_get_ray_mask, pyattr_set_ray_mask),
+	KX_PYATTRIBUTE_RO_FUNCTION("speed", KX_VehicleWrapper, pyattr_get_khspeed),
 	{ NULL }	//Sentinel
 };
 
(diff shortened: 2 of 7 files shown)


extern/bullet2/src/BulletDynamics/Vehicle/btRaycastVehicle.h
```diff
@@ -61,7 +61,7 @@ class btRaycastVehicle : public btActionInterface
 	btVehicleRaycaster*	m_vehicleRaycaster;
 	btScalar		m_pitchControl;
 	btScalar	m_steeringValue; 
-	btScalar m_currentVehicleSpeedKmHour;
+	btScalar m_currentVehicleSpeedMetersPerSec;
```
youle31: Hi! I don't know if this is a good idea to modify bullet sources... I'd use the existing m_currentVehicleSpeedKmHour, getCurrentSpeedKmHour() const

youle31:
Else the comment I left, it sounds good to me. Why I wouldn't change bullet sources? Because it's an external dependancie and it will create a conflict if we switch to bullet 2.85 or so...

extern/bullet2/src/BulletDynamics/Vehicle/btRaycastVehicle.h
DCubix: I don't like it. It should not be KmH. And that value is calculated VERY wrong.

youle31:
In fact, as we already have linearVelocity and that we can get linearVelocity.y for example, and as it works for all gameobjects, not only vehicles, I'm wondering if this patch is really needed (sorry to be argue against your patch... I try to be objective). I'm not opposed to the patch, I think it can be added but I'd change the design to avoid to modify bullet sources.

And I forgot, maybe BU/second would be more exact.

When you do:

vec3(0.5, 0.6, 0.7).dot(0.0, 1.0, 0.0) it returns 0.6 no (I didn't checked sorry)? It's like if we say gameobj.linearVelocity.y? I'll check.

extern/bullet2/src/BulletDynamics/Vehicle/btRaycastVehicle.h
panzergame: @DCubix : You should add in the bullet2 directory a patch file containing these changes: it's for futur upgrade to reapply patches.

source/gameengine/Ketsji/KX_VehicleWrapper.cpp
```diff
@@ -407,4 +409,11 @@ int KX_VehicleWrapper::pyattr_set_ray_mask(void *self, const struct KX_PYATTRIBU
 	return PY_SET_ATTR_SUCCESS;
 }
 
+PyObject *KX_VehicleWrapper::pyattr_get_speed(void *self, const struct KX_PYATTRIBUTE_DEF *attrdef)
+{
+	KX_VehicleWrapper *wrapper = static_cast<KX_VehicleWrapper*>(self);
+	MT_Scalar speed = std::abs(wrapper->m_vehicle->GetCurrentSpeed()); // No negative speeds
```
panzergame: Negative speed is possible in this level ?

source/gameengine/Physics/Bullet/CcdPhysicsEnvironment.cpp
```diff
@@ -376,6 +376,10 @@ class WrapperVehicle : public PHY_IVehicle
 	{
 		return m_raycaster->GetRayCastMask();
 	}
+
+	virtual MT_Scalar GetCurrentSpeed() const {
```
panzergame: Brace on new line.

youle31:
http://pasteall.org/blend/index.php?id=44168 (compare linearVelocity and speed)

source/gameengine/Ketsji/KX_VehicleWrapper.cpp
DCubix: Yes why not? It's speed. It would be very weird if a speedometer showed -50 Km/h :stuck_out_tongue:

panzergame:
i have some doubt on the way we compute the speed: if the vehicle is falling but the wheel not turning then the speed increase contrary to the speed gives a real car. To solve this issue we have to simply get the speed from the wheel rotation it's simple but unstable for some reasons: Each wheel are giving a different speed...

lordloki:
About modifying the bullet code. In this case I think it is not necessary as we can obtain the info need easily, i.e:

virtual MT_Scalar GetCurrentSpeed() const {
......return (m_vehicle->getCurrentSpeedKmHour() / 3.6f);
}

We have only to modify external libraries when we can obtain the info needed.

panzergame:
@lordloki: The vehicle speed computed by bullet is wrong because it only does the lenght of the velocity in the right unit. It doesn't care of the wheel direction.

panzergame:
As exposed, there's not an unique vehicle speed: it can be the vehicle body speed, the vehicle body speed in front wheel axis, the speed from wheel rotation ect…

So it's better to let the user compute the speed he wants.

PR #287, opened 2016-11-04
UPBGE: Makes distance and angle python attributes for Radar Sensor writables

lordloki:
This patch adds the posibility to modify the radar cone's distance and angle attributes during the runtime through BGE python.

Credits for this patch goes to @Kseniya Halezova (blenderprowlee) too.

An example to can check the feature http://pasteall.org/blend/index.php?id=44156

### DIFF
--- a/extern/bullet2/src/BulletCollision/CollisionShapes/btConeShape.h
+++ b/extern/bullet2/src/BulletCollision/CollisionShapes/btConeShape.h
@@ -43,6 +43,15 @@ public:
 	btScalar getRadius() const { return m_radius;}
 	btScalar getHeight() const { return m_height;}
 
+	virtual void setRadius(btScalar radius)
+	{
+		m_radius = radius;
+	}
+
+	virtual void setHeight(btScalar height)
+	{
+		m_height = height;
+	}
 
 	virtual void	calculateLocalInertia(btScalar mass,btVector3& inertia) const
 	{
--- a/source/gameengine/Physics/Bullet/CcdPhysicsController.h
+++ b/source/gameengine/Physics/Bullet/CcdPhysicsController.h
@@ -768,6 +768,16 @@ public:
 		m_cci.m_radius = margin;
 	}
 
+	// Cone Shape (radius & height)
+	virtual void SetConeShape(float radius, float height)
+	{
+		if (m_collisionShape && (m_collisionShape->getShapeType() == CONE_SHAPE_PROXYTYPE)) {
+			btConeShape *coneShape = static_cast<btConeShape *>(m_collisionShape);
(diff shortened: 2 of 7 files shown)


extern/bullet2/src/BulletCollision/CollisionShapes/btConeShape.h
```diff
@@ -43,6 +43,15 @@ ATTRIBUTE_ALIGNED16(class) btConeShape : public btConvexInternalShape
 	btScalar getRadius() const { return m_radius;}
 	btScalar getHeight() const { return m_height;}
 
+	virtual void setRadius(btScalar radius)
```
youle31: Hi! I made the same comment for TwisterGE's patch about vehicle speed: I don't know if this is a good idea to change bullet sources as it's an external lib and that this will create conflict when we'll switch to another bullet version.
Maybe there is another way to design this patch without modifying bullet sources? @panzergame: What do you think?

lordloki: that change was upstream to bullet sources already and it is included in new bullet 2.85

youle31: ah ok :) So no problem

source/gameengine/Physics/Bullet/CcdPhysicsController.h
```diff
@@ -768,6 +768,16 @@ class CcdPhysicsController : public PHY_IPhysicsController
 		m_cci.m_radius = margin;
 	}
 
+	// Cone Shape (radius & height)
+	virtual void SetConeShape(float radius, float height)
+	{
+		if (m_collisionShape && (m_collisionShape->getShapeType() == CONE_SHAPE_PROXYTYPE)) {
+			btConeShape *coneShape = static_cast<btConeShape *>(m_collisionShape);
+			coneShape->setRadius(radius);
```
panzergame: Bullet support well the shape changes ? No AABB to update or collision pair to remove ?

lordloki: The unique recomendation that Erwin Coumans said me was: "Just remember to remove the rigid body from the world before making those changes, and then re-insert it."
I will modify the patch to follow it

panzergame: In this case the best is to reinstance the pĥysics shape.

panzergame:
Failed compilation: http://download.upbge.org/download/buildbot/logs/ge_radar.txt

lordloki:
I will update to current master

lordloki:
Close this branch as I have created one new updated to master

PR #293, opened 2016-11-09
UPBGE: Optimize real-time mesh modification

DCubix:
By using glBufferSubData and GL_DYNAMIC_DRAW, I was able to optimize it (KX_MeshProxy) in almost 80%.
Previously, the mesh was reallocated in the GPU all the time, which is a terribly slow process. Now only the mesh data is updated instead of reallocating memory, making if a lot faster.

Demo: http://pasteall.org/blend/index.php?id=44237
(The plane has +/- 3000 vertices, the program modifies the vertex positions and recalculated the normals of the entire mesh, all with 60FPS)

wow

### DIFF
--- a/source/gameengine/Rasterizer/RAS_OpenGLRasterizer/RAS_StorageVBO.cpp
+++ b/source/gameengine/Rasterizer/RAS_OpenGLRasterizer/RAS_StorageVBO.cpp
@@ -120,7 +120,7 @@ void VBO::AllocData()
 	glBindBufferARB(GL_ARRAY_BUFFER_ARB, 0);
 
 	glBindBufferARB(GL_ELEMENT_ARRAY_BUFFER_ARB, m_ibo);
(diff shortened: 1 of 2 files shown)


youle31:
Hi!

Perf tests on my computer:

your demo file:

- Master: 165 fps
- ge_vbo_dynamic: 173 fps

In water demo file 0.99 I have ~the same fps and in Sintel desert maybe a slight perf increase.

So it's really cool :) Congrats and thanks! If it increase perfs of  80 % on your computer this is awesome.

Akira1San:
The file give 107fps, upbge 011: 111fps.
Made a test with the file: desert scene from the blendswap with little modifications.
http://pasteall.org/pic/index.php?id=108634
pic 1: bf bge 278a,  vbo dynamic, upbge 011

youle31:
It's ok for me.

PR #305, opened 2016-11-23
UPBGE: Add an option to Get/Set softbody linear/angluar stiffness in realtime

youle31:
test file: http://pasteall.org/blend/index.php?id=44419

I make the pull request only to have a better visibility of the code but this is just a prototype.

### DIFF
--- a/doc/python_api/rst/bge_types/bge.types.KX_GameObject.rst
+++ b/doc/python_api/rst/bge_types/bge.types.KX_GameObject.rst
@@ -106,6 +106,26 @@ base class --- :class:`SCA_IObject`
 
          The object must have a physics controller for the angular damping to be applied, otherwise the value will be returned as 0.0.
 
+   .. attribute:: linearStiffness
+
+      The object's linear stiffness (only for softbodies)
+
+      :type: float between 0 and 1 inclusive.
+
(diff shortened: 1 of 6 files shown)


youle31:
@panzergame: sdfgeoff asked if it was possible in this file: http://pasteall.org/blend/index.php?id=44434 to make the object stiffness low then hard like if you pump up the sphere (gonflait la balle/sphere). I think it's impossible because you can't revert simulation that has already happened... I mean -> if you increase stiffness, only the future deformations will be harder but this won't "revert" softbody simulation.

Unless you have an idea?

youle31:
I tried volume stiffness (kVST) but it doesn't work. There are tons of attributes to test

youle31:
I added CM_PythonAttributError for the stiffness: here is the outpu when I make an error:

Error: body.py(10), KX_GameObject.linearStiffness, expected a float
Error: Python.002(Icosphere), Python script error
  File "C:\Users\yul\Desktop\softBodyPinStiffness2.blend\body.py", line 10, in hard
TypeError: a float is required

Is this normal?

youle31:
Not really interesting. I close this for now. But the code will still be available in closed feature requests

PR #309, opened 2016-11-26
UPBGE: Vibration Actuator improvement

lordloki:
Now the vibration actuator has 2 methods (start and stop). Start method
starts the vibration and also it can update the parameters of the
vibration on fly. Stop method stops the vibration.

Now there is 2 strenght parameters the "left" one sets the strenght of
the Low frequency joystick's motor and the "right" one sets the strenght
of the High frequency joystick's motor.

Additionaly, 2 new bool attributes have been included (isVibrating and
hasVibration) along the new python methods.

### DIFF
--- a/build_files/cmake/macros.cmake
+++ b/build_files/cmake/macros.cmake
@@ -637,8 +637,8 @@ function(SETUP_BLENDER_SORTED_LIBS)
 		extern_curve_fit_nd
 		ge_logic_ketsji
 		extern_recastnavigation
-		ge_device
 		ge_logic
+		ge_device
 		ge_oglrasterizer
 		ge_rasterizer
 		ge_common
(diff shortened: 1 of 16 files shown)


panzergame:
@lordloki : pushed some review changes in ge_vibration_leftright. I also noticed an issue:
It appears sometimes that the controller failed run any rumble at game start, when i debug effect creation i see that it is always creating an new effect even i the duration is not over. `(m_private->m_hapticEffectStatus == JOYHAPTIC_UPDATING_EFFECT)` seems to be false even at game start.

lordloki:
@panzergame: I will debug it to see where is the issue.

lordloki:
@panzergame: A fix was pushed. Could you check it again?

panzergame:
@lordloki : Thanks, the bug seems fixed now.

lordloki:
OK, then I close this pull request and open the pull request from ge_vibration_leftright branch

PR #310, opened 2016-11-27
UPBGE: Makes distance and angle python attributes for Radar Sensor

lordloki:
writables

This patch adds the posibility to modify the radar cone's distance and
angle attributes during the runtime through BGE python.

Credits for this patch goes to @Kseniya Halezova (blenderprowlee) too.

An example to can check the feature http://pasteall.org/blend/index.php?
id=44156

### DIFF
--- a/extern/bullet2/patches/coneshape_setheight.patch
+++ b/extern/bullet2/patches/coneshape_setheight.patch
@@ -1,0 +1,26 @@
+diff --git a/extern/bullet2/patches/coneshape_setheight.patch b/extern/bullet2/patches/coneshape_setheight.patch
+new file mode 100644
+--- /dev/null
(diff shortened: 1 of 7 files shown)


ondredeniso:
Thanks you. Looks like useful feature.

youle31:
Hi, for me this is good.

Just 2 comments/questions:
1) For the bullet patch, I think the patch won't apply on my PC for example because there is no new line at end of file.
2) I still use PyErr_SetString in my own patches but I was wondering if this had changed. We have to use something else now? CM_Error or something like that? CM means colored message? I'll check.

lordloki:
@youle, I will take a look. thanks :-)

youle31:
Sorry I finally tested and the bullet patch applies correctly.

PR #324, opened 2016-12-07
UPBGE: Securise std::vector erase iteration.

youle31:
Using a fixed end iterator or a incrementing the iterator after
delete its previous is unsafe for std::vector. Because erasing
an element can realloc the whole container and then invaldate
the existing iterators.

see: 2c669cc7adfc71de78470dbc17a969e542fb2807

for example file see: https://github.com/UPBGE/blender/issues/321 (Every man's sky-> crash at start in MSVC debug mode)


panzergame:
Nice catch :) commit it ! :)

youle31:
commited

PR #325, opened 2016-12-07
UPBGE: Disable MSVC 2015 warning 4250

youle31:
KX_2DFilter: SetProg Inherits from BL_Shader via dominance

### DIFF
--- a/source/gameengine/Ketsji/KX_2DFilter.h
+++ b/source/gameengine/Ketsji/KX_2DFilter.h
@@ -31,6 +31,10 @@
 #include "BL_Shader.h"
 
 #ifdef _MSC_VER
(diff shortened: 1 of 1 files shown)


source/gameengine/Ketsji/KX_2DFilter.h
```diff
@@ -30,6 +30,10 @@
 #include "RAS_2DFilter.h"
 #include "BL_Shader.h"
 
+#ifdef _MSC_VER
+#  pragma warning(disable:4250)
```
panzergame: Could you leave a little comment as we do for ge_batch ?

youle31:
commited

PR #326, opened 2016-12-07
UPBGE: replace setIdentity each frame with const matrix

youle31:
In RAS_2DFilter::Start, each frame a MT_Matrix4x4 was created and set to
identity. I replaced it with static const mat already set to identity.


panzergame:
An equivalent is `MT_Matrix4x4 mat = MT_Matrix4x4::Identity();`, else can you find similar case in the code ?

panzergame:
I'm wrong sorry

youle31:
No, I didn't searched similar case in the code. I just did the modif as I thought that setting matrix each frame was not needed

youle31:
I'll maybe do another patch to replace static matrix with a member matrix

PR #329, opened 2016-12-08
UPBGE: Fix format for HDR_NONE RGB RGBA8

youle31:
Before
https://github.com/UPBGE/blender/commit/732d176d09b6c2497232bd3ecfced9042b66deff
the format used to make bgl Buffer when you called texture.source.image
was RGB. The default format has been replaced with GL_RGBA8. This caused
error mentionned here:
https://github.com/UPBGE/blender/issues/311
To avoid to break backward compatibility, I restore the old default format
for render to texture. Indeed old files that use texture.source.image used
a 3 components R, G, B buffer. I guess changing that can cause issues to
have now 4 values.


panzergame:
Your idea was right, i confused image format and image internal format, it's applied in 5ae8cbde14c80e47691f8c71e4ff4b8372fc6a01

PR #346, opened 2016-12-20
UPBGE: Add an option to create objects list looping in c++

youle31:
temp work. creates objects list that match name or property. Keyword args

test file: http://pasteall.org/blend/index.php?id=44729

requires performances tests

Performance test file: http://pasteall.org/blend/index.php?id=44738 ... Wait... I fix a mistake in the code

So not big difference on a list of 100 objects. This only makes difference for really big lists. But before sending it to trash, maybe we could test with getMatching(materialName = "...") or with other things... Also on other computers than mine who has a powerful CPU

### DIFF
--- a/source/gameengine/Ketsji/KX_Scene.cpp
+++ b/source/gameengine/Ketsji/KX_Scene.cpp
@@ -2141,6 +2141,7 @@ PyMethodDef KX_Scene::Methods[] = {
 	KX_PYMETHODTABLE(KX_Scene, suspend),
 	KX_PYMETHODTABLE(KX_Scene, resume),
 	KX_PYMETHODTABLE(KX_Scene, drawObstacleSimulation),
+	KX_PYMETHODTABLE_KEYWORDS(KX_Scene, getMatching),
 
 	
 	/* dict style access */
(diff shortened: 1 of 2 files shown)


DCubix:
283fps = with getMatching
260fps = without
Seems like a good job to me :+1:

panzergame:
@youle31 : The function you added check for name and property, these two attributes are shared by all CValue sub class and CValue is contained in CListValue. Why not move this function to CListValue ?

source/gameengine/Ketsji/KX_Scene.cpp
```diff
@@ -2533,4 +2534,33 @@ KX_PYMETHODDEF_DOC(KX_Scene, get, "")
 	return def;
 }
 
+/* Find gameobjects matching a name, a property... */
+KX_PYMETHODDEF_DOC(KX_Scene, getMatching, "scene.getMatching(objectsName, objectsProperty...)")
+{
+	PyObject *objectslist = PyList_New(0);
+	char *objectsName = "";
```
panzergame: You could use NULL to remove strlen check later.

source/gameengine/Ketsji/KX_Scene.cpp
```diff
@@ -2533,4 +2534,33 @@ KX_PYMETHODDEF_DOC(KX_Scene, get, "")
...
+	char *objectsProperty = "";
+
+	static const char *kwlist[] = { "name", "property", NULL };
+
+	if (!PyArg_ParseTupleAndKeywords(args, kwds, "|ss:getMatching", (char **)(kwlist), &objectsName, &objectsProperty))
+		return NULL;
+
+	for (CListValue::iterator<KX_GameObject> it = m_objectlist->GetBegin(), end = m_objectlist->GetEnd(); it != end; ++it) {
+		KX_GameObject *gameobj = *it;
+		if (strlen(objectsName)) {
```
panzergame: The user want to have a list of object with giving name AND property. or a list of object with the giving name OR property ?

source/gameengine/Ketsji/KX_Scene.cpp
```diff
@@ -2533,4 +2534,33 @@ KX_PYMETHODDEF_DOC(KX_Scene, get, "")
 	return def;
 }
 
+/* Find gameobjects matching a name, a property... */
+KX_PYMETHODDEF_DOC(KX_Scene, getMatching, "scene.getMatching(objectsName, objectsProperty...)")
+{
+	PyObject *objectslist = PyList_New(0);
```
panzergame: Why not return a CListValue with python owner proxy ?

panzergame:
I propose you to use an optional regex system. For example if the user want to look at all the object starting with name "Cube" and containing the property "Prop":
```
scene.objects.getMatching("^Cube.*", "Prop")
```

Regex could be used with the C++11 regex library: http://en.cppreference.com/w/cpp/regex.

youle31:
Yes. About regular expressions, why not.
About this patch, I just begun to implement it to see if this could be increase performances when we create lists of objects. But for now, I don't know if this is really useful. For 100 objects, on my computer I see no performance difference beetween a list created with comprehension lists or with the actual getMatching. For 1000 objects we can see a diference but not as important as expected. So I don't know if I continue to work on it or not.

@DCubix : What do you think? What could be the usages? Do you think it's still interesting? Would you want to continue to work on it?

@panzergame && @DCubix : Do you think we could use this patch to: instead of just create a list of objects, we ask to apply a function/method for a list of objects: e.g: scene.getMatching(property = "prop", function = "applyRotation((0.0, 0.0, 0.1) False")) ?

lordloki:
For the perfomance test file I obtain 55fps using getMatching and 41fps using current way.

BluePrintRandom:
I need to batch update terrain mesh instances using a table of stored
vertex offsets,  normals and vertex color for splat map blending.

(so as the actor moves forward new patches are loaded and deformed and
textured and old patches behind you unload)

I will also need batch add object if possible

scene.addBatch([ ['buildingpiece', loc xyz, rot], ['buildingPiece2', loc
xyz, rot]])

youle31:
Thanks @lordloki for performances test. On your computer it seems it makes a big difference, so we could eventually continue to work on this patch.

youle31:
panzergame commited another approach with regex here https://github.com/UPBGE/blender/commit/4441d58bc3c051f61989761453650470b2245418 so I close this

Issue #319, opened 2016-12-05
Feature Request: List uniforms in BL_Shader


sdfgeoff:
I'd like to see the ability to list uniforms in a shader:

**Why?**
When shaders are compiled, unused uniforms are compiled out, and this can splat your console with "Invalid Uniform" warnings. So if you replace a shader with one to help you debug, then your console is useless. It doesn't even appear as a python error, so it can't be caught/handled.

Adding the ability to list uniforms would allow this to be checked beforehand to stop the errors happening.

**Suggested Implementation**
The simple solution:
`BL_Shader.getUniformList()`

The amazing solution:
`BL_Shader.uniform_dict = {'color':mathutils.Vector([0.5, 0.2, 0.1]), 'other_uniform':7}`

youle31:
http://pasteall.org/141496

I propose to say:

if (location == -1 && weAreInFirstFrame) {
           CM_Error("invalid uniform value: " << name << ".");
}

to avoid to slow down the shader after the first frame.

About uniform list, I look at https://www.opengl.org/wiki/Program_Introspection (Uniforms and blocks) to see if I can add an API.

youle31:
I don't know about this request if this is important enough.
@sdfgeoff : Do you have a test file that shows a case where this feature would be really/very useful ?
@panzergame : What do you think? Do you see other advantages to have uniforms list ?

panzergame:
The current API obligate the user to know it's uniform type, with a list or dict it will be a little be harder to set an uniform because the user have to know: the name, the type and the value.

@sdfgeoff : Maybe you though about an option for debugging ?

sdfgeoff:
The case where it would be useful is when writing more general shader management tools. In a complex system, you may have a shader management system, that wants to treat all shaders the same. In this case, being able to know if a shader has a specific uniform is important. 

I have a render pipeline made from 9 render to textures. Each stage has an associated shader on the plane it displays. Now some of these have common uniforms, such as zoom (influences blur radius), and some clipping plane data. But some of the shaders do not use these. As a result, I can't just do a blanket "for pass in render_passes: set_zoom_uniform(pass)" because it floods the console with warnings about invalid uniforms. (This system is huge and is under NDA, so I can't share it with you as a test blend).
I can't write a system that does shader text introspection either, because uniforms that aren't used get compiled out.

@panzergame
[...]

PR #348, opened 2016-12-21
UPBGE: Display "Invalid Uniform Value" error message only at the firs…

youle31:
…t run

This is to avoid to fill the console with repeated error messages each
frame. This is related to https://github.com/UPBGE/blender/issues/319

test file: http://pasteall.org/blend/index.php?id=44739
better test file: http://pasteall.org/blend/index.php?id=44742

### DIFF
--- a/source/gameengine/Rasterizer/RAS_Shader.cpp
+++ b/source/gameengine/Rasterizer/RAS_Shader.cpp
@@ -166,7 +166,9 @@ RAS_Shader::RAS_Shader()
 	m_use(0),
 	m_attr(0),
 	m_error(0),
-	m_dirty(true)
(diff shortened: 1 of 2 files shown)


panzergame:
@youle31 : i don't really see what we need to avoid error message repeat, if there's an error print it, don't try to make it more quiet. Also for 2d filter: it will not print an error for each unused object properties now ?

youle31:
@panzergame: the reason is explained here: #319

Else you're right, I restore debug

youle31:
For this I don't know, I feel not really concerned but it was just to answer a feature request. If a better alternative is found, I can close this request.

youle31:
Not important enough. I close

PR #353, opened 2016-12-31
UPBGE: Prevent cubemap rendering in some cases

youle31:
the issue with this patch is that we assume that 1) only one object is using the cube map 2) this object will keep the same visible state at the next frame (multiple viewports -> culling state)

test file: for the viewport issue: http://pasteall.org/blend/index.php?id=44875


youle31:
arf I think it's good now...

youle31:
Well, I'll try another method so I close this pull request for now

PR #355, opened 2017-01-03
UPBGE: Draw cameras frustum at runtime for debugging purposes

youle31:
http://pasteall.org/pic/show.php?id=110741

video test: https://youtu.be/70kriVlWBUY :P

Apply a shader to the frustum! http://pasteall.org/178675/diff

### DIFF
--- a/source/gameengine/Ketsji/KX_KetsjiEngine.cpp
+++ b/source/gameengine/Ketsji/KX_KetsjiEngine.cpp
@@ -612,7 +612,9 @@ void KX_KetsjiEngine::Render()
 				if (showCamerasFrustum) {
 					for (CListValue::iterator<KX_Camera> it = cameras->GetBegin(), end = cameras->GetEnd(); it != end; ++it) {
 						KX_Camera *cam = *it;
-						DrawActiveCameraFrustum(cam, scene);
+						if (cam != activecam) {
+							DrawActiveCameraFrustum(cam, scene);
+						}
 					}
 				}
 			}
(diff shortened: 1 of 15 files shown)


youle31:
KX_files are not included anymore in RAS ;)

source/gameengine/Ketsji/KX_KetsjiEngine.cpp
```diff
@@ -1744,34 +1745,27 @@ bool KX_KetsjiEngine::GetShowArmatures() const
...
-void KX_KetsjiEngine::DrawCameraFrustum(KX_Camera *cam, KX_Scene *scene)
+void KX_KetsjiEngine::DrawCameraFrustum(KX_Camera *cam, KX_Scene *scene, const RAS_Rect& viewport)
 {
 	bool perspective = cam->GetCameraData()->m_perspective;
 
-	Scene *sc = scene->GetBlenderScene();
-	float aspx = (float)sc->r.xsch * sc->r.xasp;
-	float aspy = (float)sc->r.ysch * sc->r.yasp;
-	float ratiox = min_ff(aspx / aspy, 1.0f);
-	float ratioy = min_ff(aspy / aspx, 1.0f);
```
youle31: @panzergame : This code was intended to match this behaviour: https://www.youtube.com/watch?v=be-ku3YVgZ8 beginning of the video

panzergame: Yes, my behaviour is wrong here, I noted the issue and maybe a little fix proposal in the framapad:
```
    It's pretty hard to draw the culling box of a camera which doesn't been used for culling (we don't know what the viewport/camera ratio will look like).

    I'm proposing to be able to render the scene from an other camera than the one used for culling and then be able to draw the frustum box of this culling camera. Is it correct ?
```

youle31: In the viewport you can see the behaviour of the camera (It should normally be the same as for the frustum in the ideal case). (add a camera, reset rotation, go in top view, look at the behaviour changing the resolution (1920x1080 or so))

panzergame:
Merge in master at 5a72f50b8aa9d4a34e8bb917a1edf709ec4f5719.

PR #359, opened 2017-01-08
Ge vibration leftright

lordloki:
Now the vibration actuator has 2 new methods (start and stop). Start method
starts the vibration and also it can update the parameters of the
vibration on fly. Stop method stops the vibration.

Also, there is 2 strength parameters: the "left" one sets the strength of
the Low frequency joystick's motor and the "right" one sets the strength
of the High frequency joystick's motor.

Additionally, 2 new bool attributes have been included (isVibrating and
hasVibration) along the new python methods.


youle31:
Hi! Minor issue (I let you decide if this is a bug or not): http://pasteall.org/blend/index.php?id=44994 (else this is still ok to me)

lordloki:
Yes, it is a bug. Thank you so much for test

lordloki:
It has been a long way as this part of SDL2 library is not very well documented (only few examples around the net and moreover very simples) and it is not clearly explained what functions or effects are clearly/partially/little supported but I think we are ready to merge it.

The behaviour mimic the sound actuator one.

youle31:
Hi! Thanks for your hard work. It's ok for me :) my test file: http://pasteall.org/blend/index.php?id=45021

source/gameengine/Device/DEV_Joystick.h
```diff
@@ -170,15 +170,20 @@ class DEV_Joystick
...
-	 * for rumble only via a python API.
+	 * Force Feedback - Vibration
+	 * We could add many optional arguments to these functions to take into account different sort of vibrations.
+	 * But we propose to keep the UI simple with only joyindex, force (in both motors) and duration.
+	 * As the vibration strength and duration can be updated on-fly it is possible to generate several types of vibration  
+	 * (sinus, periodic, custom, etc) using BGE python scripts for more advanced uses.
 	 */
-	bool RumblePlay(float strength, int duration);
+	bool RumblePlay(float strength_left, float strength_right, unsigned int duration);
+	bool RumbleStop(void);
```
panzergame: We are not in C, let's remove these `(void)` :)

panzergame:
@lordloki Could you prepare a commit message concerning the complex part in DEV_JoystickVibration ?
Thanks in advance.

lordloki:
2 new methods (start and stop) have been added to the vibration actuator. The Start method starts the vibration and it can update the parameters of the vibration on fly too. The Stop method stops the vibration.

Additionally, there is 2 strength parameters instead of one: the "left" one sets the strength of the Low frequency joystick's motor and the "right" one sets the strength of the High frequency joystick's motor.

Also, to check the vibration capabilities of the controller, 2 new bool attributes have been included (isVibrating and hasVibration).

In the implementation side:

To obtain the vibration, in first place, we checked for the LEFTRIGHT effect. This effect is very well supported in all XInput controllers. In case of this effect is not supported by the controller we added a back-up SINE effect that it is the most supported in all controllers but XInput ones. The cons of this back-up SINE effect is that only supports vibration for one of both motors (the low frequency one).
Additionally, we added another effect (CUSTOM). This effect is used by some NOT XInput controllers to make the LEFTRIGHT effect. Also, is the effect most supported by controllers in the Mac OS.

The vibration algorithm is:
- Define vibration effect type and parameters
- Create (or update if already exists) the effect
- Run the effect.

The vibration duration would have to be done using SDL_HapticGetEffectStatus function but this function is not well supported by the most current controllers. In its place, we used an own time management. Due to this, we have to process this own time management using the ProcessRumbleStatus function before we can get the rumble status too.

panzergame:
Merged.

PR #360, opened 2017-01-10
UPBGE: Basic implementation for cubemaps fresnel

youle31:
No python API yet. Neither fresnel corrections (could be nice to implement
an option to make the dark darker and the white whiter)

Neither the possibility to "invert fresnel" (don't know if it is really
useful though)

look at:
http://www.standardabweichung.de/code/javascript/webgl-glsl-fresnel-schlick-approximation

schlik fresnel approximation

test file: http://pasteall.org/blend/index.php?id=45022


youle31:
I think this small patch is ready to land (for a quick implementation of fresnel). I had feedbacks in my tutorials about cubemaps and some users requested this. But this can wait for 0.1.6.

youle31:
I can rework it to make it compatible not only with cubemaps if you want @panzergame

PR #375, opened 2017-01-24
UPBGE: Add an option to skip Invalid Uniform warnings

youle31:
test file: http://pasteall.org/blend/index.php?id=45225

### DIFF
--- a/doc/python_api/rst/bge_types/bge.types.BL_Shader.rst
+++ b/doc/python_api/rst/bge_types/bge.types.BL_Shader.rst
@@ -267,3 +267,10 @@ base class --- :class:`PyObjectPlus`
 
       Validate the shader object.
 
+   .. method:: SetInvalidUniformWarningDisplayed(value)
+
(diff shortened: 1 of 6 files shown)


panzergame:
One of the issue founding a uniform name error is that we use the same technique to detect if the name is mistyped and if the name is the one of an useless uniform.

So if we hide the warning we will silent the mistyped name errors whereas if we enable all the warning the useless uniform will be show as invalid name.

I think i will look at a parser library to know if an uniform doesn't exist or if it is just unused…

panzergame:
http://jonmacey.blogspot.fr/2012/02/bit-of-parsing.html

youle31:
ok, I you have the courage to parse shader... Why not? I'm personally not motivated enough to do that :P

youle31:
I close this as we don't all agree about. @panzergame : Sorry about misanderstanding about GetUniformLocation.
@DCubix : I'm not opposed to uniforms list feature :P, so if you want to work on there is no problem :)

PR #379, opened 2017-01-27
UPBGE: Move "world" physics properties to Scene panel

youle31:
The physics properties naturally belong to scene. Properties defined in
RNA_scene, scene.gm.settings etc...

Each scene can have its own physics properties.

Furthermore, no need to have a "World" to have physics properties.

Thanks @Blendify for judicious advice

EDIT: Finally this might be commited in BF bge


youle31:
Commited in BF blender

PR #390, opened 2017-02-11
UPBGE: Add unlimited Planar Reflections/Refractions

panzergame:
See #252

It needs to reimplement clip plane and UI refactor/design.


source/gameengine/Converter/BL_BlenderDataConversion.cpp
```diff
@@ -1858,25 +1857,16 @@ void BL_ConvertBlenderObjects(struct Main* maggie,
...
+							viewpoint = obj;
 						}
-
-						kxscene->GetCubeMapManager()->AddCubeMap(tex, viewpoint);
-					}
-					else {
-// 						EnvMap *env = tex->GetTex()->env;
-						KX_GameObject *viewpoint = gameobj;
-						// Prevent two sided rendering on mirror's material to avoid artifacts
-						polymat->SetRasMode(0);
```
youle31: iirc this was used to prevent two sided rendering on planar reflections because I had artifacts with backface culling enabled (because of switch beetween frontface/backface during planar rendering). I have to check if with your implementation the problems are gone.

source/gameengine/Ketsji/KX_TextureProbe.cpp
```diff
@@ -0,0 +1,250 @@
...
+{
+	return "KX_TextureProbe";
+}
+
+KX_GameObject *KX_TextureProbe::GetViewpointObject() const
+{
+	return m_viewpointObject;
+}
+
+void KX_TextureProbe::SetViewpointObject(KX_GameObject *gameobj)
```
youle31: I'm not sure this should be allowed for mirrors (The viewpoint object must not be necessarly the mirror object itself? If we change viewpoint object, doesn't it make the mirror reflection inaccurate?). But I didn't thought enough about it to see in what case it could be useful for mirrors.

source/gameengine/Ketsji/KX_TextureProbe.cpp
```diff
@@ -0,0 +1,250 @@
...
+}
+
+int KX_TextureProbe::pyattr_set_viewpoint_object(PyObjectPlus *self_v, const KX_PYATTRIBUTE_DEF *attrdef, PyObject *value)
+{
+	KX_TextureProbe *self = static_cast<KX_TextureProbe *>(self_v);
+	KX_GameObject *gameobj = NULL;
+
+	SCA_LogicManager *logicmgr = KX_GetActiveScene()->GetLogicManager();
+
+	if (!ConvertPythonToGameObject(logicmgr, value, &gameobj, true, "cubeMap.object = value: KX_TextureProbe"))
```
youle31: KX_TextureProbe is the base class for cubeMap/planar->reflections/refractions? The message "cubeMap.object" is only related to cubeMaps.

source/gameengine/Ketsji/KX_TextureProbeManager.cpp
```diff
@@ -0,0 +1,208 @@
...
+		const MT_Matrix4x4 viewmat(camtrans);
+
+		rasty->SetViewMatrix(viewmat, m_camera->NodeGetWorldOrientation(), m_camera->NodeGetWorldPosition(), m_camera->NodeGetLocalScaling(), m_camera->GetCameraData()->m_perspective);
+		m_camera->SetModelviewMatrix(viewmat);
+
+		m_scene->CalculateVisibleMeshes(rasty, m_camera, ~probe->GetIgnoreLayers());
+
+		/* Updating the lod per face is normally not expensive because a cube map normally show every objects
+		 * but here we update only visible object of a face including the clip end and start.
+		 */
```
youle31: I think the comment has to be updated (concerns not only cube maps)

youle31:
I'm testing (I'll edit if I see something special):
- needs to add condition for ImageNew in texture.c iirc to not resize down when texture type is TEX_ENVMAP and the other flag & ENV_PLANE iirc.

youle31:
About performances, I pushed ge_planar4 (ge_planar2 on last master), and ge_planar4 is a bit less performant than your branch, so the issue comes from elsewhere

PR #392, opened 2017-02-15
implement UI for Mipmap. TwisterGE proposal

youle31:
TwisterGE proposal.
Test file: http://pasteall.org/blend/index.php?id=45565

I added optionnal arg mipmap in add filter API. Missing API doc: REMOVED after panzergame review

Do we need to add something in versioning? ANSWERED by panzergame

### DIFF
--- a/source/gameengine/Converter/KX_ConvertActuators.cpp
+++ b/source/gameengine/Converter/KX_ConvertActuators.cpp
@@ -977,7 +977,7 @@ void BL_ConvertActuators(const char* maggiename,
 			}
 
 			tmp = new SCA_2DFilterActuator(gameobj, filtermode,  _2dfilter->flag,
-			                               _2dfilter->float_arg, _2dfilter->int_arg,
+			                               _2dfilter->float_arg, _2dfilter->int_arg, _2dfilter->mipmap,
 			                               ketsjiEngine->GetRasterizer(), scene->Get2DFilterManager(), scene);
 
 			if (_2dfilter->text)
--- a/source/gameengine/Ketsji/KX_2DFilterManager.cpp
+++ b/source/gameengine/Ketsji/KX_2DFilterManager.cpp
@@ -101,8 +101,9 @@ KX_PYMETHODDEF_DOC(KX_2DFilterManager, addFilter, " addFilter(index, type, fragm
 	int index = 0;
 	int type = 0;
 	const char *frag = "";
+	int mipmap = 0;
 
-	if (!PyArg_ParseTuple(args, "ii|s:addFilter", &index, &type, &frag)) {
+	if (!PyArg_ParseTuple(args, "ii|s:addFilter", &index, &type, &frag, &mipmap)) {
 		return nullptr;
 	}
 
@@ -124,6 +125,7 @@ KX_PYMETHODDEF_DOC(KX_2DFilterManager, addFilter, " addFilter(index, type, fragm
 	data.filterPassIndex = index;
 	data.filterMode = type;
 	data.shaderText = std::string(frag);
+	data.mipmap = mipmap;
 
 	AddFilter(data);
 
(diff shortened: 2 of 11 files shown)


source/gameengine/Rasterizer/RAS_2DFilterData.h
```diff
@@ -43,6 +43,8 @@ class RAS_2DFilterData
 	std::vector<std::string> propertyNames;
 	/// The KX_GameObject (or something else?) that provides the values for the uniforms named above.
 	CValue *gameObject;
+	/// Enable/Disable mipmap in 2DFilter actuator at conversion
+	int mipmap;
```
panzergame: Why do you not use a bool and do an implicit cast at the assignment ?

source/gameengine/Ketsji/KX_2DFilterManager.cpp
```diff
@@ -101,8 +101,9 @@ KX_PYMETHODDEF_DOC(KX_2DFilterManager, addFilter, " addFilter(index, type, fragm
 	int index = 0;
 	int type = 0;
 	const char *frag = "";
+	int mipmap = 0;
 
-	if (!PyArg_ParseTuple(args, "ii|s:addFilter", &index, &type, &frag)) {
+	if (!PyArg_ParseTuple(args, "ii|s:addFilter", &index, &type, &frag, &mipmap)) {
```
panzergame: I doubt for this part. The user have the possibility to set the mipmap just after he created the filter as he sets uniforms.

panzergame:
@DCubix : If we let the mipmapping disabled as default, you don't need to add a version conversion.

source/gameengine/Rasterizer/RAS_2DFilterData.h
youle31: There is everywhere int before conversion in ras-2dfilter. Due to bool not supported in dna

source/gameengine/Ketsji/KX_2DFilterManager.cpp
youle31: Yeah we Can remove it

youle31:
Ok no need for versioning. Thanks for review

source/gameengine/Converter/KX_ConvertActuators.cpp
```diff
@@ -977,7 +977,7 @@ void BL_ConvertActuators(const char* maggiename,
 			}
 
 			tmp = new SCA_2DFilterActuator(gameobj, filtermode,  _2dfilter->flag,
-			                               _2dfilter->float_arg, _2dfilter->int_arg,
+			                               _2dfilter->float_arg, _2dfilter->int_arg, _2dfilter->mipmap,
```
panzergame: @youle31: if you set the constructor argument type of mipmap to a bool, the int value from _2dfilter->mipmap will be casted to it implicitly, if you compiler complain to an explicit cast: `bool(value)`.

youle31: It Is an ont no in filter constructor? I can't check. I'm at Montpellier. I'll see tomorrow

youle31: Int

youle31: the nsca_2dfilteractuator constructor is:
SCA_2DFilterActuator(
	        class SCA_IObject* gameobj,
	        int type,
	        short flag,
	        float float_arg,
	        int int_arg,
	        int mipmap,
	        RAS_IRasterizer* rasterizer,
			RAS_2DFilterManager *filterManager,
	        SCA_IScene* scene);

source/gameengine/Rasterizer/RAS_2DFilterData.h
youle31: I do the cast in ras_2dfilter

source/gameengine/Converter/KX_ConvertActuators.cpp
panzergame: What i mean is that you could let the conversion from int to bool only in this file. And so use bool everywhere in SCA_2DFilterActuator and KX_2DFilterData.

youle31: okok I'll update. Thanks for review!

youle31: I changed that. You're right. I didn't realized my mistake. Thanks for have pointing it.

PR #401, opened 2017-02-26
UPBGE: Add a button to resize cubemap texture

youle31:
The cubemap texture height must be a power of 2 and the image width must be = 3/2*height


panzergame:
All is good for me, you could commit when the 0.1.6 will begin.

youle31:
Thanks for review! Can't we consider this as a bug fix and commit it before 0.1.6 (in 15 days)... We'll see later. take your week and more! "Please do not answer to this email"

PR #402, opened 2017-02-27
UPBGE: Quick refactor of texts drawing

### LINKED ISSUE #396: Bug: Profile texts not drawn correctly
http://pasteall.org/pic/show.php?id=112836
http://pasteall.org/blend/index.php?id=45661

youle31:
See comments on both commits.

test file: http://pasteall.org/blend/index.php?id=45723

Fixes #396, improves performances in some files and make the profile more "accurate" I guess (as the drawing profile takes less performances).

I didn't noticed different behavior with master (except the fixed bug).

The part in blf can be simplified I guess because some code is already in game engine code... But as it is, I think it costs nothing to commit.


youle31:
Wait for other plan for @DCubix or @panzergame .

At least the mini fix for enlightened profile texts can be commited I think

PR #403, opened 2017-02-28
UPBGE: Don't display IOR and MixRatio when we are in PLANE mapping

youle31:
for environment maps


youle31:
I think this can be commited. @panzergame : What do you think? Previously, IOR and mix factor appeared in Image ot Movies when the texture_coords was RELFECTION.
With this patch it appears only if we have an environment map and the "mapping" is "CUBE" and the texture_coords are "REFLECTION". This can be considered as a "bug fix or a cleanup".

youle31:
Ok with your last commit. I didn't saw the else (maybe you force --pushed) which was wrong in the case we had an Image or Movie and a mapping set to reflection.

For me this is good.

PR #404, opened 2017-02-28
UPBGE: Temporarly remove parallax uv shift option

youle31:
This option was already disabled in
https://github.com/UPBGE/blender/commit/686112da3bdb527447933fbf1fa0f7684ae433ca
This leads to confusion for the user:
https://github.com/UPBGE/blender/issues/354

There was a real problem with naming in the code:
- float scale in the shader (not explicit)
- heightscale in the UI (not explicit)
- parallax uv or parallax uv shift (last name is the most explicit name
  for me)

I think if we want such an option we have to allow shifting on U and V
axis.

Waiting we find a new solution, I propose to remove this from the UI and
the code.


youle31:
Hello! @lordloki : I know that you wanted to keep this option, but as it has already been disabled in a previous patch, would you agree to disable it the time we replace it with something else? (because there is a button that does nothing for now in the UI). Thanks!

lordloki:
@youle31 : Totally agree. Much better to remove the slider than to have it in IU not working. +1

youle31:
commited. Thanks for review :)

PR #405, opened 2017-02-28
Refactor blender converter.

panzergame:
Previously the KX_BlenderSceneConverter was shared between all the scene and used by all the scene during conversion. This was dangerous because the blender scene converter was containing some map of blender data to BGE data (Object to KX_GameObject for example)  and these map could be written in the same time by asynchronous lib loading.

To avoid this issue the KX_BlenderSceneConverter must be unique per scene to convert/load. The scene converter is also temporary because it is merged at the end of the scene conversion into the new class KX_BlenderConverter. This class contains a list of slot corresponding to the scene and owning the BGE datas (materials, meshes ect…). We don't reuse the KX_BlenderSceneConverter because we don't need and don't want to be able to use the blender data maps.


source/gameengine/Converter/BL_BlenderDataConversion.cpp
```diff
@@ -904,7 +900,7 @@ static void BL_CreatePhysicsObjectNew(KX_GameObject* gameobj,
 	}
 }
 
-static KX_LodManager *lodmanager_from_blenderobject(Object *ob, KX_Scene *scene, KX_BlenderSceneConverter *converter, bool libloading)
+static KX_LodManager *lodmanager_from_blenderobject(Object *ob, KX_Scene *scene, KX_BlenderSceneConverter& converter, bool libloading)
```
panzergame: I set the KX_BlenderSceneConverter to reference instead of pointer to forbid store it in a class without copying it.

PR #410, opened 2017-03-03
UPBGE: Add sky texture option in world tab

youle31:
TODO:

- Split shader in several shaders (do as for normal cubemaps with IOR and mixratio)
- Add an option to map reflection according ti view position instead of
  view normal (why? Because if we have a worldtex that represents a room,
  the objects must reflect the context room according to their view
  position because the room is near, whereas if we have a distant sky, the
  mapping must be done according to view normal)


panzergame:
Can these option move into the world texture ?

youle31:
yes, you're right, I'll do that. Thanks for review

PR #412, opened 2017-03-04
UPBGE: Add an option to debug light shadow frustum and try to check intersection (old)

youle31:
see old branch for comments. up to date with master 03/03/2017

### DIFF
--- a/release/scripts/startup/bl_ui/properties_game.py
+++ b/release/scripts/startup/bl_ui/properties_game.py
@@ -790,6 +790,8 @@ class DATA_PT_shadow_game(DataButtonsPanel, Panel):
         col.prop(lamp, "shadow_color", text="")
         if lamp.type == 'SUN':
             col.prop(lamp, "show_shadow_box")
+        if lamp.type in ('SUN', 'SPOT'):
+            col.prop(lamp, "show_shadow_box_bge")
         col.prop(lamp, "static_shadow")
 
         col = split.column()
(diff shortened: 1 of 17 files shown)


youle31:
This needs to be rewritted in my opinion. And DrawDebugCameraBox too...

To make the method BoxInsideFrustum work, we have to contruct the box exactly as it is done in: MT_Vector3 box[8];
gameobj->GetSGNode()->GetBBox(box);

Else BoxInsideFrustum won't work I think.

youle31:
Not necessarly DrawDebugCameraBox. But we need a custom box for light compatible with BoxInsideFrustum, and why not take the same box to draw and to check if it is inside camera frustum?

youle31:
it seems to ~work but in the drawing doesn't correspond exactly to the calculation of intersection...

youle31:
test file: http://pasteall.org/blend/index.php?id=45804 (nothing works as expected...)

youle31:
cam->GetModelviewMatrix() returns identity

youle31:
So it seems my code is correct. It works in most cases but BoxInsideFrustum has an issue (a known issue on the net) because the computation is not 100% accurate and doesn't handle every case, particularly with large boxes. There are tons of articles which talk about frustum and box intersection, and possibility of optimizations. So it could be the occasion to refactor/optimize BoxInsideFrustum or in general culling computation.

youle31:
needs refactor of BoxInsideFrustum

panzergame:
@youle31 : Was the computing of the box using a matrix working ?

youle31:
@panzergame : We can also try this: http://www.yosoygames.com.ar/wp/2016/12/frustum-vs-pyramid-intersection-also-frustum-vs-frustum/ for spots... Now that I ~understood how it works I guess I can implement it...

youle31:
I close this as I have a new branch

PR #413, opened 2017-03-05
UPBGE: Fix errors signaled by cppcheck

youle31:
2 mem leaks.

### DIFF
--- a/source/gameengine/Ketsji/KX_NavMeshObject.cpp
+++ b/source/gameengine/Ketsji/KX_NavMeshObject.cpp
@@ -368,12 +368,7 @@ bool KX_NavMeshObject::BuildNavMesh()
 	const int dataSize = headerSize + vertsSize + polysSize + nodesSize +
 		detailMeshesSize + detailVertsSize + detailTrisSize;
 	unsigned char* data = new unsigned char[dataSize];
-	if (!data) {
-		if (vertsi) {
-			delete [] vertsi;
-		}
-		return false;
-	}
+
 	memset(data, 0, dataSize);
 
 	unsigned char* d = data;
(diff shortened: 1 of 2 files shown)


source/gameengine/Ketsji/KX_NavMeshObject.cpp
```diff
@@ -368,8 +368,12 @@ bool KX_NavMeshObject::BuildNavMesh()
 	const int dataSize = headerSize + vertsSize + polysSize + nodesSize +
 		detailMeshesSize + detailVertsSize + detailTrisSize;
 	unsigned char* data = new unsigned char[dataSize];
-	if (!data)
+	if (!data) {
```
panzergame: Well, this check seems useless finally as we don't check for every memory allocation, and that, i hope, the `dataSize`is never 0.

source/gameengine/Physics/Bullet/CcdPhysicsController.cpp
```diff
@@ -1600,6 +1600,9 @@ PHY_IPhysicsController *CcdPhysicsController::GetReplicaForSensors()
 
 			default:
```
panzergame: @youle31 : Could you check that `cinfo.m_collisionShape = m_shapeInfo->CreateBulletShape(m_cci.m_margin, m_cci.m_bGimpact, !m_cci.m_bSoft);` can do the full work ?

youle31: I can't reach this part of the code

panzergame: Well, i doubt that there's a real error here:

The m_collisionShape could be leaked by the return nullptr only if this one was possible when the if (m_shapeInfo) succeed. But this is not the case as the return is in the switch in the else if.

If `cinfo.m_collisionShape = m_shapeInfo->CreateBulletShape(m_cci.m_margin, m_cci.m_bGimpact, !m_cci.m_bSoft);` is called then `return nullptr` is never called and the memory should not leak.

I hope i'm right……

youle31: So I close this?

panzergame:
Merge in master: 30a939fbe869d38aca6350fa709531c69ac11ab2.

PR #417, opened 2017-03-07
UPBGE: Fix replication of armature constraint objects.

panzergame:
The object used for the armature constraint are blender objects, but previously
when an armature were replicated the blender objects wasn't, so the solve of
the constraints was wrong.

To solve this issue we need to use a new blender object, instead of replicating
the original blender object targeted, we simply use a dummy object which is
an enpty blender object created by each BL_ArmatureConstraint, this object is
set to the constraints calling flush_constraint_targets with no_copy to 0.

For constraint targeting an other armature bone, the pose is set into the blender
object in BL_ArmatureConstraint::UpdateTarget.

These object are then freed at the BL_ArmatureContraint desctruction with a pose
set to nullptr to avoid free a not owned armature pose.

@youle31, @lordloki, @DCubix : Could you please test this branch with any file using an armature with a bone constraint and review this branch too ?

### DIFF
--- a/source/gameengine/Converter/BL_ArmatureConstraint.cpp
+++ b/source/gameengine/Converter/BL_ArmatureConstraint.cpp
@@ -30,13 +30,20 @@
  */
 
 
-#include "DNA_constraint_types.h"
-#include "DNA_action_types.h"
 #include "BL_ArmatureConstraint.h"
 #include "BL_ArmatureObject.h"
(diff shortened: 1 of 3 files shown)


youle31:
it's ok with my test files :) . I can't really comment the code as I don't really understand it. Good job and Thanks!

PR #420, opened 2017-03-08
UPBGE: Add a check for empty value in property sensor

### LINKED ISSUE #419: Bug: LB setup with property bricks crash.
blend file: http://pasteall.org/blend/index.php?id=45839
Ive made a test logic brick setup with property bricks  and get a crash when p key pressed. Crashes in upbge 014 and master, no crash in bf bge.

youle31:
fixes #419

### DIFF
--- a/source/gameengine/GameLogic/SCA_PropertySensor.cpp
+++ b/source/gameengine/GameLogic/SCA_PropertySensor.cpp
@@ -113,6 +113,9 @@ SCA_PropertySensor::~SCA_PropertySensor()
 
 bool SCA_PropertySensor::Evaluate()
 {
+	if (!(strlen(m_checkpropval.c_str()))) {
+		return false;
+	}
 	bool result = CheckPropertyCondition();
 	bool reset = m_reset && m_level;
 	


source/gameengine/GameLogic/SCA_PropertySensor.cpp
```diff
@@ -113,6 +113,9 @@ SCA_PropertySensor::~SCA_PropertySensor()
 
 bool SCA_PropertySensor::Evaluate()
 {
+	if (!(strlen(m_checkpropval.c_str()))) {
```
panzergame: You could use std::string, also i think the better place is in CheckPropertyCondition because this function is the only one reading the property string (as far as i see).

panzergame: std::string::empty :)

youle31: evaluate is the function called each frame to see if the sensor send true pulse or not. I think it is better to stop at the beginning of evaluate

PR #422, opened 2017-03-10
UPBGE: Apply modifiers at game start for rendering part to avoid slow rendering

youle31:
**APPLY MODIFIERS AT GAME START FOR RENDERING PART TO AVOID SLOW RENDERING**

I think:
- It restores the performances as before RAS_ListRasterizer removal
- RAS_ListRasterizer was to render only the meshes without armatures I think
-----------------------------------------------------------------------------------------------
- So the meshes with an armature parent are rendered with the viewport code, as in BF.
- The meshes without armature but with modifiers are rendered with RAS_VBO.cpp.
-----
- All modifiers available in embedded will be available in standalone.
-----------------------------------------------------------------------------------------------

**Issues:**
- Problem with objects with modifiers and multiple mesh users: **Fixed**
- Problem with modifier Armature + modifier subsurf: http://pasteall.org/blend/index.php?id=45877: **Fixed**
- Dark_War: with libload (iirc the game shared by joel gomes in the issue related to libfree) (crash at start): **Fixed**
- The file BPR doesn't want we share (the head doesn't move... issue with constraints?): **Fixed**

**Test files:**
- For issue with armature + subsurf modifier: http://pasteall.org/blend/index.php?id=45896
- For issue with BPR file, I can't share it here because he doesn't want we share it on the net but panzergame and me have it. The issue was related to mesh name in the logicmanager. I kept same name as before this patch in the logicmanager and for KX_MeshProxy, but another name for RAS_MeshObject (the mesh rendered).
- For issue with multiple mesh users: http://pasteall.org/blend/index.php?id=45897 (By the way you can test physics behaviour in this file). The original issue was that when we have an object with modifiers, if it has a mesh shared by several objects in viewport, at runtime we have to make the ras mesh unique for the rendering part. See commit message in old branch.
- For libload issue, see bge_dw test file here: https://github.com/UPBGE/blender/issues/382 (Fixed by adding a check for null blenderobj in BL_ConvertMesh.

**Performances:**
- Modifiers not applied: https://drive.google.com/file/d/0B3GouQIyoCmrRXRYY21PNVB2LVE/view?usp=sharing
- Modifiers applied: https://drive.google.com/file/d/0B3GouQIyoCmrdm40TU1pT1p5R00/view?usp=sharing

**Historic:**
- https://github.com/UPBGE/blender/pull/361
- https://github.com/UPBGE/blender/tree/ge_nomodifier

**Note:**
In this patch I didn't included patch for physics DerivedMeshes removal made by panzergame because I had issues with this commit and I didn't noticed slowdowns for Physics DerivedMeshes (the performance leak comes only from the rendering part as far I saw).

### DIFF
--- a/source/gameengine/Converter/BL_BlenderDataConversion.cpp
+++ b/source/gameengine/Converter/BL_BlenderDataConversion.cpp
@@ -1076,7 +1076,7 @@ static KX_GameObject *gameobject_from_blenderobject(
 		bool bHasArmature = (BL_ModifierDeformer::HasArmatureDeformer(ob) && ob->parent && ob->parent->type == OB_ARMATURE && bHasDvert);
 
 		/* Since we apply modifiers at game engine start, we set bHasModifier to false by default */
-		bool bHasModifier = false; //BL_ModifierDeformer::HasCompatibleDeformer(ob);
+		bool bHasModifier = bHasArmature ? BL_ModifierDeformer::HasCompatibleDeformer(ob) : false;
 
 #ifdef WITH_BULLET
 		bool bHasSoftBody = (!ob->parent && (ob->gameflag & OB_SOFT_BODY));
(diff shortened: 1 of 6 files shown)


youle31:
Voilà I think it works with all my test files now...

panzergame:
ge_no_modifier answer this goal.

PR #432, opened 2017-03-17
Ge fresnel2

youle31:
I did it for textures but in fact this should be done for materials in material tab.......... I'm not sure in fact....

http://pasteall.org/blend/index.php?id=45953


panzergame:
@youle: I would like to refactor the fresnel settings:

I want to follow the implementation from https://www.blend4web.com/doc/en/materials.html#fresnel-effect-for-reflection to match with the fresnel method used in blend4web.

This method has two settings: fresnel and blend. Once these settings you can use a fresnel factor per textures.

The fresnel and blend values will be shared per texture and then stored in the material UI panel. Else the fresnel factor will be stored in the texture UI panel with a check at is left as the other influence setting for textures.

@youle31, @DCubix, @lordloki : What do you think about this proposal ?

@youle31 : Maybe we could add a menu choosing between the current fresnel implementation (from the material node) and the Schlick’s implementation ?

adriansnetlis:
I haven't tried out, yet I want to note things about fresnel in real-time apps:

- Suggested calculation for fresnel is float fresnel = pow(dot(normal, view), 5.0);
- For F0 to look right it must be mix()-ed not max()-ed (mix(f0, fMax, fresnel) not max(f0, fresnel * fMax))
- f0 and fMax must both be adjustable. If you wish, you can also make the 5.0 in first equation adjustable

Well, note that this is all just my opinion.

youle31:
I don't know, do what you think is good

PR #437, opened 2017-03-19
UPBGE: Add an option to debug light shadow frustum and try to check intersection (new)

youle31:
I'm so lucky :) It worked at the first try

adapted from: 

- http://www.yosoygames.com.ar/wp/2016/12/frustum-vs-pyramid-intersection-also-frustum-vs-frustum/ (8 points instead of pyramid)
- and http://www.iquilezles.org/www/articles/frustumcorrect/frustumcorrect.htm
------------
- Fixes all kind of boxes accepted in the intersection check (spot shape)
- Use matrices everywhere instead of "geometry calculus"
--------
I can't believe this seems to work 0;0


youle31:
I close this to replace it with a new branch ut to date with master

PR #438, opened 2017-03-19
UPBGE: Shadow Culling and Debug Light Frustum

youle31:
*Branch up to date with master 19/03/2017.*
---------------  
- Supports Sun and Spot shapes (in intersection check)
- Use Matrices instead of geometry calculus.
---------------  
- Source: http://www.yosoygames.com.ar/wp/2016/12/frustum-vs-pyramid-intersection-also-frustum-vs-frustum/
---------------
1. Normal test files:  
- Test file: http://pasteall.org/blend/index.php?id=45977 (I animated the light with location keyframes)
- Test file2: http://pasteall.org/blend/index.php?id=45978  
- Culling test in top view: http://pasteall.org/blend/index.php?id=45987
---------------
2. Performances test files: 
- http://pasteall.org/blend/index.php?id=45985 (Each cube is subdivided 5 times (under each spot))
- http://pasteall.org/blend/index.php?id=45988 (Move the camera---Compare with master)
---------------
Pics:  
- Master: http://pasteall.org/pic/show.php?id=113721  
- Shadow Culling ON: http://pasteall.org/pic/show.php?id=113722  (green means the spot frustum is intersecting with camera frustum. I disabled shadow culling for viewport camera (__default_cam_) to have a better visibility when we are in viewport view.)
---------------  
- Maybe the current culling test for geometry rendering is false/wrong/bad (maybe we should have more FPS if we discard false positives in the culling test for geometry like we did for shadows). In the performance test file we could disable physics to have a better test file for normal culling performances.
**EDIT: I tested and didn't noticed differences.**
- If we want to avoid backward compatibility issues, we can add an UI option to disable shadow culling (but I propose to enable it by default on suns and spots)

### DIFF
--- a/release/scripts/startup/bl_ui/properties_game.py
+++ b/release/scripts/startup/bl_ui/properties_game.py
@@ -790,6 +790,8 @@ class DATA_PT_shadow_game(DataButtonsPanel, Panel):
         col.prop(lamp, "shadow_color", text="")
         if lamp.type == 'SUN':
             col.prop(lamp, "show_shadow_box")
+        if lamp.type in ('SUN', 'SPOT'):
+            col.prop(lamp, "show_shadow_box_bge")
         col.prop(lamp, "static_shadow")
 
         col = split.column()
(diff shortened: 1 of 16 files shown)


panzergame:
@youle31 : Did ge_lightfrustum contains the minimum for only the debug drawing ?

youle31:
yes, there is all what you need here. And the branch is based on the master from yesterday. I think both light shadow frustum debugging and shadow culling should be commited at the same time because I used light frustum debugging to check intersection. If there is an intersection, the frustum becomes green.

youle31:
Still false positives...

youle31:
panzergame has another branch ge_cull so I close this pull request

PR #448, opened 2017-03-30
UPBGE: Branch to test/fix bullet culling

youle31:
To fix this bug: https://github.com/UPBGE/blender/issues/443

Adds a simple test before bullet culling test to avoid false positives. If you want to restore false positives, just remove out = PreIntersectionCheck(corners, se.node->volume.Mins(), se.node->volume.Maxs()); line 1118 in E:\Compilation\Blender\upbgeSources\extern\bullet2\src\BulletCollision\BroadphaseCollision\btDbvt.h

I add it in pull request just to have a better visibility

test file: http://pasteall.org/blend/index.php?id=46072


youle31:
Warning: Wrong code

PR #453, opened 2017-04-02
Implement CListValue name and property regex search.

panzergame:
This branch implement a method named `getMatching`in `CListValue`which take two arguments `name` and `prop`. These two arguments are used in a regex search of the list item which have a matching name or a matching property.

Example file: http://pasteall.org/blend/index.php?id=46149.

### DIFF
--- a/source/gameengine/Expressions/intern/ListValue.cpp
+++ b/source/gameengine/Expressions/intern/ListValue.cpp
@@ -18,6 +18,7 @@
  */
 
 #include <stdio.h>
+#include <regex>
 
 #include "EXP_ListValue.h"
 #include "EXP_StringValue.h"
@@ -483,6 +484,7 @@ PyMethodDef CListValue::Methods[] = {
 
 	// Dict style access.
 	{"get", (PyCFunction)CListValue::sPyget, METH_VARARGS},
+	{"getMatching", (PyCFunction)CListValue::sPygetMatching, METH_VARARGS},
 
 	// Own cvalue funcs.
 	{"from_id", (PyCFunction)CListValue::sPyfrom_id, METH_O},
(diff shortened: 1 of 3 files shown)


source/gameengine/Expressions/intern/ListValue.cpp
```diff
@@ -598,6 +600,54 @@ PyObject *CListValue::Pyget(PyObject *args)
...
+		if (strlen(namestr) == 0 || std::regex_match(item->GetName(), namereg)) {
+			if (strlen(propstr) == 0) {
+				result->Add(item);
+			}
+			else {
+				const std::vector<std::string> propnames = item->GetPropertyNames();
+				for (const std::string& propname : propnames) {
+					if (std::regex_match(propname, propreg)) {
+						result->Add(item);
+						break;
```
youle31: hello, why break here?

source/gameengine/Expressions/intern/ListValue.cpp
```diff
@@ -598,6 +600,54 @@ PyObject *CListValue::Pyget(PyObject *args)
...
+					}
+				}
+			}
+		}
+	}
+
+	return result->NewProxy(true);
```
youle31: this totally free memory for CListValue *result = new CListValue(); ?

youle31:
Sounds great :) I just have to learn how to use regex now :P

source/gameengine/Expressions/intern/ListValue.cpp
panzergame: It's to avoid pushing the item in the result list multiple time if it has more than one property name matching.

source/gameengine/Expressions/intern/ListValue.cpp
panzergame: Yes, the `true` in `NewProxy`means that the python will free the CListValue when the python object will be freed.

source/gameengine/Expressions/intern/ListValue.cpp
youle31: don't we want a list of ALL the items matching a property name?

panzergame: If two properties are matching the `prop` regex the item will be added twice in the list without a break, we want to avoid this.

youle31: Can't we do something like that:
if (result->find(item)) { continue; } so the item is not added several times but other items can be added to result list?

panzergame: The break is only in the property loop not in the item loop, so every items are iterated.

youle31: oh! sorry :)

panzergame:
Merged in master : 4441d58bc3c051f61989761453650470b2245418

PR #461, opened 2017-04-13
Implement font object culling bounding box.

panzergame:
Previously the font objects wasn't doing any culling computation they there
always visible. This behaviour have three desavantages:
- the render can't be optimised with culled font object
- the sources request to indentify font objects at rendering
  stage to don't care about the culling state
- the culling state from python is totally unusable.

To solves all the issues the best way is to implement a computation of the
bounding box for the fonts. This computation is made in the function
KX_FontObject::GetTextAabb. This function get the text AABB using BLF
function BLF_boundbox with the bound settings for dpi and size.

To make sure the AABB is computed in the wanted cases, the modification
of the text is done calling the function SetText. This function set the
text and compute the AABB.

One of the encountered issue was that the font objects could use a property
name "Text", but this property was read in UpdateBucket just before the
render and the AABB need to be computed before.
To avoid this issue the scene is iterating over all the font objects at the
logic frame end and calling the function KX_FontObject::UpdateTextFromProperty.
This function compare the current text with the text in a possible property
named "Text" and copy (and compute AABB so) if they are different. To compare
the font object have to store the full text into a full string and a array of
string per lines, this is why SetText use as argument a full string instead
of an array.

In the same time a bug with KX_FontObject.dimensions in python was noticed.
Indeed this function was based on a BLF function returing the size of a
single string, this technique totally failed for a multilines font object
even applying a line spaccing or something else.
The solution was to get the width and height of the bounding box.

Previously the culling pass was checking that the object were using a mesh
before process theirs culling, but this fails for font objects. Instead
checking the number of mesh we call the function KX_GameObject::UseCulling.

@youle31, @DCubix, @lordloki : Could you review this patch please ?

### DIFF
--- a/source/gameengine/Ketsji/KX_Scene.cpp
+++ b/source/gameengine/Ketsji/KX_Scene.cpp
@@ -1446,7 +1446,7 @@ void KX_Scene::DrawDebug(RAS_DebugDraw& debugDraw)
 		for (CListValue::iterator<KX_GameObject> it = m_objectlist->GetBegin(), end = m_objectlist->GetEnd(); it != end; ++it) {
 			KX_GameObject *gameobj = *it;
 
-			if (!gameobj->GetCulled() && gameobj->GetMeshCount() != 0) {
+			if (!gameobj->GetCulled() && gameobj->UseCulling()) {
 				const MT_Vector3& scale = gameobj->NodeGetWorldScaling();
 				const MT_Vector3& position = gameobj->NodeGetWorldPosition();
 				const MT_Matrix3x3& orientation = gameobj->NodeGetWorldOrientation();
@@ -1637,6 +1637,10 @@ void KX_Scene::LogicEndFrame()
 	//prepare obstacle simulation for new frame
 	if (m_obstacleSimulation)
 		m_obstacleSimulation->UpdateObstacles();
+
+	for (CListValue::iterator<KX_FontObject> it = m_fontlist->GetBegin(), end = m_fontlist->GetEnd(); it != end; ++it) {
+		(*it)->UpdateTextFromProperty();
+	}
 }
 
 
--- a/source/gameengine/Converter/BL_ArmatureObject.cpp
+++ b/source/gameengine/Converter/BL_ArmatureObject.cpp
@@ -399,7 +399,7 @@ void BL_ArmatureObject::ProcessReplica()
 	GetActionManager();
 }
 
-int BL_ArmatureObject::GetGameObjectType()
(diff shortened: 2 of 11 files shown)


source/gameengine/Ketsji/KX_Scene.cpp
```diff
@@ -1637,6 +1637,10 @@ void KX_Scene::LogicEndFrame()
 	//prepare obstacle simulation for new frame
 	if (m_obstacleSimulation)
 		m_obstacleSimulation->UpdateObstacles();
+
+	for (CListValue::iterator<KX_FontObject> it = m_fontlist->GetBegin(), end = m_fontlist->GetEnd(); it != end; ++it) {
```
youle31: can we use c++11 style here? for (CValue KX_FontObject *f: m_fontlist) {

youle31:
In BGMC 22 Game with no name, the bounding box seems too big, like if there were 2 empty lines at the bottom of the text (on X axis this is good, on Y the box seems too large. Can we remove the 2 empty lines at the bottom to make the box smaller?)

panzergame:
For BGMC 22 i think that the text have a `\n` at its end which creating a new empty line.

source/gameengine/Ketsji/KX_Scene.cpp
panzergame: Currently we can't but i investigated on making it possible, i'm going to back to my investigations.

lordloki:
It seems good and very logical to me. I will try to make a branch removing options from Text/Font tab when we are under Game Engine as only a few options are working.

BluePrintRandom:
Thank you Tristian

youle31:
Hello, I did tests again, The bounding box sounds nice (I checked in game with no name and it seems normal). I just noticed something weird in this file: http://pasteall.org/blend/index.php?id=46279 (Maybe this is not related to this branch). When you scale negative on X axis, we have not the same behaviour as when you negative scale on Y axis.

panzergame:
@youle31: The object center isn't the same as the bounding box center, the one displayed. For text the object center is top-left. I think that explains your doubts, no ?

Anyway, i have a worrying issue in your file, when the x scale is negative i have all my memory (4Go ram + 4Go swap) filled……… @youle31 : Can you reproduce this issue of it is only in `ge_text` ?

youle31:
No, I already noticed this issue. This was why I clamped the text size in KX_Font API... I have no idea why it does that

panzergame:
Merged in master.

PR #467, opened 2017-04-23
Ge activityculling

youle31:
test file for physics: http://pasteall.org/blend/index.php?id=46325

new test file for logic: http://pasteall.org/blend/index.php?id=46326

test file for suspend/restore constraints: http://pasteall.org/blend/index.php?id=46337

Ask panzergame for structure/code organization/Default values

Possible options:
- Add the possibility to adjust physics culling according to active camera (at runtime) clipend and a factor -> e.g: physicsCullingRadius = 1.5 x mainCameraClipend
- Realtime API

### DIFF
--- a/source/blender/makesdna/DNA_object_types.h
+++ b/source/blender/makesdna/DNA_object_types.h
@@ -301,6 +301,8 @@ typedef struct Object {
 	float ima_ofs[2];		/* offset for image empties */
 	ImageUser *iuser;		/* must be non-null when oject is an empty image */
 
+	float physicsCullingRadius, logicCullingRadius; /* For game engine, values around active camera where physics or logic are suspended */
+
 	ListBase lodlevels;		/* contains data for levels of detail */
 	LodLevel *currentlod;
 	float lodfactor, pad4;
--- a/release/scripts/startup/bl_ui/properties_game.py
+++ b/release/scripts/startup/bl_ui/properties_game.py
@@ -889,6 +889,24 @@ class OBJECT_MT_culling(ObjectButtonsPanel, Panel):
         layout.label(text="Predefined Bound:")
         layout.prop(game, "predefined_bound", "")
 
+class OBJECT_PT_activity_culling(ObjectButtonsPanel, Panel):
+    bl_label = "Activity Culling"
+    COMPAT_ENGINES = {'BLENDER_GAME'}
+
+    @classmethod
+    def poll(cls, context):
+        ob = context.object
+        return context.scene.render.engine in cls.COMPAT_ENGINES and ob.type not in {'CAMERA', 'EMPTY', 'LAMP'}
+
+    def draw(self, context):
+        layout = self.layout
+        game = context.active_object.game
+
+        row = layout.row()
+        row.prop(game, "physics_culling_radius", text="Physics Culling Radius")
+        row = layout.row()
+        row.prop(game, "logic_culling_radius", text="Logic Culling Radius")
+
 class OBJECT_PT_levels_of_detail(ObjectButtonsPanel, Panel):
     bl_label = "Levels of Detail"
     COMPAT_ENGINES = {'BLENDER_GAME'}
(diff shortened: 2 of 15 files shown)


youle31:
Voilà, I did my TODOs... I don't know if this can remain as simple as it is or if we need more advanced stuff.

youle31:
Oh I forgot to disable this for ImageRender

youle31:
I'm ok with your approach @panzergame (take the distance to the nearest camera to avoid to suspend physics or logic whereas the object can be in the field of an ImageRender camera for example).
About loop/iter with all objects in the scene, we could do that temporarly (unless we let it in the CalculateVisibleMeshes loop but iter with all cameras per object if physics or logic activity culling is enabled) but maybe it would be nice to try to group a maximum gameobj->Updates to avoid to have  many loops (Update LOD, Update Buckets, CalculateVisibleMeshes, Actions..... See what we can group to avoid multiples loops when it is possible)

source/gameengine/GameLogic/SCA_IObject.cpp
```diff
@@ -248,7 +248,7 @@ SCA_IActuator* SCA_IObject::FindActuator(const std::string& actuatorname)
 }
 
 
-void SCA_IObject::Suspend()
+void SCA_IObject::SuspendLogic()
```
panzergame: @youle31 : This function was used in the original UpdateObjectActivity to disable object logic update, i prefered to use instead of the layer shift.

source/gameengine/Ketsji/KX_Camera.cpp
```diff
@@ -56,6 +56,7 @@ KX_Camera::KX_Camera(void* sgReplicationInfo,
       m_set_projection_matrix(false),
       m_delete_node(delete_node),
 	  m_lodDistanceFactor(1.0f),
+	  m_activityCulling(false),
```
panzergame: @youle31 : There's three levels of enable option: scenes, cameras, objects. Do you agree with this ?

release/scripts/startup/bl_ui/properties_data_camera.py
```diff
@@ -151,8 +151,8 @@ def draw(self, context):
         col.prop(cam, "lod_factor", text="Distance Factor")
 
 
-class DATA_PT_frustum_culling(CameraButtonsPanel, Panel):
```
panzergame: I made the UI very quick, i certainly didn't choose the best presentation... @youle31 : Feel free to rework it.

source/blender/makesdna/DNA_object_types.h
```diff
@@ -112,6 +112,21 @@ typedef struct LodLevel {
...
+	float physicsRadius;
+	float logicRadius;
+
+	int flags;
+	int pad;
+} ObjectActivityCulling;
+
+/* object activity flags */
+enum {
+	OB_ACTIVITY_PHYSICS = (1 << 0),
```
panzergame: @youle31 : I'm using a flag to disable some culling without setting the radius to zero in case the user want to test the culling without loose his radius settings.

source/gameengine/GameLogic/SCA_IObject.cpp
youle31: ok, thanks for explaination.

source/gameengine/Ketsji/KX_Camera.cpp
youle31: yeah, this is cool

release/scripts/startup/bl_ui/properties_data_camera.py
youle31: I'll read it. if I have comment, I'll do it in the final review

source/blender/makesdna/DNA_object_types.h
youle31: ok, I understand now, good idea

panzergame:
@youle31 : Do you agree to open a pull request for ge_activityculling2 ?

youle31:
hello, of course, no need to ask to me for that :) I close this one

PR #468, opened 2017-04-23
Add option to override scene culling by a camera.

panzergame:
Idea from youle in ge_lightfrustum2.

The user can now debug the culling of a scene by using a culling camera
which have a transform different to the render camera.
To set a camera as culling override, the user have to check an option in
UI into the new panel named "Frustum Culling", or set in python the value
KX_Scene.overrideCullingCamera to the used camera for culling.

### DIFF
--- a/doc/python_api/rst/bge_types/bge.types.KX_Scene.rst
+++ b/doc/python_api/rst/bge_types/bge.types.KX_Scene.rst
@@ -89,6 +89,12 @@ base class --- :class:`PyObjectPlus`
          
          This can be set directly from python to avoid using the :class:`KX_SceneActuator`.
 
+   .. attribute:: overrideCullingCamera
+
+      The override camera used for scene culling, if set to None the culling is proceeded with the camera used to render.
+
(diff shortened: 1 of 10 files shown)


youle31:
Thanks @panzergame! This is cool and we can see that there are very annoying false positives with the bullet culling test. I think we should draw all the bounding boxes (even if their object is culled) and choose a color code for bounding boxes lines to indicate that when the bounding box is purple, its object is rendered, and when it is "another color", its object is not rendered. This would make the display more clear and we could see better the intersection beetween camera frustum and bounding boxes.

youle31:
e.g: http://pasteall.org/pic/show.php?id=114913 (image from laboulette branch)

panzergame:
Maybe adding an option to draw non-culled object is better in case where it
is unwanted and could spend a long time. Otherwise i agree to allow in the
absolute to draw these objects as it could be very useful for our works.

2017-04-23 18:47 GMT+02:00 youle31 <notifications@github.com>:

panzergame:
Merged in master.

PR #470, opened 2017-04-25
dirty dirt patch to restore gemat texts (just proposal... to discuss)

youle31:
test file: http://pasteall.org/blend/index.php?id=46348


youle31:
As we'll work on merge with 2.8 and that it is not a prior feature, I close this pull request

PR #474, opened 2017-04-27
UPBGE: Add an API to control environment light energy

youle31:
test file: http://pasteall.org/blend/index.php?id=46386


source/gameengine/Ketsji/KX_WorldInfo.cpp
```diff
@@ -177,6 +178,7 @@ void KX_WorldInfo::UpdateBackGround(RAS_Rasterizer *rasty)
 		GPU_horizon_update_color(m_horizoncolor.getValue());
 		GPU_zenith_update_color(m_zenithcolor.getValue());
 		GPU_update_exposure_range(m_exposure, m_range);
```
panzergame: Indeed you can move the exposure+range+envlight update only in UpdateWorldSettings

youle31:
Thanks @panzergame . I did what you suggested so.

youle31:
commited in https://github.com/UPBGE/blender/commit/db1da8d921ae5666a5f33f95f44c896c2512846f

PR #487, opened 2017-05-07
Revert exposure and range dynamic uniforms to test performances

youle31:
This branch is just to point a performances problem with this commit: https://github.com/UPBGE/blender/commit/887962d5b6d6cb5440a9c91721239bcf9095412e

Test file: http://pasteall.org/blend/index.php?id=46519

NOTE: ALL constants are checked.

@Akira1San && @panzergame: Can you confirm that you have a big performance increase when you revert this commit with this test file?

On my computer I have 415 fps in master and 450 in this branch.


Akira1San:
Yey, this revert fixes the realism file fps, img: http://pasteall.org/pic/show.php?id=115308

PR #490, opened 2017-05-08
UPBGE: Change of method to Set world exposure and range for world

youle31:
The previous method had the problem in some cases, whereas constant world
was checked, to decrease performances.

This commit fixes that problem.

Note: When world is not checked, this can cost a lot of performances. So
I'm wondering if it won't be preferable to totally remove this API

test file: http://pasteall.org/blend/index.php?id=46527

2 scripts (1 for constants and 1 to test set exposure and range)


youle31:
not very convincing

PR #491, opened 2017-05-11
UPBGE: Add read only API for use envlight and envlight color

youle31:
This answers request: https://github.com/UPBGE/blender/issues/452

test file: http://pasteall.org/blend/index.php?id=46581

### DIFF
--- a/doc/python_api/rst/bge_types/bge.types.KX_WorldInfo.rst
+++ b/doc/python_api/rst/bge_types/bge.types.KX_WorldInfo.rst
@@ -102,3 +102,16 @@ base class --- :class:`PyObjectPlus`
 
       :type: float from 0.0 to infinite
 
+   .. attribute:: envLightEnabled
+
+      Returns True if Environment Lighting is enabled. Else returns False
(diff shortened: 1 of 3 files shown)


doc/python_api/rst/bge_types/bge.types.KX_WorldInfo.rst
```diff
@@ -102,3 +102,16 @@ base class --- :class:`PyObjectPlus`
 
       :type: float from 0.0 to infinite
 
+   .. attribute:: envLightEnable
```
panzergame: Can we do something similar for `mistEnable`. Also `enable` or `enabled` ?

youle31: There is already an RW api for "mistEnable".
I thought there was enough RW and as Maujoe asked for RO, I left in RO.

About the name, I already said that I prefered mistEnabled, envLightEnabled, but it has been decided that it will be enable without "d",  so I did like that.

PR #498, opened 2017-05-28
Object activity culling.

panzergame:


youle31:
http://pasteall.org/blend/index.php?id=46750 (I don't know if there were errors during commit but this doesn't work as excpected for me)
- Scene Activity culling option should be enabled by default and same for camera else the feature is too hard to use.
- Camera should not have activity culling option in object tab (else this makes the feature too hard to understand)
- bug with constraints
- Logic should disable animations but animations should not disable logic...

panzergame:
`Scene Activity culling option should be enabled by default and same for camera else the feature is too hard to use.`

Object activity culling requires performance when computing distance to camera. Also allowing the user to disable the culling per scene and control per camera can be useful.

Anyway i understand your request, what i propose is to set the option enable by default in UI and in sources to optimize distance computation. Do you agree ?

`bug with constraints`

This bug is shared with the usage of suspend/restorePhysics, it will be treated as regular bug, i have a workaround branch for the moment.

`Logic should disable animations but animations should not disable logic...`

This topic is hard; Logic action actuator is launching an animation so we could said that once the animation is launched the actuator should not do anything else repeating or ping pong on negative event.

On the other hand what the user except for actuators triggered by controller and sensors of an other active object ? Should the actuator do nothing ?

Here the action actuator defines a special behaviour when it has no link. Noticing that actually if the action actuator is linked to a controller of an other active object the animation isn't stopped.

I'm not yet sure about this subject.

panzergame:
Object activity culling is a big feature and very complex, @youle31, @DCubix, @lordloki : Do you agree to merge it in a week at the beginning of the next release if it is ready ?

youle31:
1) yes, I agree
2) waiting for suspend/restore constraints, we could only avoid to remove constraints in RemovePhysicsController like I did in my activityCulling branch
3) The problem I encountered was in the UI (When I checked animation or logic both were enabled/disabled). About animation culling, in my branch logic culling did it. I mean when you SetState the animations are suspended and the last played frame is kept waiting the animation is restored. In fact, I'm wondering, for more simplicity, if we shouldn't just remove the culling code from animation update (because we had to much problems since a long time with animation and culling and let the users do their own culling with logic culling option (because I don't see when the user would want to suspend only animations but not logic and if he wants to do that the api and the bricks allow that)
4) yes I agree to delay to 0.1.9

panzergame:
1. Done
2. Done in 0.1.8 master

panzergame:
Up.

youle31:
Hello, I did my review with this file: http://pasteall.org/blend/index.php?id=47691
It acts similarly than my old patch. I'm ok with it. I added 2 comments. Thanks for your work

youle31:
about camera culling option, this seems good in fact but it's really hard to understand like in this file: http://pasteall.org/blend/index.php?id=47701

youle31:
I'm ok with the patch as it is because cam option is enabled by default even if i don't really see in what case it can be used. So all is right for me

PR #499, opened 2017-05-30
Fix constraint when suspending physics.

panzergame:
Previously when we was suspending physics of an object using constraints, they were removed. This behaviour could be needed in some rare case but for the most case where the user just want to optimize physics computation the engine should keep the constraints.

To do so the constraints have to be removed from the dynamic world but not removed from the controller  or deleted in RemoveCcdPhysicsController. In the same time AddCcdPhysicsController iterate over all the constraint in the activated controller and add they in the dynamic world.

Unfortunately there's some limitation, first we could not do redundant operation in removing and adding of the constraint in the world, second we have to keep the disable collision between bodies flag used when adding a constraint into world.
To solve both of these issues the user pointer of the constraint are used to store a new type: CcdConstraint (in CcdPhysicsController). This new type only store the disable collision flag and a boolean named m_valid set to true when the constraint is added in world else to false when removed.

Two functions are added to clean the code: RemoveConstraint and RestoreConstraint.

@youle31, @DCubix, @lordloki : Could you review the patch please ?

Example file: http://pasteall.org/blend/index.php?id=46774
Press Space to apply torque, Q to suspend physics of a cube, E to restore.

### DIFF
--- a/doc/python_api/rst/bge_types/bge.types.KX_GameObject.rst
+++ b/doc/python_api/rst/bge_types/bge.types.KX_GameObject.rst
@@ -741,9 +741,13 @@ base class --- :class:`SCA_IObject`
       :arg angular_damping: Angular ("rotational") damping factor.
       :type angular_damping: float ∈ [0, 1]
 
-   .. method:: suspendPhysics()
+   .. method:: suspendPhysics([freeConstraints])
 
       Suspends physics for this object.
+
+      :arg freeConstraints: When set to `True` physics constraints used by the object are deleted.
+          Else when `False` (the default) constraints are restored when restoring physics.
+      :type freeConstraints: bool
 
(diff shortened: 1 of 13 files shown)


youle31:
Hi, I did a test file: http://pasteall.org/blend/index.php?id=46775
I noticed 2 things:
- I think constraints shouldn't be removed by default as you say (I think you inverted defaut boolean)
- The python API is not edited.

Else this worked for me, I don't review the code because this looks too advanced to me and I'm a bit tired.

youle31:
Hello panzergame! I think you forgot to invert default values for actuators too.

panzergame:
Merged in 00836b4e3454f45e34decbaed00f7cde4445ec1e.

PR #534, opened 2017-07-13
Fixed child scaling bug

sdfgeoff:
This bug:
https://developer.blender.org/T50806

This patch:
https://developer.blender.org/P444

Did some testing and it worked out fine in my test cases.


sdfgeoff:
Actually, see: 
https://github.com/UPBGE/blender/tree/ge_parent_scale

Which needs testing.

PR #551, opened 2017-08-04
Ge hardware skinning

lordloki:
This pull request is to have a general vision about changes. I will make the development in other branch


lordloki:
The current method (don't work yet) has a limitation due to max number of uniforms (this imply a max number of bones). After it works we can investigate in using TBO to remove this limitation.

PR #581, opened 2017-08-29
Implement custom camera viewport in UI.

panzergame:
Previously the camera were able to use custom viewport only
in python.
An UI is created to do the same based on entire viewport
ratios. The UI expose an option named "Custom Viewport"
and four ratio for left, right, bottom and top. These ratio
correspond with the width or height of the canvas multiplied
by the ratio.


youle31:
Oh I didn't thought it was for that. I tried here: http://pasteall.org/blend/index.php?id=47696 but I didn't really understood the principle. I have black screen with 1.0 values for ratio. I'll retry later and read the commit message again.

(I thought it was to do a render with a rendercam in overlay of the main camera (indicating the position and the scale of the window... Like to display kindof UI at some position in overlay)

PR #582, opened 2017-08-29
Implement SCA_InputEvent attribute helpers.

panzergame:
Previously the user needed to read the values or queue list
in an input to know if it was up, down, pressed or released.
To simplify this and avoid the lookup in list from python
attributes are added in SCA_InputEvent to do the same, they
are:
up
down
pressed
released

Note that when they are the status of the input from the last
frame so when an input is released then it is both up and down.

### DIFF
--- a/doc/python_api/rst/bge_types/bge.types.SCA_InputEvent.rst
+++ b/doc/python_api/rst/bge_types/bge.types.SCA_InputEvent.rst
@@ -46,6 +46,30 @@ base class --- :class:`PyObjectPlus`
 
       :type: list of integer.
 
+   .. attribute:: up
+
+      True if the input was up from the last frame.
+
+      :type: boolean
+
+   .. attribute:: down
+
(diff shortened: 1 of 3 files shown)


youle31:
For me it's ok. But I find the names confusing.
My test files: http://pasteall.org/blend/index.php?id=47695
Why confusing: for example, if you write:
if keyboard.inputs[bge.events.UPARROWKEY].up: go forward, then the object will go forward without pressing any key.
I'd keep the same names as before:
if keyboard.inputs[bge.events.UPARROWKEY].none:
if keyboard.inputs[bge.events.UPARROWKEY].active:
if keyboard.inputs[bge.events.UPARROWKEY].activated:
if keyboard.inputs[bge.events.UPARROWKEY].released:

Else this is very handy with keyboard.inputs (I still tend to use keyboard.events[bge.events.UPARROWKEY] == 2 to make my player go forward but I have to abandon this habit :P)

PR #595, opened 2017-09-14
Corrected reversed values.

TryExceptElse:
Previously, input was displayed after 'expected' and expected value was displayed after 'given'


youle31:
fixed in https://github.com/UPBGE/blender/commit/c8b5bb4fff4b392a143a13b4f077bbd2f0f8f8d3
Thanks @TryExceptElse

PR #611, opened 2017-10-10
Improve vertex memory management

panzergame:
Previously the vertex were using a common base class named RAS_IVertex which owned basic vertex data as position, normal, tangent. This base class also defined function to mutate these value and virtual functions for value in subclasses. The subclasses were RAS_Vertex, a template class, which owned the uv and color data with size depending on the template paramaters. This class also defined the virtual function to access uv or color.

This conception is correct as it allows using small vertex memory if the user doesn't need a lot of uv or color layers. But the usage of virtual function imposed the class to store a virtual ptr table, this table is meanly a pointer (8o), this pointer was send to VBO and used memory for nothing from OpenGL side.

To avoid this vptr, the vertex class is split in two classes: RAS_VertexData and RAS_Vertex.
The first class is only vertex data without function. It use template parameters for uvs and colors and is based upon RAS_VertexDataBasic for the position, color, normal data.
The second class is containing the functions and used only to mutate the vertex data. When accessing a vertex a RAS_Vertex instance is created with as parameters the vertex data pointer and the vertex format. This vertex format is used in uv and color getter to compute the return value based on the vertex data pointer plus a gap corresponding to the looked layer.

@youle31, @DCubix, @lordloki : I would like you to review this request please ? Feel free to ask, I'm not the best for explicit description.


youle31:
@panzergame: Hello! 

In this branch all my usual performances test files doesn't work because of a bug we had but which has been fixed later.

To build I had to comment //BLI_assert(shapeProps); in BL_BlenderDataConversion line 663.

I'm wondering if this branch + restoring the system we had before to sort polygons when alpha sort is enabled could solve this bug we had with Akira_san: https://github.com/UPBGE/blender/issues/430
(remove glMapBufferRange/unmap and restore).

Else with the files I tested it seems to work fine but didn't bring performance increase.

PR #695, opened 2018-05-25
Add component script generation.

panzergame:
To ease the definition of a component a button named "Create Component" is added.
This button ask for a path module.Class and generates a script named module.py
with a component class Class inside.
This template class contains a start and update function commented
and a definition block before the class for global game engine variable.

If the file already exists an error is raised.

The previous button named Add Component is now renamed Register Component.

### DIFF
--- a/release/scripts/templates_py/python_component.py
+++ b/release/scripts/templates_py/python_component.py
@@ -2,8 +2,8 @@ import bge
 from collections import OrderedDict
 
 if not hasattr(bge, "__component__"):
-	# Put shared definitions here executed only in game engine.
-	# e.g:
+    # Put shared definitions here executed only in game engine.
+    # e.g:
     # scene = bge.logic.getCurrentScene()
     pass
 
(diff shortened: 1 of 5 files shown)


release/scripts/templates_py/python_component.py
```diff
@@ -0,0 +1,24 @@
+import bge
+from collections import OrderedDict
+
+if not hasattr(bge, "__component__"):
+	# Put shared definitions here executed only in game engine.
```
panzergame: Convert tab to spaces.

panzergame:
Template :

```python
import bge
from collections import OrderedDict

if not hasattr(bge, "__component__"):
    # Put shared definitions here executed only in game engine.
    # e.g:
    # scene = bge.logic.getCurrentScene()
    pass

class Component(bge.types.KX_PythonComponent):
    # Put your arguments here of the format ("key", default_value).
    # These values are exposed to the UI.
    args = OrderedDict([
    ])

    def start(self, args):
        # Put your initialization code here, args stores the values from the UI.
        # self.object is the owner object of this component.
        pass

    def update(self):
        # Put your code executed every logic step here.
        # self.object is the owner object of this component.
        pass
```

PR #700, opened 2018-05-27
UPBGE: Fix ray cast with xray and property filtering.

### LINKED ISSUE #671: rayCast property not working
The propertie argument of the rayCast isn't working on 0.2.3 anymore... If you try to rayCast to a certain prop, it will return an object even if there is no props on it...

```
from bge import logic

def a(cont):
    obHit,_,_ = cont.owner.rayCast([0,0,0],[0,10,0],0.0,"test",1,1,0)
    
    if obHit != None:
        print(obHit)
```

If you run this code and put an object without the "test" prop, it will print them. :/

panzergame:
A check of prop.size() == 0 was rewritten in prop.size() instead of prop.empty().

In the same time the function CheckRayCastObject is implemented to factorise the
proprety and mask check made in RayHit and NeedRayCast.
CheckRayCastObject also add a little optimization by not getting the object collision
group when the mask accept all objects.

Fix #671.

### DIFF
--- a/source/gameengine/Ketsji/KX_GameObject.cpp
+++ b/source/gameengine/Ketsji/KX_GameObject.cpp
@@ -3923,16 +3923,22 @@ struct KX_GameObject::RayCastData {
 	KX_GameObject *m_hitObject;
 };
 
+static bool CheckRayCastObject(KX_GameObject *obj, KX_GameObject::RayCastData *rayData)
+{
+	const std::string& prop = rayData->m_prop;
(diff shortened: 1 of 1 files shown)


panzergame:
Unit test file : http://pasteall.org/blend/index.php?id=49593

UnidayStudio:
I don't know if I don't understand exactly how git works and pull the wrong "master" branch, but I get this assert trying the test file:
http://prntscr.com/k0bkvl
(Master branch pulled today. I guess)

PR #701, opened 2018-05-28
 UPBGE: Improve frustum culling performance.

panzergame:
Frustum culling was improved in two ways. First, simplificate the bounding
box update. Second the test are parallelized using TBB range loop.

The bounding box update is simplified by adding the call to RAS_Deformer::UpdateBuckets
in KX_GameObject::UpdateBounds after we checked that the objects as a
bounding box. By doing this we avoid a call to GetDeformer for all objects
without a bounding box, with an unmodified bounding box or without auto
update.

All the computation of culling objects is moved into KX_CullingHandler,
this class construct it's own objects list and returns it in Process function.
Process function build a CullTask and launch it usign tbb::parallal_reduce.

Each CullTask have an operator() to test a range of object, any objects passing
the culling test is added in a task local objects list. Once the tests
finished the CullTask merge these objects list in function join to end
up with the list of all non-culled objects.
This technique of reduce of list is way better than using a shared object
list for all tasks and lock a mutex before adding an object. The method
with mutex was always slower than without parallelization.

This patch was tested with cube meshes :
number of object | previous time | new time
1000 | 0.06 | 0.07
8000 | 1.04 | 0.55
27000 | 3.81 | 1.90
125000 | 16.16 | 8.31

### DIFF
--- a/build_files/cmake/macros.cmake
+++ b/build_files/cmake/macros.cmake
@@ -518,6 +518,8 @@ function(setup_liblinks
 
 	#target_link_libraries(${target} ${PLATFORM_LINKLIBS} ${CMAKE_DL_LIBS})
 	target_link_libraries(${target} ${PLATFORM_LINKLIBS})
+
+	target_link_libraries(${target} ${TBB_LIBRARIES})
 endfunction()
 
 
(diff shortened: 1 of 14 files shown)


lordloki:
Sorry, i'm trying to test this pull request but I'm not able to compile (tbb/tbb.h not found).
It is a weird error as tbb include as tbb lib are set correctly. I will try to find a solution

panzergame:
If your on linux:
```
TBB_INCLUDE_DIR /usr/include
TBB_LIBRARY /usr/lib/libtbb.so
```
The include dir is the parent directory of tbb/tbb.h

lordloki:
Ok, I think I know that what is it the issue. 
In Windows TBB_INCLUDE_DIR and TBB_LIBRARY are only set if you configure CMAKE with "WITH_OPENVDB" set to ON. I will check it later and If it works I will upload a fix.

PR #708, opened 2018-06-02
UPBGE: Move SG_Controller entirely in BL_Action.

panzergame:
Previously SG_Controller were owned and updated by SG_Node. This class
is not the best place for it as SG_Controller are not always modifying spatial,
for example Light or Camera controllers.
The best place is in BL_Action as this class was setting the simulation time
and updating the controllers indirectly.

In the same time the option "child" in action actuator which enabled recursive
time update for controllers is removed. Indeed every call to SetSimulatedTime
was followed by Update for the controllers in BL_Action.

### DIFF
--- a/source/blender/editors/space_logic/logic_window.c
+++ b/source/blender/editors/space_logic/logic_window.c
@@ -1415,8 +1415,6 @@ static void draw_actuator_action(uiLayout *layout, PointerRNA *ptr)
 		uiItemR(row, ptr, "frame_end", 0, NULL, ICON_NONE);
 	}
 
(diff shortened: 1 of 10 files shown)


panzergame:
@lordloki : If you have time to make a quick review please, i would like to merge this patch as some branches are based on.

lordloki:
@panzergame : i did a visual review and everything seemed ok to me. Regarding windows compiling and testing, currently i'm at a holiday family trip and i won't get access to my computer until 12th July. Sorry

panzergame:
Thank you for the review, if someone report a windows build error i will try to fix it. Anyway have a good holiday :P

PR #709, opened 2018-06-03
Use mt::vec instead of float array if possible.

panzergame:


lordloki:
Compiled OK. I double-checked all the conversions and it seemed OK to me. Good work

PR #717, opened 2018-06-15
UPBGE: Move RAS_Deformer in RAS_MeshUser.

panzergame:
Previously the deformers were owned in BL_DeformableGameObject. But this place
was not optimal for two reasons : first by design deformers are directly linked
to the display array of the obejct mesh slots, RAS_MeshUser is the nearest level
to the mesh slots of the object. Second a futur patch will allow fast mesh replace
by switching of mesh user, to do so the mesh user must include its deformer to avoid
recreating the current deformer for the current mesh user.

By moving the the deformer in RAS_MeshUser, only a last frame variable stays in
BL_DeformabelGameObject, this variable is then moved into BL_MeshDeformer and
BL_DeformableGameObject is removed.
Because this class is removed the function LoadDeformer is changed in BL_ConvertDeformer
which returns a deformer for the given mesh and object, could be null.
This last function is called in AddMeshUser and makes obsolete the replication
of deformers as it is called after any object creation.
The only drawback of avoid replication is to let all modifier deformers using
unique derived mesh, this impact only non-dynamic derived mesh.

Tested with sintel desert_level.


lordloki:
Compiled and tested with several animation files without any problems.
Refactor seems ok to me. Good work.

PR #721, opened 2018-06-25
UPBGE: Use bullet 3 and support multithreading physics.

panzergame:
The bullet library used by blender and game engine is update to the latest bullet 3
version. Only minor modification were requeired for the switch.

This bullet version allow multi threading for rigid body collision and constraints,
the main world for this prupose is btDiscretDynamicsWorldMt. Unfortunatly this
class doesn't manage soft bodies and to fill this gap btSoftRigidDynamicsWorldMt
which is a copy of btSoftRigidDynamics on top of btDiscretDynamicsWorldMt.
Once these modification the world is initialised with a special dispatcher and
a constraint solver pool for multi threading. The solver pool own as many regular
solver by threads available.

A compatiblity issue is introduced in KX_CharacterWrapper, bullet now supports vector
for the gravity instead of simple float, then a vector is exposed into the API.

Tests:
2700 spheres on two planes :
		Previous	Current
Begining	100ms		70ms
Stable	21ms		7.5ms

16000 point constraint on 10 cube chains.
		Previous	Current
		141ms		75ms


adriansnetlis:
According to your tests, constraint calculation speed is doubled while collision calculation is slower?

adriansnetlis:
Ah, so Bullet 3 provides significant performance improvements. I expect this means that physics sub-stepping can be pushed up for more stable rigid springs?

BluePrintRandom:
yeah! Thanks you so much guys! 🥇 

can we expose the new batched raycast api :3 

does this support threaded - rebuilding of triangle mesh bvhtree?

panzergame:
Is the compatibility issue on `KX_CharacterWrapper.gravity` fine ?

lordloki:
For me it is ok and open more possibilities

PR #730, opened 2018-07-03
UPBGE: Improve animation and deformer update performance.

panzergame:
Previously the animations were using parallelism, each step consisted in:
- Detect if the object is an armature
- In case of armature detect a child with satified condition to allow
	updating armature
- Update the animations through the action manager
- In case of armature update the deformers of all children

The deformers were all calling a lazy update of the armature thanks to
function ApplyPose.

This system is quite correct but has some limitation. First it exists a
lot of different callers for updating armature pose (ApplyPose), this
function is called from deformers or from bone parent relation which
doesn't benefit of parallelism.
Also each step were not of the same execution time depending if the
object was an aramture or not, with many children or none. The load
of threads could suffer of this irregularity.

The changes of this commit target these limitations.

Firstly the animations update is separated in three stages :
Actions, Armatures, Deformers. Each step is procceded with tbb using
parallel_for with a custom task class, each task class constructor aim
to list all objects concerned by the stage, e.g AnimationTask list all
animated objects with an action manager and a non suspended manager.

Actions

AnimationTask filters all the animated objects and keeps only the ones with
non-suspended action manager, then the parallel part for each object
checks if it is an armature, if yes all children are tested. Previously
the children were get from KX_GameObject::GetChildren but this functions
is quite expensive as it constructs recursively a new list, an iteration
over child scene graph nodes is preferred. Finally the action manager
is updated similary to before.
Inside BL_Action::Update to avoid substracting the current time by
the suspended time, the time sent to the action manager is now current
time minus the suspended time, also a boolean named redundant is
passed to the action manager to explicitly notify that the update
is redundant (used only for applying armatures on now visible armatures).
Function IsActionsSuspended and UpdateActionManager are not anymore
calling GetActionManager as the context ensure that the action manager
is existing (only object with an action manager are in the animated object
list), instead they directly use m_actionManager.

Armatures

First of all the scene now registers indepently active armatures in
m_armatureList in AddNodeReplica and from the converter via AddArmature.
ArmatureTask is filtering all the scene active armatures and test
for function NeedApplyPose which basically compare the last pose
update time with the last animation update time. Then the parallel
part calls ApplyPose.

Deformers

RAS_Deformer expose two fundamental functions, NeedUpdate and Update.
The first return a non zero flag when the deformer is tagged for update
and used by DeformerTask constructor to register a deformer. Its
implementation in soft body justs return always 1 and for skin/shape deformer
it checks for the previous armature pose update or for modification of
the original deformer display arrays and returns a combinaison of values
from BL_MeshDeformer::UpdateReason. The second function is used to
actually update the deformer with as argument the reson returned by
NeedUpdate to select what data needed to be udpated.
Most of the deformer are now using a separate loop to compute the AABB
to avoid branches and not waste time when the automatic update is
disabled.

Performance is pretty similar and this system is more flexible over
all.

This commit is quite large and not so simple, feel free to ask any questions and thanks for futurs reviewing :)


youle31:
Hello, I just bought a new CPU (i7 8700k received it this monday). This is new intel CPUs (8th generation with 6 cores), and it works very fast (I downclocked it for ecologic reasons :) ). Anyway, I'll do a comparison between master and ge_tbb_anim with my CPU (and my CPU settings) on Windows. Unfortunately I can't test with my old i7 2600k.

I'll try to get and share some test files for armatures and deformers... If you or @lordloki have test files, could you share it somewhere? Thanks 

I won't do the review (or maybe I can do a quick review, I'll see if I'm motivated) right now because I'm a bit busy but I'll try this month and I'll share my test files.

Thanks very much for your work!

youle31:
Hi, I pushed that: https://github.com/UPBGE/blender/commit/622ec8aa915096b55f70e2e3847b6d552c659566
to fix compilation on windows.

I just did a first test with Defender file:

Master:
master
ge_tbb_anim:
tbb_anim

lordloki:
@youle tbb library version is a bit old. I will update it we i can get access to my computer to test it

youle31:
hello, Thanks @lordloki .

I did another test with shape keys:
Master:
master
ge_tbb_anim_win:
ge_tbb_anim_win
File:
http://pasteall.org/blend/index.php?id=49864

(It's weird, in the viewport the shading seems flat and at runtime it seems smooth)

As for the first test, I have no noticeable performance difference in this file. Maybe the new tbb library will be more performant. I'll wait lordloki lib update for more tests.

panzergame:
TODO: currently the at each update the deformers are listed from the all objects because they can change easily at each mesh replace (LOD) but at the end only a certain type of objects could use deformers. The idea is to detect these objects and register them in the scene.

lordloki:
I have also less performance (around 10% in youle's both blends) but if you join shapekeys and armatures I have a gain around 3-4% (plus a more stable fps)

Youle, could you check this file? https://mega.nz/#!UlNhxQIb!9UcEpjoEaPxZs2cJIy2QrDCzVy-ebbV5quu9kKEezis

youle31:
@lordloki  && @panzergame : Yes, I confirm that I have a performance increase in lordloki file mixing both shapekeys and the armatured guys.

Master:
master
ge_tbb_anim:
ge_tbb_anim

@lordloki : I had a virus on my computer trying to activate windows 10 illegally whereas I have a legal windows 7 key which works on windows 10 lol (I didn't know we can do that). In the panic, I forgot to backup all my testing blends (both for upbge and eevee in bge :S). If you still have your mega folder with test files, could you give me the link? Thanks!

lordloki:
@youle31 : sorry to hear that. In the Mega repository I almost have nothing yet ([https://mega.nz/#F!NxFFASLL!R7UkM0CpJlL9_-XBPIPJSQ]). But there is a repository with a lot test files maintained by panzergame [https://github.com/UPBGE/testfiles]

youle31:
Thanks very much to you both :)

PR #733, opened 2018-07-05
UPBGE: Avoid calling glGetUniformLocation while setting a uniform through BL_Shader.

panzergame:
All the setUniform functions in BL_Shader are using name to identify the uniform,
these functions call RAS_Shader::GetUniformLocation.

This last function was naively calling indirectly glGetUniformLocation, but even
simple OpenGL calls cost more than caching.
Fortunatly OpenGL allows to enumerate all active uniforms thanks to glGetActiveUniform,
this function is used in GPUShader side inside function GPU_shader_get_uniform_infos
constructing a list of GPUUniformInfo storing the uniform location, size, type and
name.

These uniform info are used in RAS_Shader::ExtractUniformInfos which insert the
info in a map associating to the name the location, size and type> This map is then
used into GetUniformLocation.

Tested with a 4 uniforms shader and 1000 updates :
Previous : 3.26ms
Current : 2.88ms

Test file :
[ge_uniforms.zip](https://github.com/UPBGE/blender/files/2167666/ge_uniforms.zip)

### DIFF
--- a/source/blender/gpu/GPU_shader.h
+++ b/source/blender/gpu/GPU_shader.h
@@ -73,6 +73,16 @@ void GPU_shader_unbind(void);
 
 int GPU_shader_program(GPUShader *shader);
 
+typedef struct GPUUniformInfo
+{
+	unsigned int location;
+	unsigned int size;
+	unsigned int type;
+	char name[255];
+} GPUUniformInfo;
+
+int GPU_shader_get_uniform_infos(GPUShader *shader, GPUUniformInfo **infos);
+
 void *GPU_shader_get_interface(GPUShader *shader);
 void GPU_shader_set_interface(GPUShader *shader, void *interface);
 int GPU_shader_get_uniform(GPUShader *shader, const char *name);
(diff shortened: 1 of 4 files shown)


youle31:
Hi, I just tested your file disabling vsync and "use framerate" and enabled mouse cursor to launch my screen capture program:
Master:
master
ge_uniforms:
ge_uniforms
I don't really notice significant performance difference but I'd have to test with other files.
I can't for now but I'll try later.

panzergame:
I think some drivers are implementing similar optimizations, in my case I tested with a intel GPU.
Anyway thanks for the tests you made for the pull requests :)

panzergame:
Note that ExtractUniformInfos could also be used to detect the usage of `gl_LightSource` or any GLSL material builtins to disable the update of these builtins.

youle31:
Yes, it could be nice to get only active uniforms to see if we can do optimizations, not necessarly in GLSL but maybe elsewhere? I don't really know. I used it only one time (glGetActiveUniforms) but I can't remember for what purpose.

youle31:
If both of you notice performance increase, this is great :) uniform caching can't be a bad thing anyway

PR #734, opened 2018-07-06
UPBGE: Implement dynamic object ligthing layer.

panzergame:
Previously the ligthing of the objects based on a layer was fixed
at the conversion time by the construction of BL_BlenderShader
with the member m_lightLayer. Also because of the blender
material managment the object layer was used only when binding
the material and not when binding object dependent uniforms.

The situation is changed firstly because the usage of one
layer for all the object using the same material is totally wrong.
Secondly the layers were not dynamic per object.

To solve this situation the fastest way is to expose a layer into
the shader, the function lamp_visible is designed for. It receives
the lamp layer, object layer, lamp color and lamp energy and
updates out color and energy if the lamp layer mask the object
layer, to avoid branching the test computes a value of 0 or 1
and multiply the color and energy by it.

The object layer is defined as a material builtin uniform and
each GPULamp holds a dynlayer used as a dynamic uniform.
At each call to GPU_material_bind_uniforms the object layer
is updated. Concerning the lamps the dynlayer value is updated
in a preparing stage before rendering any material, this stage
intents to update the textures and the lamps. The main caller
is RAS_BucketManager::PrepareBuckets calling for each material
the Prepare function redirected to BL_BlenderShader::UpdateLights.
This last function is calling GPU_material_update_lamps which
over all lamps update the "dyn" value for color and matrices.

Inside GPU_material_bind the lamp layer is computed. As it is
not the only way to decide of the visibility of a lamp, if the
lamp is hidden or the layer is set to 0, else if the lamp is not
using a "layer only" option then all layers are enabled, else if
the layer is not in the scene layer the layer is set to 0
else to the actual lamp object layer.

As GPU_material_bind_uniforms now use the object layer,
RAS_MeshUser store the layer of the object to allow BL_BlenderShader
to get this value and send it to the GPU_ function.
In the same time the calls to SetFrontFace are reduced to
only updating when the object node transform changed.

No performance win neither loss were noticed.

Test file, the two cube are from different layer and the lamp is switching of layer:
[ge_lamp_layer.zip](https://github.com/UPBGE/blender/files/2171721/ge_lamp_layer.zip)

### DIFF
--- a/source/blender/editors/space_view3d/view3d_draw.c
+++ b/source/blender/editors/space_view3d/view3d_draw.c
@@ -3107,7 +3107,7 @@ static void view3d_main_region_clear(Scene *scene, View3D *v3d, ARegion *ar)
 		GPUMaterial *gpumat = GPU_material_world(scene, scene->world);
 
 		/* calculate full shader for background */
-		GPU_material_bind(gpumat, 1, 1, 1.0, true, rv3d->viewmat, rv3d->viewinv, rv3d->viewcamtexcofac, (v3d->scenelock != 0));
+		GPU_material_bind(gpumat, 1, 1.0, true, rv3d->viewmat, rv3d->viewinv, rv3d->viewcamtexcofac, (v3d->scenelock != 0));
 
 		bool material_not_bound = !GPU_material_bound(gpumat);
 
--- a/source/blender/gpu/GPU_material.h
+++ b/source/blender/gpu/GPU_material.h
@@ -76,6 +76,8 @@ typedef enum GPUType {
 	GPU_MAT3 = 9,
 	GPU_MAT4 = 16,
 
+	GPU_INT = 17,
+
 	GPU_TEX2D = 1002,
 	GPU_SHADOW2D = 1003,
 	GPU_TEXCUBE = 1004,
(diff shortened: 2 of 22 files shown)


youle31:
Hi, this is working for me. I didn't looked at the code.

- What is cool is that it offers a new option for light layers
- But I think it should bring performance improvement. The shader can't be stopped at early stages in GPU_shader_material.glsl to say if (inactiveLayer || ...) return; ? Or maybe you didn't noticed performances because the test file was too simple? It might be interesting to test this in complex scenes (https://www.blendswap.com/blends/view/88644 , https://www.blendswap.com/blends/view/89039 (I choosed black and white scene because I like it. I played a bit with the second one when I was testing eevee but I don't really know if it renders nice in BI/BGE)

panzergame:
@youle31 : Indeed a check could be better, but there's no way currently to do conditional nodes in materials. So the stop is just setting a 0 intensity. I tested with the two files, but as they have no material a few material switch happen (just the default material bind/unbind).

youle31:
@panzergame: ok, thanks for your explaination.

parenthesis: I don't want to do true "reviews" as I'm not really qualified to check your work. You can add me as a reviewer if you want, but I think most of time I'll just give an opinion or general reflexions or test files feedback, then don't wait for my review approval to commit/merge your work :)

panzergame:
@youle31 : This is already great :)

PR #736, opened 2018-07-08
UPBGE: Implement BVH generation from KX_Mesh.

panzergame:
The python function KX_Mesh.constructBvh now returns a new BVH Tree
based on the mesh geometry where every indices elements of the tree
correspond to the polygons indices.

Test file:
[ge_bvhtree.zip](https://github.com/UPBGE/blender/files/2173760/ge_bvhtree.zip)

### DIFF
--- a/doc/python_api/rst/bge_types/bge.types.KX_Mesh.rst
+++ b/doc/python_api/rst/bge_types/bge.types.KX_Mesh.rst
@@ -154,3 +154,10 @@ base class --- :class:`EXP_Value`
 
       :return: a duplicated mesh of the current used.
       :rtype: :class:`KX_Mesh`.
+
+   .. method:: constructBvh()
+
+      Return a BVH tree based on mesh geometry. Indices of tree elements match polygons indices.
+
+      :return: A BVH tree based on mesh geometry.
+      :rtype: :class:`mathutils.bvhtree.BVHTree`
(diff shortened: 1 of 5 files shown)


youle31:
Hi, interesting. I didn't looked at the code but it seems to work fine for me (a poly seems to be selected with bvh.ray_cast, then each vertex of the poly is moved. This is only the rendered mesh (RAS_MeshObject) which is transformed? The physics is updated according to the rendered mesh? Can we use ReinstancePhysics to recalculate physics?).

Remark: Else, when I worked on eevee, I tested to move MVerts (MVert.co iirc) and it has not the same behaviour than in BGE: The polygons are "stick" together so when you move a MVert (with depsgraph), all the polygons linked to this MVert are recalculated too (like when you select a vertex in edit mode and move it) (For info purpose and if you want to try with KX_GameObject::m_pBlenderObject, the notifier to update mesh is DEG_id_tag_update(&ob->id, OB_RECALC_DATA) then BKE_scene_update_tagged(KX_Scene->GetBlenderScene().....) to flush the tagged updates (I don't know if the names are the same in 2.7 but it might be the same principle) (I think this is slower than in BGE but it is cool too)

panzergame:
Yes the mesh can be reused for physics reinstancing, as other KX_Mesh.

PR #745, opened 2018-07-14
UPBGE: Add `onRemove` callbacks

marechal-p:
Register functions into a callback list on scenes.

```py
def example(controller):
	scene = controller.owner.scene

	@scene.onRemove.append
	def callback(scene):
		print('exiting %s...' % scene.name)
```

---

Before this there was no event to listen for the engine shutdown. I would use the `weakref` module and monitor the `bge.logic` object destruction to trigger code.

This commit adds a list of callbacks on scenes to listen for engine shutdown.

### DIFF
--- a/source/gameengine/Ketsji/KX_KetsjiEngine.cpp
+++ b/source/gameengine/Ketsji/KX_KetsjiEngine.cpp
@@ -1054,7 +1054,7 @@ void KX_KetsjiEngine::StopEngine()
 
 		while (m_scenes->GetCount() > 0) {
 			KX_Scene *scene = m_scenes->GetFront();
-			m_converter->RemoveScene(scene);
+			DestructScene(scene);
 			// WARNING: here the scene is a dangling pointer.
 			m_scenes->Remove(0);
 		}
@@ -1296,7 +1296,7 @@ void KX_KetsjiEngine::RemoveScheduledScenes()
 
 			KX_Scene *scene = FindScene(scenename);
 			if (scene) {
-				m_converter->RemoveScene(scene);
+				DestructScene(scene);
 				m_scenes->RemoveValue(scene);
 			}
 		}
@@ -1408,7 +1408,8 @@ void KX_KetsjiEngine::ReplaceScheduledScenes()
 					// avoid crash if the new scene doesn't exist, just do nothing
 					Scene *blScene = m_converter->GetBlenderSceneForName(newscenename);
 					if (blScene) {
-						m_converter->RemoveScene(scene);
+						DestructScene(scene);
+						m_scenes->RemoveValue(scene);
 
 						KX_Scene *tmpscene = CreateScene(blScene);
 						ConvertScene(tmpscene);
@@ -1441,6 +1442,12 @@ void KX_KetsjiEngine::ResumeScene(const std::string& scenename)
 	if (scene) {
 		scene->Resume();
 	}
+}
+
+void KX_KetsjiEngine::DestructScene(KX_Scene *scene)
+{
+	scene->RunOnRemoveCallbacks();
+	m_converter->RemoveScene(scene);
 }
 
 double KX_KetsjiEngine::GetTicRate()
--- a/source/gameengine/Ketsji/KX_Scene.cpp
+++ b/source/gameengine/Ketsji/KX_Scene.cpp
@@ -196,6 +196,7 @@ KX_Scene::KX_Scene(SCA_IInputDevice *inputDevice,
 
 #ifdef WITH_PYTHON
 	m_attrDict = nullptr;
+	m_removeCallbacks = nullptr;
 
 	for (unsigned short i = 0; i < MAX_DRAW_CALLBACK; ++i) {
 		m_drawCallbacks[i] = nullptr;
@@ -283,6 +284,7 @@ KX_Scene::~KX_Scene()
 	}
 
 	// These may be nullptr but the macro checks.
+	Py_CLEAR(m_removeCallbacks);
 	for (unsigned short i = 0; i < MAX_DRAW_CALLBACK; ++i) {
 		Py_CLEAR(m_drawCallbacks[i]);
 	}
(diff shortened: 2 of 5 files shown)


marechal-p:
Would it be better to have some kind of "general" listener list such as `bge.logic.on_exit` independent from the scenes ?

source/gameengine/Ketsji/KX_Scene.cpp
```diff
@@ -1731,6 +1733,19 @@ void KX_Scene::RunDrawingCallbacks(DrawingCallbackType callbackType, KX_Camera *
 	}
 }
 
+void KX_Scene::RunExitCallbacks()
+{
+	PyObject *self = this->GetProxy();
```
panzergame: We generally avoid using `this->`in the game engine sources.

source/gameengine/Ketsji/KX_KetsjiEngine.cpp
```diff
@@ -1054,6 +1054,7 @@ void KX_KetsjiEngine::StopEngine()
 
 		while (m_scenes->GetCount() > 0) {
 			KX_Scene *scene = m_scenes->GetFront();
+			scene->RunExitCallbacks();
```
panzergame: You can call the same function inside `RemoveScheduledScenes` when a scene is removed separately.

source/gameengine/Ketsji/KX_Scene.cpp
```diff
@@ -2090,6 +2135,7 @@ PyAttributeDef KX_Scene::Attributes[] = {
 	EXP_PYATTRIBUTE_RW_FUNCTION("pre_draw", KX_Scene, pyattr_get_drawing_callback, pyattr_set_drawing_callback),
 	EXP_PYATTRIBUTE_RW_FUNCTION("post_draw", KX_Scene, pyattr_get_drawing_callback, pyattr_set_drawing_callback),
 	EXP_PYATTRIBUTE_RW_FUNCTION("pre_draw_setup", KX_Scene, pyattr_get_drawing_callback, pyattr_set_drawing_callback),
+	EXP_PYATTRIBUTE_RW_FUNCTION("on_exit", KX_Scene, pyattr_get_exit_callback, pyattr_set_exit_callback),
```
panzergame: I know that the other callbacks are named using `_`, but this will change (see the issue related to API renaming), so I think you should use camelCase : `onExit`.

source/gameengine/Ketsji/KX_Scene.cpp
panzergame: Just use :
`PyObject *self = GetProxy()`

Also as you are using this value once you could call the function directly in `args` constructor :
`PyObject *args[1] = {GetProxy()};`

source/gameengine/Ketsji/KX_KetsjiEngine.cpp
panzergame: I think that per scene is more interesting, but i would like to hear opinion from @lordloki, @youle31  and other.

source/gameengine/Ketsji/KX_Scene.cpp
marechal-p: To be fair, I am kind of mixed here, but let's go with camelCase.

source/gameengine/Ketsji/KX_KetsjiEngine.cpp
marechal-p: Is it silly to trigger the callbacks inside `BL_Converter::RemoveScene` ?

marechal-p:
Erreur de manipulation :/

Anyway, I think I checked in some junk, should I remove it or should it stay in the PR ?
(the submodules and release files ?)

panzergame:
`Anyway, I think I checked in some junk, should I remove it or should it stay in the PR ?
(the submodules and release files ?)` You should not track these submodule, but anyway i can handle this before the merge.

source/gameengine/Ketsji/KX_KetsjiEngine.cpp
panzergame: We are going to duplicate code by calling RunExitCallbacks and it's why you want to move it in common function BL_Converter::RemoveScene, but this class is not really the best place.
I would prefer use a function like `KX_KetsjiEngine::DestructScene`doing the remove from the converter and the callback but not the remove from m_scenes.

panzergame:
I will be on IRC tonight if you have some questions.

youle31:
hey guys, nice to see you working together :) I never dealt with callbacks before, then I think panzergame or lordloki are more qualified for review

just general remarks:

- If I had to do this patch, I'll try to take example on other callbacks in upbge code. iirc, there is an enum for scene callbacks somewhere looking like CALLBACK_PRE_SCENE. Maybe this should be used (I haven't my IDE opened but this is just I think I saw something like that before)
- For symmetry, what about on_start_callbacks (But this can be done in another patch)
- I don't know if this is good, but since a moment, I prefer to group this kind of things in KX_Scene destructor/constructor (but this is only my own preference and I don't know if there are conventions about that). This has to be discussed with Tristan.

Bonne soirée!

marechal-p:
@youle31:

The scope of this PR was originally to detect the engine shutdown, but now it is a callback for the scene removal. But the draw enum was specific to the drawing, hence why I made the new list of callbacks.

On the other hand, I think we should have hooks everywhere. `onAddObject`, `onAddScene`, `onRemoveObject`, etc...

panzergame:
Other than KX_KetsjiEngine::DestructScene i'm ok.

marechal-p:
@panzergame I am sorry what do you mean ?

panzergame:
I mean adding a function in `KX_KetsjiEngine`which is called only when a scene is removed by game end or by the user directly:

```
void KX_KetsjiEngine::DestructScene(KX_Scene *scene)
{
    // Call the callbacks
    m_converter->RemoveScene(scene);
}
...
// At all actual BL_Converter::RemoveScene calls from KX_KetsjiEngine
     DestructScene(scene);
     m_scenes->RemoveValue(scene);

```

marechal-p:
Why isn't `m_scenes->RemoveValue` called from `KX_KetsjiEngine::StopEngine` though ?

marechal-p:
Ok, so maybe it is because in `KX_KetsjiEngine::StopEngine` everything is deleted in one line with `m_scenes=>Remove(0);`, my guess is that it avoids to parcours the whole list on each deletion ? I am not sure...

marechal-p:
@panzergame changes addressed.

panzergame:
@lordloki : I'm waiting your review.

youle31:
yes

Issue #746, opened 2018-07-14
Navmesh Clear Data: Blender stop responding.


UnidayStudio:
It crashes if you click on "NavMesh Clear Data" and press P to play. To fix, you've to reset index values again.

I already pressed Clear Data. So just press P and enjoy it stop responding. :)

Blend file:
http://pasteall.org/blend/index.php?id=49940

panzergame:
Bug confirmed.

lordloki:
Closing. Fixed in master. Thanks for reporting.

PR #752, opened 2018-07-17
UPBGE: Fix for bug #746 Navmesh related

lordloki:
Previously when the navmesh was removed from the object this one kept the OB_NAVMESH flag set and the engine tried without success to process it as a navmesh.

Now OB_NAVMESH flag is removed.

Additionally, Name for button "Navmesh Clear Data" is modified to "Remove Navmesh" to avoid possible misunderstandings.

### DIFF
--- a/source/blender/editors/mesh/mesh_navmesh.c
+++ b/source/blender/editors/mesh/mesh_navmesh.c
@@ -709,6 +709,7 @@ static int navmesh_clear_exec(bContext *C, wmOperator *UNUSED(op))
 	Mesh *me = ob->data;
 
 	CustomData_free_layers(&me->pdata, CD_RECAST, me->totpoly);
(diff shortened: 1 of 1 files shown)


BluePrintRandom:
any hope  at getting rebuild navmesh to work using KX_MeshProxy instead of the old BPY scene mesh?

panzergame:
The only noticeable side effect is that when the user remove the data from a navmesh this object is not showed as a navmesh any more in the physiscs panel.

lordloki:
I think that this behavior is wanted at the end as when you removed the navmesh from a mesh if you assign it again to the navmesh physics tab then a new navmesh should be created. This way we will avoid the issue if we launch the engine with the mesh with the navmesh data removed but with the navmesh tab selected (the engine will try to convert the mesh to navmesh without success).

PR #761, opened 2018-07-24
UPBGE: Additional fixes to can compile under MacOS X

lordloki:
related to TBB library detection and not following the std::abs standard

@panzergame are the changes over the mathfu library ok for you?

### DIFF
--- a/intern/mathfu/mathfu/vector.h
+++ b/intern/mathfu/mathfu/vector.h
@@ -969,9 +969,13 @@ static inline T DotProductHelper(const Vector<T, 4>& v1,
 
 template <class T, int d>
 static inline bool FuzzyZeroHelper(const Vector<T, d>& v) {
-  T abs = 0;
+  T absolute = 0;
   for (int i = 0; i < d; ++i) {
-    abs += std::abs(v[i]);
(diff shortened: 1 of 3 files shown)


intern/mathfu/mathfu/vector.h
```diff
@@ -969,9 +969,13 @@ static inline T DotProductHelper(const Vector<T, 4>& v1,
...
   for (int i = 0; i < d; ++i) {
-    abs += std::abs(v[i]);
+#ifdef __APPLE__
+    absolute += abs(v[i]);
+#else
+    absolute += std::abs(v[i]);
+#endif
   }
 
   return FuzzyZero(abs);
```
panzergame: Is this line working fine ?

lordloki: Totally sure it won't work. copy&paste error.

lordloki:
Regarding to replace std::abs for abs, I have just found that including <cstdlib> header then std::abs could be used under Mac. I will check.

lordloki:
After tested what i said before same error. Then I use abs instead of std::abs for all OS.

PR #766, opened 2018-07-28
UPBGE: Fix compound ray cast with triangle child shape.

panzergame:
Previously when a compound shape was hit by a ray cast the information about
the children triangles was inaccessible because of a lack from the bullet
API.

To fix this issue we have to keep the triangle index but add a child index
of the compound child shape hit. m_childIndex is dedicated for in LocalRayResult
and its value is set in rayTestSingleInternal, -1 for non compound shapes.

This value is catch from FilterClosestRayResultCallback and used in RayTest
to get the actual collision shape. To still get the shape construction info
of the children, each collision shape created has as user pointer the
shape costruction info. This object is get in RayTest to find the mesh
and polygon.

Tested with:
- mix of convex and concave shape in a compound shape.
- non compound concave and convex shape.

Example file:
[ge_compound.zip](https://github.com/UPBGE/blender/files/2237920/ge_compound.zip)
The hit polygon color is set to red. The compound shape is made of two triangle mesh (concave) and a box (convex), the fourth object is a non-compound triangle mesh (concave).

### DIFF
--- a/extern/bullet/src/BulletCollision/CollisionDispatch/btCollisionWorld.cpp
+++ b/extern/bullet/src/BulletCollision/CollisionDispatch/btCollisionWorld.cpp
@@ -340,8 +340,8 @@ void	btCollisionWorld::rayTestSingleInternal(const btTransform& rayFromTrans,con
 						collisionObjectWrap->getCollisionObject(),
 						0,
 						castResult.m_normal,
-						castResult.m_fraction
-						);
+						castResult.m_fraction,
+						-1);
 
 					bool normalInWorldSpace = true;
 					resultCallback.addSingleResult(localRayResult, normalInWorldSpace);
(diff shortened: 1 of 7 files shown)


youle31:
Hi, I just tested and I noticed the different behaviours in master and in ge_compound (So I guess this is working) but I don't really understand (I know nothing about compound...).

BluePrintRandom:
compound is super useful for setting up character collisions or vehicles,

we have a root, and the child shapes can be remove parent / updated /
re-parented to a core object and it has accurate collision for complex
characters with little overhead.

now with this patch, I don't have to check which part I hit with python :D

can we return hitChildObject w/the collision callbacks later if the object
is compound etc?



I am eager to test but I won't have much time for the next 2 weeks and my
county is on fire....

panzergame:
This patch is compatible with previous master as only the triangle information is provided. Theoretically the ray cast could also return the child object directly, but in this case the behaviour is totally changed (before only the root object was hit) and i don't see a correct way to return both compound root object and child object in a ray cast.
The same apply for collision callbacks but in this case I can add a child index member per collision point.

BluePrintRandom:
Can we overload the return of the raycast and return the child shape as the
last arg?

Issue #728, opened 2018-07-02
Spherical Area Light Fall-Off Attenuation


adriansnetlis:
So currently this is how inverse squared calculation looks:
`attenuation = 1.0 / pow(distance / the_distance_number_in_light_settings, 2);` (it may look a bit different, but that's the math.
However, this does not provide the best results possible. Real lights have radius.
Sourcing for this(https://imdoingitwrong.wordpress.com/2011/01/31/light-attenuation/), you can calculate spherical area light attenuation by this calculation:
`attenuation = 1.0 / pow(distance / radius + 1.0, 2);`
If one wanted to keep the distance setting, radius can be multiplied by the distance number in light settings. 
This is not very costly calculation by itself, however, the source provides even additional optimizations for deferred rendering.

Now adding "radius" option in light settings and switching to this equation could provide pretty nice results.

panzergame:
Thanks for the detailed report, I leave it to new developer to get experience:
The fragment shader `source/blender/gpu/shaders/gpu_shader_material.glsl`need to be modified by adding a new attenuation function and a check added in gpu_material.c to select it.

lordloki:
This was already implemented by @DCubix in the ge_brdf branch (https://github.com/UPBGE/blender/commit/eb97104d06e965239e42129dcb2536491c65f28e#diff-cd270f0456f47fde6ee278bee899bdb7R1810). Maybe we can commit it to master.
I will test it.

lordloki:
committed in master. Thanks for testing and proposal

PR #768, opened 2018-07-29
UPBGE: Implement new Lamp Falloff

lordloki:
- Inverse Square Cutoff for Lamps. Describes here:
https://imdoingitwrong.wordpress.com/2011/01/31/light-attenuation/

Code by @DCubix

For feature request #728


source/blender/gpu/shaders/gpu_shader_material.glsl
```diff
@@ -1847,6 +1847,16 @@ void lamp_falloff_curve(float lampdist, sampler2D curvemap, float dist, out floa
 	visifac = texture2D(curvemap, vec2(dist / lampdist, 0.0)).x;
 }
 
+void lamp_falloff_invsquarecutoff(float radius, float dist, float cutoff, out float visifac)
+{
+	float d = max(dist - radius, 0.0);
+	float denom = d / radius + 1.0;
+ 	float att = 1.0 / (denom * denom);
+	att = (att - cutoff) / (1.0 - cutoff);
```
panzergame: `cutoff` is from 0 to 1 in the UI definition:
```c
RNA_def_property_range(prop, 0.0f, 1.0f);
```
And here it could cause 0 division. You could limit it to 0.99.

panzergame:
@adriansnetlis I invited you so that you could do a review on this patch if you please.

source/blender/gpu/shaders/gpu_shader_material.glsl
lordloki: I just to check other shaders functions and the indeterminations are handle inside the shader instead of clamping the property range.

Perhaps we can optimize the lamp_falloff_invsquarecutoff function as follow:
```c
void lamp_falloff_invsquarecutoff(float radius, float dist, float cutoff, out float visifac)  
{
	if (cutoff == 1.0) {
		visifac = 0.0;
	}
	else {
		float d = max(dist - radius, 0.0);  
		if (d == 0) {
			visifac = 1.0;
		}
		else {
			float denom = d / radius + 1.0;  
			float att = 1.0 / (denom * denom);  
			att = (att - cutoff) / (1.0 - cutoff);  
			att = max(att, 0.0);  
			visifac = att;
		}
	}
}
```
What do you think?

panzergame: Is there's an interest for the user to set a cutoff to 1 or a zero radius ?

I agree for the second check level, you can even remove the max a check d <= 0.0.

Also about the UI, you can define a new RNA attribute named radius that is the same as distance attribute.

lordloki: Ok, changes done

source/blender/blenloader/intern/versioning_defaults.c
```diff
@@ -310,5 +311,13 @@ void BLO_update_defaults_startup_blend(Main *bmain)
 			br->flag |= BRUSH_ACCUMULATE;
 		}
 	}
+
+	for (Lamp *lamp = bmain->lamp.first; lamp; lamp = lamp->id.next) {
```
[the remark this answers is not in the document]
lordloki: Ok. Done.

I changed the coeff_* as they didn't have init values (setted to 0.0) but i will do in an independent commit.

Issue #756, opened 2018-07-19
Interface Cleanup and Improvements


joelgomes1994:
Hello! Here's a proposal to remove some useless stuff from UPBGE's interface and improve the current one.

- **Info** editor
   - Rename renderer from *Blender Game* to *UPBGE*
- **3D View** editor
   - **Add** context menu
      - Remove non-compatible objects from menu, like **Speaker, Force Fields**, etc *(only for *UPBGE* renderer, don't know if it's possible)*
- **Properties** Editor
   - **Scene** tab
      - Remove **Keying Sets** panel *(hardly would apply to game dev process)*
      - Remove **Audio** panel *(don't affects GE)*

There's a chance of some of those features being useful on GE and I don't know how (like the **Color Management** / **Color Space** stuff), but I'm sure the remaining aren't useful (like the **Audio** panel, only working in **Blender Render**).

Thanks. :smile:

adriansnetlis:
Bake panel is often used for quick AO or normal baking (although I more often bake AO in Cycles), removing it makes no sense.
Color management, instead of being removed, should work for GE (being able to switch between sRGB and Linear or even using built-in Filmic tonemap is always a good thing).

Color space for images matters even in game engine. Like you want to set that to Linear for normalmaps, roughness maps, metallness maps etc. Makes difference and can fix things that would appear broken otherwise.

UnidayStudio:
I liked the idea to remove things that don't work in BGE and change from Blender Game to UPBGE. But I also think that we should try to make some of the things that don't work to work (Idk if english works this way, but you get the point): The Speaker object (The Converter should convert them into an empty with a sound component, or something like this), Color Management (yeah, It will be great to see this working) and Audio. I also use Bake very often to get information from the high poly versions of my models.

joelgomes1994:
@adriansnetlis Thanks for the remarks, it's a proposal of removing useless features in GE, so I will update the top post according to the replies. Nice to know about the color management (although I didn't noticed changes in GE).

@UnidayStudio Although I agree implementing those features (specially a functional Audio panel), I think they don't fit in a 'cleanup' issue, which aims to be easy and fast to implement (to remove stuff?). However, it would be redundant to use an Speaker object, as any object type can be a sound source in 3D space. The idea is to cleanup interface to avoid useless features to be accessed, even if it means to remove the useless features and then add them again once they are implemented in GE.

panzergame:
About the add context, lattice is still used to mesh construction.

PR #772, opened 2018-07-31
UPBGE: Cleanup UI.

panzergame:
Some unused options/entities are removed from the UI in game engine mode:
- scene keying set
- scene audio
- speaker and forec filed in add menu

Also the game engine is renamed UPBGE.

Fix issue #756.

### DIFF
--- a/release/scripts/startup/bl_ui/properties_game.py
+++ b/release/scripts/startup/bl_ui/properties_game.py
@@ -720,6 +720,30 @@ class SCENE_PT_game_console(SceneButtonsPanel, Panel):
         row.prop(gs, "python_console_key4", text="", event=True)
 
 
+class SCENE_PT_game_audio(SceneButtonsPanel, Panel):
+    bl_label = "Audio"
+    bl_options = {'DEFAULT_CLOSED'}
+    COMPAT_ENGINES = {'BLENDER_GAME'}
+
+    @classmethod
+    def poll(cls, context):
+        scene = context.scene
+        return (scene and scene.render.engine in cls.COMPAT_ENGINES)
+
+    def draw(self, context):
+        layout = self.layout
+
+        scene = context.scene
+
+        split = layout.split()
+
+        col = layout.column()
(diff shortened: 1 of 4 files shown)


marechal-p:
> I'd like to point that a speaker is much more intuitive to the new user to be a "sound reproducer".It will increase the discoverabilty of these feature (sound) in the engine. And, of couse, even for the experienced user, having a GameObject that is alrady prepared to reproduce sounds, without any other configuration, will increase the development speed.

That's an issue on its own that deserves its own PR. Better remove the item from the UI until it has an actual purpose.

youle31:
I agree with this patch and I agree both with @UnidayStudio  and @marechal-p remarks

UnidayStudio:
We can remove Area lights as well

panzergame:
I can't remove AREA in the lamp type enum because the front end UI doesn't provide a way to filter enum properties.
Otherwise it possible to replace the UI of area lamp by a message like 'not supported in game engine'.

panzergame:
@lordloki : The function using these variables is:
```c
void BKE_sound_update_scene_listener(struct Scene *scene)
{
	AUD_Sequence_setSpeedOfSound(scene->sound_scene, scene->audio.speed_of_sound);
	AUD_Sequence_setDopplerFactor(scene->sound_scene, scene->audio.doppler_factor);
	AUD_Sequence_setDistanceModel(scene->sound_scene, scene->audio.distance_model);
}

```

Though it is using `scene->sound_scene` which doesn't seems global or registered to a global variable and also not used by the game engine. The type of AUD_Sound is also strange.

I didn't do a deep research, @lordloki : Do you have a example of these values used in game ?

panzergame:
@lordloki : These attributes are now exposed.

lordloki:
@panzergame I have a issue with audio panel when revert to other render from UPBGE.

Fix plus cosmetic refactor: https://gist.github.com/lordloki/a34ef600e1255f4b93d087eaaf7505e9

PR #774, opened 2018-08-03
UPBGE: Implement render extra attachment.

panzergame:
In the purpose of complex filter, data from the rendering other than the final
color or the depth could be needed in filter like space reflection or ambient
occlusion. The obvious one is the view normal, but any kind of data could be
interesting passing from the albedo or diffuse to a debug color for screen ray
cast.
To expose such feature to the user, first a list of attachment need to be
defined and secondly a way to output to these from simple material or node
material.

The attachments are to the number of 7, because the 8th is the first one
dedicated to the final color. The UI is defiend in "Render" panel under
"Attachments" sub panel. A list of 7 items is exposed and the user can add an
attachment or remove one from any item, the items have four properties, the name
(only for debug), the type : either custom, normal or albedo, in case of custom
the properties size and hdr are exposed, they both select the size of data into
the attachment texture and it's precision type.

Each GPU material is now tracking 8 out links, the first one is the same as the
previous only out link and the other match the extra attachment. They are all
iterated in codegen_call_functions to generate the GLSL lines for
"gl_FragData[i] = ", a out link is registered using the function
GPU_material_output_link.

The attachment type "custom" is dedicated only for node materials where any kind
of data could be output to a special node. This node is named "Output
Attachment" and has only one input and a property for the index of the
attachment, this node is responsible of registering the output link.
The other attachment types "normal" and "abeldo" are detected in
GPU_shaderesult_set and generate link for the view normal and the material
diffuse (including vertex color and textures) multiplied by the object color.

From the game engine side, the attachment info are converted into LA_Launcher
and put inside RAS_ICanvas, this same class is now responsible of owning and
creating the offscreens, by doing this the canvas can also detect properly a
resize and recreate the offscreens. Other than that the function are equivalent,
GetOffScreen is now called from RAS_ICanvas instead of RAS_Rasterizer.

The attachments are readable form any filters defining the sampler uniform
"bgl_DataTextures[7]", this uniform is bind as well as other uniforms in
RAS_2DFilter::BindUniforms and the attachments textures in
BindTextures/UnbindTextures.

Because GPUOffScreen is not anymore used and is less flexible than hand
configuration of a frame buffer, their sources are restored to the blender ones.

Example file:
[ge_render_attachment.zip](https://github.com/UPBGE/blender/files/2257463/ge_render_attachment.zip)

Press key 1 or 2 to select the albedo attachment or the custom attachment, the default one if view normals.


youle31:
wow, big work. The test file seems to work fine for me. I don't know if there are some volounteers to create AO and SSR filters, this would be very cool. Maybe some martinsh works done with image render could be ported using this feature instead of image render

lordloki:
@panzergame could you check valgrind? 
I always have "not freed memory error" with this pull request.

release/scripts/startup/bl_ui/properties_game.py
```diff
@@ -1026,9 +1062,11 @@ def draw(self, context):
     RENDER_PT_game_stereo,
     RENDER_PT_game_shading,
     RENDER_PT_game_system,
+    RENDER_PT_game_attachments,
     RENDER_PT_game_animations,
     RENDER_PT_game_display,
     RENDER_PT_game_debug,
+	RENDER_UL_attachments,
```
lordloki: TAB instead of spaces

lordloki:
Win64 build to test it: https://mega.nz/#!8x8gQADL!VzigBmKfl36FaL5ASEWSsnZ1Lz4uIb1rMU6IM34I-A0

panzergame:
@lordloki : I'm fixing this, thanks for reporting.

panzergame:
@lordloki : The memory leak is fixed.

lordloki:
@panzergame Leak is gone. Good work. Code seems OK to me.

sergio5528:
@panzergame  and upbge team thanks,this will help to make a lot of optimized filter shaders <3

source/blender/makesrna/intern/rna_nodetree.c
```diff
@@ -3593,6 +3593,17 @@ static void def_sh_output(StructRNA *srna)
 	RNA_def_property_update(prop, NC_NODE | NA_EDITED, "rna_Node_update");
 }
 
+static void def_sh_output_attachment(StructRNA *srna)
+{
+	PropertyRNA *prop;
+
+	prop = RNA_def_property(srna, "attachment", PROP_INT, PROP_NONE);
+	RNA_def_property_int_sdna(prop, NULL, "custom1");
+	RNA_def_property_range(prop, 0, 7);
```
panzergame: Use GAME_ATTACHMENT_COUNT

lopas85:
if i'm using custom material shaders, (for example, i need to move vertices in my VertexShaders to make grass or trees) what i have to do in order to output in bgl_DataTextures[0], bgl_DataTextures[1] ..etc?
In fragmentShaders i'll simply output to gl_FragColor 
so i'll do gl_FragColor = vec4(albedo , 1.0);
then i would like to have normals and pbrMatData like so:
bgl_DataTextures[0] <--- normals
bgl_DataTextures[1] <--- pbrMatData
How to achieve this?

PR #777, opened 2018-08-05
UPBGE: Implement color management option.

panzergame:
Previously the color management was always of sRGB and the user get no control
over it. To fill this gap the user should have the ability to define the final
color space, linear or sRGB.

First of all an option named "Color Management" in System tab under Render panel
is added, it can be "Linear" or "sRGB", the default is "sRGB", When game engine
is selected changing this option will force the recompilation of all shaders and
they will add or remove color space conversion at the end of the shader.

In game the color space is not managed at the shader level, but at the final
copy of the offscreen to the screen. Firstly the material in game must disable
the color management, this is done by a new set of flags passed to
GPU_material_from_blender and the flag GPU_MATERIAL_NO_COLOR_MANAGEMENT.
Secondly the three shaders used to copy the offscreen to screen (normal, stipple
stereo, anaglyph stereo) now have a sRGB version including a call to
linearrgb_to_srgb enabled by the macro COLOR_MANAGEMENT.
The color space used is stored in the rasterizer in enum variable
m_colorManagement for linear and sRGB, this variable is used to select the
correct shader to bind before flushing to screen.

In the same time the color management used in KX_FontObject or KX_WolrdInfo is
now obsolete and removed.

To test this PR, you can take any blend file and try to change the "Color Management" option in "System" panel.

Is the option at the right place for you ?


youle31:
Hi, it seems there is an issue with background color in this file: http://pasteall.org/blend/index.php?id=50094

I'd have put the linear to srgb code in gpu_shader_frame_buffer_frag.glsl too but I don't know if this can cause issues with previous behaviour. I have to test overlay/background scenes too to see if there is no issue.

Else, there is an existing UI for color management when we are in Blender Render in scene tab. This can lead to confusion if there is 2 UIs I think

panzergame:
@youle31 : Thanks for the report, indeed the world material wasn't removing the linear to sRGB conversion, this is now fixed.

Concerning the UI, the previous was using a complex system based on IMB_* that queried all the color management available (like the one proposed by OCIO). Also the option was per scene which is not the case currently.
Although i'm really not convinced of the place of this option in the UI, i'm just certain that it should be in render tab, but after that i'm still hesitating to let it in an unique panel.

youle31:
I confirm the bug is fixed.

About the UI, maybe AA samples, HDR options and color management could be grouped in another commit... I don't really know.

Else I have banding effect in BGE that I can't see in the viewport. This issue is not in master. Maybe it needs a large screen to see it:
viewport:
viewport
bge:
bge

panzergame:
Hum, I think that it's because previously the linear to sRGB conversion was using the float color from the material shader, now it's using 255 range values.
You certainly have better result with HDR, or by using GL_R11F_G11F_B10F instead of GL_RGBA8 in GPU_texture_create_nD.

lordloki:
Tested and ok.
Regard to ui i think it is better an own "Color Management" panel under reder tab (removing current under scene tab).

adriansnetlis:
Cannot review at the moment. Looks like all Blender versions on git have a dependency of Python 3.6 which is not a thing in Arch repositories any more from what I can see. Probably could solve that by having a manually installed directory for it, but I am not willing to do that yet.

panzergame:
About the quality, the issue is that RGBA8 is not precise enough to avoid bands after color management.

The idea is to use R11G11B11F instead of RGBA8 and avoid the alpha channel, with the following conditions:
The user is not using alpha in 2d filters or screenshots. This is hard because the background is always rendered with an opaque color.
Concerning render to texture, it will keep RGBA8 as the alpha is more frequently used.

adriansnetlis:
Hm... R11G11B11F is 32-bit. I think one of channels is supposed to be 10-bit.

Also, if this bitrate is used, it could be viable to send 10-bit color data to screen if HDR monitor is connected, right?

panzergame:
Yes, it's R11G11B10, unfortunately we can't use it because the user can select a blending using the destination alpha.

What about defaulting the HDR setting to RGBA16F ?

youle31:
The old files saved with RGBA8 would have a loss of quality. As you want

panzergame:
We can force to convert the old file HDR to HDR_HALF_FLOAT. Any ideas are welcome.

youle31:
Remove RGBA8? (on my screen, the banding effects is annoying then I think I won't use it anymore but I don't know if there are cases where it can still be useful)

panzergame:
GL_RGBA12 is now used for the default HDR_NONE.

lordloki:
No performance decrease noted with GL_RGBA12

Issue #633, opened 2017-11-08
navMesh.replaceMesh('mesh',1,1) + navmesh.rebuild() does not change the navmesh


BluePrintRandom:
I believe rebuild is supposed to create a new navmesh after mesh updates or replacement has happened,

I am not sure why but this appears to be failing

http://pasteall.org/blend/index.php?id=48266
steps to see bug

press P to play

press space bar to run
```python
import bge


def main():

    cont = bge.logic.getCurrentController()
    own = cont.owner
    own.draw(0)
    own.replaceMesh('Navmesh',1,1)
    print('Replacing mesh')
    
    own.rebuild()
    print('rebuilt')
main()
```

this shows the bounds of the original mesh, not the new mesh 

rebuild does not appear to be working* or replace mesh does not change the mesh used by the navmesh?

panzergame:
There's a way in KX_NavMeshObject to convert from KX_Mesh but is never used and also nav mesh have an extra CD_RECAST layer that i have to identify…

panzergame:
I finished a patch (currently writing the commit message), but I noticed that even rebuild with a mesh including detailed nav mesh information was doing nothing…

PR #778, opened 2018-08-06
UPBGE: Fix navmesh rebuild after mesh replacement.

panzergame:
Previously the navmesh rebuilding was failing because of two issues,
first the mesh used as source was always the one from the blender object
secondly the navmesh itself wasn't updated in the obstacle simulation.

To solve this first issue the function BuildVertIndArrays is calling
two sub functions, BuildFromDerivedMesh for creating from a derived
mesh and benefit from CD_RECAST layer, if this function fails to get
the recast layer BuildFromMesh is called and generate information from
a RAS_Mesh without detailed meshes.
In this last function the indices computation is solved by the requirement
of Detour library to use 6 indices per polygon, the unused indices are
discarded by value 0xFFFF.

The second issue is solved by calling AddObstaclesForNavMesh only from
KX_NavMeshObject::BuildNavMesh, this call is preceded by a call to
DestroyObstacleForObj intenting to remove the previous obstacle
associated to this navmesh.

Fix issue #633.

Example file : 
[ge_navmesh.zip](https://github.com/UPBGE/blender/files/2262978/ge_navmesh.zip)

Press `Space` to replace the mesh and rebuild.

### DIFF
--- a/source/gameengine/Ketsji/KX_NavMeshObject.cpp
+++ b/source/gameengine/Ketsji/KX_NavMeshObject.cpp
@@ -124,11 +124,6 @@ void KX_NavMeshObject::ProcessReplica()
 		CM_FunctionError("unable to build navigation mesh");
 		return;
 	}
-
-	KX_ObstacleSimulation *obssimulation = GetScene()->GetObstacleSimulation();
-	if (obssimulation) {
-		obssimulation->AddObstaclesForNavMesh(this);
-	}
 }
 
 int KX_NavMeshObject::GetGameObjectType() const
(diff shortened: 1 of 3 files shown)


BluePrintRandom:
Thank you panzer game!!

youle31:
Hello, I did a quick test and it seems to work fine. Great job! Thanks :)

source/gameengine/Ketsji/KX_NavMeshObject.cpp
```diff
@@ -124,186 +124,220 @@ void KX_NavMeshObject::ProcessReplica()
...
 }
 
 int KX_NavMeshObject::GetGameObjectType() const
 {
 	return OBJ_NAVMESH;
 }
 
-
-bool KX_NavMeshObject::BuildVertIndArrays(float *&vertices, int& nverts,
+bool KX_NavMeshObject::BuildFromDerivedMesh(float *&vertices, int& nverts,
```
panzergame: This function doesn't need review as it is just renamed and unindented.

PR #781, opened 2018-08-08
UPBGE: Fix orthographic camera getScreenRay.

panzergame:
Previously the function getScreenRay was adapted to only perspective
cameras because the ray cast was always starting from the center of
the camera.
In case of orthographic the ray cast should start from a point unprojected
from the near plane and continue to in the direction of the camera
(-Z orientation axis).
This is proceeded in getScreenRay and the indirect call to getScreenVect
is removed.

### DIFF
--- a/source/gameengine/Ketsji/KX_GameObject.cpp
+++ b/source/gameengine/Ketsji/KX_GameObject.cpp
@@ -3886,20 +3886,13 @@ EXP_PYMETHODDEF_DOC_O(KX_GameObject, getVectTo,
 	return returnValue;
 }
 
-struct KX_GameObject::RayCastData {
-	RayCastData(std::string prop, bool xray, unsigned int mask)
-		:m_prop(prop),
-		m_xray(xray),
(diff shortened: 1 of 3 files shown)


youle31:
Works fine for me in @UnidayStudio test file. This was an old bug I already had. Great job!

source/gameengine/Ketsji/KX_GameObject.cpp
```diff
@@ -4103,14 +4096,14 @@ EXP_PYMETHODDEF_DOC(KX_GameObject, rayCast,
 
 	if (dist != 0.0f) {
 		mt::vec3 toDir = toPoint - fromPoint;
-		if (mt::FuzzyZero(toDir.LengthSquared())) {
+		if (mt::FuzzyZero(toDir)) {
```
panzergame: This is changed because comparing the squared length is less precise than comparing the Manhattan distance for small vectors.

panzergame:
Thanks for the review and test.

PR #787, opened 2018-08-10
UPBGE: Optimize transparent mesh sorting.

panzergame:
The previous sorting of meshes was for each mesh slot (or node leaf)
computing the distance with the near plane of the camera and these
distances were used in std::sort.

This process can be improved by speed and accuracy, first the computation
of the distance and the sort could be parallelized thanks to TBB,
secondly the center of the object is not the most accurate for sorting
objects, the distance from camera to the center of the bounding box
+ the bounding box radius is more accurate.

The code managing the mesh sorting is moved in RAS_SortedMeshSlot.[h/cpp]
RAS_SortedMeshSlot is linking the distance to camera with a mesh slot,
its constructor is computing the distance using bounding box.

RAS_SortedMeshSlot::Sort is receiving a list of mesh slots or node leafs
and managing RAS_SortedMeshSlot creation and sorting. The creation is
using TBB and the task struct InitLeafsTask, then the RAS_SortedMeshSlot(s)
are sorted through tbb::parallel_sort.

To acces the bounding box, RAS_DisplayArray is using its own AABB
with the adjunction of a center and radius cache. These data are
computed into UpdateAabb called from RAS_MeshBoundingBox. In case
of test object the display array is null and the bounding box
(RAS_BoundingBox) from the mesh user is used.

Example file: 
[ge_mesh_sort.zip](https://github.com/UPBGE/blender/files/2278810/ge_mesh_sort.zip)

### DIFF
--- a/source/gameengine/Rasterizer/CMakeLists.txt
+++ b/source/gameengine/Rasterizer/CMakeLists.txt
@@ -82,6 +82,7 @@ set(SRC
 	RAS_OffScreen.cpp
 	RAS_Query.cpp
 	RAS_Shader.cpp
+	RAS_SortedMeshSlot.cpp
 	RAS_Texture.cpp
 	RAS_TextureRenderer.cpp
 	RAS_TextUser.cpp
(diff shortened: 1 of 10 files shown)


youle31:
Hi, it seems to work fine for me in your test file:
Master:
master
ge_mesh_sort:
ge_mesh_sort
(I had to replace static_partitionner with auto_partitionner to fix a tbb compile issue)

PR #790, opened 2018-08-13
UPBGE: Restore object info node.

panzergame:
Previously the object info node was doing nothing directly by the fact
that GPU_material_bind_uniforms was using nullptr for the last argument
corresponding to the float array of object info.

The pass indices used by the object info is stored either in the
blender material and the game object under the same variable m_passIndex.
The object is coppying this value to the mesh user and BL_BlenderShader::Update
is called with the material pass index as argument and inside it
get the object pass index from the mesh user. These information are
used to set the object info and the random value is computed from
the hash of the mesh user client object, a per object value.

The pass index is exposed in python too under attributes:
KX_BlenderMaterial.passIndex
KX_GameObject.passIndex

This commit is used as an example documentation : https://github.com/UPBGE/blender/wiki/Implementing-pass-index

Example file: 
[ge_pass_index.zip](https://github.com/UPBGE/blender/files/2284431/ge_pass_index.zip)

### DIFF
--- a/source/gameengine/Ketsji/BL_BlenderShader.cpp
+++ b/source/gameengine/Ketsji/BL_BlenderShader.cpp
@@ -122,6 +122,9 @@ RAS_InstancingBuffer::Attrib BL_BlenderShader::GetInstancingAttribs() const
 	if (builtins & GPU_INSTANCING_LAYER) {
 		attrib = (RAS_InstancingBuffer::Attrib)(attrib | RAS_InstancingBuffer::LAYER_ATTRIB);
 	}
+	if (builtins & GPU_INSTANCING_INFO) {
+		attrib = (RAS_InstancingBuffer::Attrib)(attrib | RAS_InstancingBuffer::INFO_ATTRIB);
+	}
 
 	return attrib;
 }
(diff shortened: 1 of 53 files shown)


panzergame:
It needs GI support.

source/gameengine/Ketsji/BL_BlenderShader.cpp
```diff
@@ -153,8 +155,15 @@ void BL_BlenderShader::Update(RAS_MeshSlot *ms, RAS_Rasterizer *rasty)
 	RAS_MeshUser *meshUser = ms->m_meshUser;
 	const float (&obcol)[4] = meshUser->GetColor().Data();
 
+	float objectInfo[3];
+	if (GPU_get_material_builtins(m_gpuMat) & GPU_OBJECT_INFO) {
+		objectInfo[0] = meshUser->GetPassIndex();
+		objectInfo[1] = matPassIndex;
+		objectInfo[2] = BLI_hash_int_2d((uintptr_t)meshUser->GetClientObject(), 0) / ((float)0xFFFFFFFF);
+	}
```
panzergame: Store this value in RAS_MeshUser as the client will not change and this value too.

youle31:
Hello, it is working for me. This changes a lot of code, but if you think it can be useful for users, then it's cool.

Maybe this kindof work can be used to eventually set materials uniforms from python (just a float that the user could set from python (like GPU_TIME but a float choosed by the user (GPU_TIME could also be set directly from python if the user want to reset it or to give it a chosen value))...

Issue #793, opened 2018-08-17
Get python component module via bpy


Momel15:
I'm working on an addon that needs to access the python component's "path". There is an attribute called "name", but this only returns the component/class name of the component, not the module (in module.ComponentName).

Is there a way to get this? If not, please see this issue as a feature request, it would be great to have access to the full path :)

PR #795, opened 2018-08-19
UPBGE: Improve component UI.

panzergame:
Previously the components name was exposed by a property non-editable which
was rendered in light gray and the component box was always expanded.
Also the module name was inaccesible from bpy.

The last issue is simply solved by exposing the "module" string to RNA.
The name property is replaced by a label in the UI. To expanse the box
the property "show_expanded" similar to the logic brick is implemented,
when this property is True, an inner box is generated containing all the
properties.

Fix issue #793.

component_ui_1
component_ui_0

### DIFF
--- a/release/scripts/startup/bl_ui/space_logic.py
+++ b/release/scripts/startup/bl_ui/space_logic.py
@@ -45,15 +45,18 @@ class LOGIC_PT_components(bpy.types.Panel):
         for i, c in enumerate(game.components):
             box = layout.box()
             row = box.row()
-            row.prop(c, "name", text="")
+            row.prop(c, "show_expanded", text="", emboss=False)
+            row.label(c.name)
             row.operator("logic.python_component_reload", text="", icon='RECOVER_LAST').index = i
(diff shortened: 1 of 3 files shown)


UnidayStudio:
Improvements here!! I liked.
What about the button to reload all the components at once? And the Color variable?

youle31:
hi, I tested with third person component in your test file @panzergame (component_0.blend). This seems to work fine.

When I open the file I have this message in the console:

Read blend: C:\Users\youle\Desktop\Component_0.blend
  File "C:\Users\youle\Desktop\Component_0.blend\test.py", line 1, in <module>
ModuleNotFoundError: No module named 'bge'

panzergame:
This message is normal, I made the mistake of defining a module of name `test`which is also used by other part of python.

panzergame:
Despite the auto exec that is the topic of #809, @UnidayStudio : Do you agree with the UI and the code making it ?

UnidayStudio:
Yes!

PR #796, opened 2018-08-19
UPBGE: Allow alpha support for mathutils.Color.

panzergame:
mathutils.Color can now optionally support the alpha channel by passing
a four elements tuple to its python constructor or passing the size
to the Color_CreatePyObject* functions.

The size fo the color can only be 3 or 4 as the fonction
mathutils_array_parse_alloc is checking. This size is stored into ColorObject
and used at every place COLOR_SIZE was used.

Operation between color of different size is disallowed same as vector.


youle31:
Quickly tested and works for me. (general remark: I'm wondering why we have 2 types: Color + Vector when the 2 types seem to behave in the same way)

panzergame:
The only advantages f Color is the hsv values. But yes this could be an extension of Vector.

PR #803, opened 2018-08-22
UPBGE: Sleep when frame time is smaller than fixed frame time.

panzergame:
Previously when the frame time was too fast the function KX_KetsjiEngine::NextFrame
concluded that 0 frames have to be proceeded and done nothing. This function was
called until the elapsed time was suffisent to process a frame.

This approch is naive, robust and precise but introduce an high CPU usage because
of the extra calls to KX_KetsjiEngine::NextFrame setting the CPU to it's maximum.

To avoid this high usage, the elapsed time to the next frame is computed and the
main thread sleep of this time.

### DIFF
--- a/source/gameengine/Ketsji/KX_KetsjiEngine.cpp
+++ b/source/gameengine/Ketsji/KX_KetsjiEngine.cpp
@@ -37,6 +37,7 @@
 #include "CM_Message.h"
 
 #include <boost/format.hpp>
+#include <thread>
 
 #include "BLI_task.h"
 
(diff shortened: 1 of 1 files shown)


marechal-p:
Although when trying to test corner cases, I don't know if one issue is due to this commit or if it is expected:

When sleeping for 15ms each frame, the framerate still holds at 60fps, one logic step per rendered frame.
The issue happens when we sleep for 16ms or more, the logic here takes a bit longer than one render step. Now for some reason, at this point, we end up doing 5 logic steps in one frame, each step taking 16ms or more, ending up in frames rendered every 80ms apart or so (11~14fps) instead of just being delayed by the overtime past the standard 16ms.

Maybe this deserve its own issue though.

panzergame:
@marechal-p I think it should deserve its own issue. The system of sub logic and physics frame should be reworked.

source/gameengine/Ketsji/KX_KetsjiEngine.cpp
```diff
@@ -349,35 +350,40 @@ bool KX_KetsjiEngine::NextFrame()
...
-		frames = int(deltatime * m_ticrate / m_timescale + 1e-6);
+			// If the elapsed time induce a higher framerate, sleep until the next frame time point.
+			if (frames == 0) {
+				const double sleeptime = 1.0 / scale - deltatime;
+				std::this_thread::sleep_for(std::chrono::nanoseconds((long)(sleeptime * 1.0e9)));
+				frames = 1;
+			}
+		}
+		else {
+			timestep = dt * m_timescale;
```
lordloki: Just a remark, since the frame variable declaration is a bit far away, wouldn't it be better to assign the frame=1 just here below? Only to make more readable but I'm ok with your solution too

PR #809, opened 2018-08-26
UPBGE: Fix default auto exec option for new created user prefs.

panzergame:
When the user pref config isn't saved, blender load one from memory and
call the function BLO_update_defaults_userpref_blend to initialize properly
the preferences such as the auto exec script option defaulted to true.

Unfortunatly this function is called after the check of the option
and so for blender auto exec is always false until the user saves the
preferences and reload.


panzergame:
sorry for the close…

panzergame:
Are you compiling with cmake option WITH_PYTHON_SECURITY=OFF ?

@youle31, @lordloki : For the next release binaries it is to take in consideration.

panzergame:
@lordloki It's added, thanks for the test.

panzergame:
@UnidayStudio Can you retry ?

PR #811, opened 2018-08-27
UPBGE: Add UPBGE version infos to bpy.app

youle31:
upbge_version
upbge_version_py
test file: http://pasteall.org/blend/index.php?id=50237

UPBGE version infos access from bpy (can be useful for addons)

Feature request from Momel on discord


panzergame:
Good work

Just untrack the release/ submodules if poosible.

PR #813, opened 2018-08-27
UPBGE: included Optimized GGX, Lambert and Burley BSDF shaders

lordloki:
Optimized GGX from http://filmicworlds.com/blog/optimizing-ggx-shaders-with-dotlh/ Optimization 1 (public domain)

Burley Diffuse from https://github.com/urho3d/Urho3D (MIT License)

New option under Material->Shading block to keep energy conservation.

Original code by @DCubix . Modified by me.

Test file: http://pasteall.org/blend/index.php?id=50244


panzergame:
Is it normal that the the reflectance property of `Lambert Custom BSDF` is linked to the one of `Optimized GGX BSFD` ?

Is it interesting to allow the user to set a 0 roughness ? If not you could clamp the value to 0.001.

GGX looks fine, Lambert has a strange behaviour concerning shadows.

@adriansnetlis, @martinsh : Could you look at it please ?

lordloki:
Teorically the reflectance in Lambert and roughness en Burley should be the same that in the GGX as the material is the same but we can make them independents to allow user customization.
Regarding "0" roughness, why not? Have you observed any firefly or something strange in the material?

adriansnetlis:
I haven't looked at it yet, but regarding your notes, panzergame:

- I believe the reflectance property should be linked. Only reason not to link it is for purely artistic reasons? Don't know any engine that does that.
- Roughness 0 should be an option (although not completely realistic). It won't work with current speculars, but, once the spherical area lights are merged, the specular vectors can be recalculated to get area light reflections. In that case 0 roughness works. Also, 0 roughness works once reflections are rendered on objects.

lordloki:
@adriansnetlis could you update the branch?, @panzergame committed a fix for that file.

adriansnetlis:
Built now and got this:
image

http://pasteall.org/1114170

lordloki:
make install

adriansnetlis:
Did do it. Didn't get any errors. It's something else.

youle31:
Hi, I did a small test file: http://pasteall.org/blend/index.php?id=50248

ggx

This sometimes gives interesting results, sometimes, it's a bit weird (0 reflectance makes the material black whereas it should affect only specular?, else sometimes there is a weird behaviour with diffuse). I don't really know if this is normal or not.

lordloki:
@youle31 additionally there is a new energy conservation parameter under shading that contributes a lot to making the materials look real.
image

youle31:
Great, I like the metallic suzanne when we remove sun shadows. I don't know for energy conservation, but it's true that we don't see it easily where it is. If you think it's an important option, maybe it could be more visible in the UI

lordloki:
Maybe, here
image

lordloki:
image
http://pasteall.org/blend/index.php?id=50252

Akira1San:
How does the texture mapping work with the new shading?

lordloki:
Exactly the same than before.

Akira1San:
Yeah, but in pbr shading you have slot for spec/gloss or metallic, but in bge material textures you dont.

Akira1San:
I tested it, gonna wait until there is a cubemap influence.

lordloki:
Cubemap influence is controlled through mirror texture slider. I will do some test to see if i can mimic pbr shader

BluePrintRandom:
I tried building this but we have a few confilcts with master

lordloki:
Here is a test build with last modifications (material extension node with new options):
https://mega.nz/#!wsFDAAAA!JorlNEWYWdDxAu7t8KizDlZpk0NSH3HggMdIOQRDfBw
image

Check energy conservation, select Burley and GGX BSDF shaders, add material extension node and play!!!!
Very very important Specular intensity (Col/Diff intensity could put to 1.0 and control everything with specular intensity)

lordloki:
More previews:
image

panzergame:
I have some hard edges with shadow, is it needed to modify the shadow shader code to be compatible with BSDF ?

screenshot_20181216_143453

lordloki:
are you referring to shadow projected or faces in shadow?

BluePrintRandom:
@martinsh - Do you think you could help us finish out SSR to use with this and IBL for fallback?
(this patch gives accurate diffuse, and spec, however it is missing reflect and we can't have a unique reflection probe per object. (as each object would then need a unique material)

we have render attachements now for screen space position buffer, and normal buffer,
I was thinking for the IBL we can use 'plot' in BPY and build the reflection texture atlas offline.

Issue #805, opened 2018-08-22
Libload/ Libfree have issues with multiple scenes.


RockyMadio:
If you execute libload and libfree on a single scene, it will work perfectly fine.
But if this scene is having an overlay scene OR being a background scene of another scene, after libfreeing the data, the game will crash right away.

You can reset the scene , but libload will not longer work. It says that library is still loaded, but nothing will appear.

I hope this issue will be set on high priority.
http://pasteall.org/blend/index.php?id=50213
http://pasteall.org/blend/index.php?id=50214

the test data has a great size for testing purpose.

RockyMadio:
S to load, and A to free

panzergame:
@RockyMadio : Could you test the branch ge_bug_fix_805 please ?

PR #824, opened 2018-09-06
UPBGE: Fix issues on LibFree/LibLoad.

panzergame:
The issue #805 reported a bug relative to LibLoad and LibFree. This bug is
reproduced by using two scenes which both use meshes without material (so using
defmaterial) and the first scene LibLoad and merge the second. What is happening
is that the new created scene will contain a RAS_MaterialBucket of defmaterial
which is also used by the host scene, but defmaterial is not owned by any file
as it's a global default material.
The technique used by LibFree was to tag all ID structs (data, e.g Mesh,
Material, Object) of a Main (the file loaded) and remove them from scenes and
users. In the situation of defmaterial, this struct is never tagged and so never
removed during LibFree. This resulted in a material always existing but all the
light objects used were removed and the function GPU_material_update_lamps
called in material preparation before rendering raise an illegal memory access
as it was iterating on freed lamps.

To solve this issue we have to find a way to remove all game engine data that
were created during the scene conversion of a libloaded scene. And the minor
task is avoiding end up with only a part of scene materials using all the lights
present.

The first point is solved by the introduction of a ressource base class:
BL_Ressource. The goal of this class is to store an identifier of the library
source of converted data. This identifier is an opaque intptr_t and the function
BL_Ressource::Belong(intptr_t) return true when the passed library identifier
match the one of the ressource.
The classes inheriting from BL_Ressource are:
- BL_ConvertObjectInfo
- KX_Mesh
- KX_BlenderMaterial
- BL_ActionData

This last class is used to wrap bAction and link the BL_ScalarInterpolator of
the action. All the actions are converted and registered during scene conversion
as the cost is negligeable.
All these ressource types are registered in BL_SceneConverter where all
registering function set the library identifier from the one passed to
BL_SceneConverter constructor. From LinkBlendFile, ConvertMeshSpecial and
ConvertScene(KX_Scene *scene) (the functions creating a scene converter)
BL_SceneConverter is created using as library identifier the Main pointer of the
current file or the loaded file, this way we also ensure that all linked data of
a file will be owned by the same library. Other than that the loading changed by
the call of the function ReloadShaders to reload all shaders of a merged scene:
the new and old shaders, to benefit of all existing lights.

Concerning LibFree, the first modification is to delay the free to the end of
the frame instead doing it instantly. To achieve it, FreeBlendFile(const
std::string& path) stores the library path to free in m_freeQueue if the library
exists and the function ProcessScheduledLibraries process the merging of the
libloads and the free of libraries, this final function is called in
KX_KetsjiEngine::NextFrame at the end of each logic frame.
The actual library free takes place in FreeBlendFileData. It iterates over all
scenes and objects, if the object convert object info belongs to the free
library the object is removed else KX_GameObject::RemoveRessources is called.
Finally the materials, meshes and actions of the library are removed from scene
data slots (m_sceneSlots). KX_GameObject::RemoveRessources is removing all the
meshes that belong to the library or using materials from the library, and do
the same for actions currently played.
Previously BL_ActionManager was also impacted by LibFree as it was using
bAction, now the actuator is using a action name to remove this complexity when
freeing an action.

Shader are now recompiled when merging and freeing a library to ensure that the
lights are properly used, the function BL_Covnerter::ReloadShaders is dedicated
for and call KX_BlenderMaterial::ReloadMaterial that can reload the shader or
create a new one if it wasn't initialized.

Fix issue: #805

### DIFF
--- a/source/gameengine/Converter/BL_Converter.cpp
+++ b/source/gameengine/Converter/BL_Converter.cpp
@@ -41,7 +41,7 @@
 #include "KX_KetsjiEngine.h"
 #include "KX_PythonInit.h" // So we can handle adding new text datablocks for Python to import
 #include "KX_LibLoadStatus.h"
-#include "BL_ScalarInterpolator.h"
+#include "BL_ActionData.h"
 #include "BL_Converter.h"
 #include "BL_SceneConverter.h"
 #include "BL_BlenderDataConversion.h"
@@ -77,6 +77,7 @@ extern "C" {
 #include "CM_Message.h"
 
 #include <cstring>
+#include <memory>
 
 BL_Converter::SceneSlot::SceneSlot() = default;
 
@@ -89,9 +90,6 @@ BL_Converter::SceneSlot::~SceneSlot() = default;
 
 void BL_Converter::SceneSlot::Merge(BL_Converter::SceneSlot& other)
 {
-	m_interpolators.insert(m_interpolators.begin(),
-	                       std::make_move_iterator(other.m_interpolators.begin()),
-	                       std::make_move_iterator(other.m_interpolators.end()));
 	m_materials.insert(m_materials.begin(),
 	                   std::make_move_iterator(other.m_materials.begin()),
 	                   std::make_move_iterator(other.m_materials.end()));
@@ -101,7 +99,9 @@ void BL_Converter::SceneSlot::Merge(BL_Converter::SceneSlot& other)
 	m_objectInfos.insert(m_objectInfos.begin(),
 	                     std::make_move_iterator(other.m_objectInfos.begin()),
 	                     std::make_move_iterator(other.m_objectInfos.end()));
-	m_actionToInterp.insert(other.m_actionToInterp.begin(), other.m_actionToInterp.end());
+	m_actions.insert(m_actions.begin(),
+					 std::make_move_iterator(other.m_actions.begin()),
+					 std::make_move_iterator(other.m_actions.end()));
 }
 
 void BL_Converter::SceneSlot::Merge(const BL_SceneConverter& converter)
@@ -115,6 +115,9 @@ void BL_Converter::SceneSlot::Merge(const BL_SceneConverter& converter)
 	for (BL_ConvertObjectInfo *info : converter.m_objectInfos) {
 		m_objectInfos.emplace_back(info);
 	}
+	for (BL_ActionData *action : converter.m_actions) {
+		m_actions.emplace_back(action);
+	}
 }
 
 BL_Converter::BL_Converter(Main *maggie, KX_KetsjiEngine *engine, bool alwaysUseExpandFraming, float camZoom)
@@ -125,16 +128,16 @@ BL_Converter::BL_Converter(Main *maggie, KX_KetsjiEngine *engine, bool alwaysUse
 {
 	BKE_main_id_tag_all(maggie, LIB_TAG_DOIT, false);  // avoid re-tagging later on
 	m_threadinfo.m_pool = BLI_task_pool_create(engine->GetTaskScheduler(), nullptr);
+
+	m_maggies.push_back(m_maggie);
 }
 
 BL_Converter::~BL_Converter()
 {
 	// free any data that was dynamically loaded
-	while (m_DynamicMaggie.size() != 0) {
-		FreeBlendFile(m_DynamicMaggie[0]);
+	while (!m_dynamicMaggies.empty()) {
+		FreeBlendFile(m_dynamicMaggies.front());
 	}
-
-	m_DynamicMaggie.clear();
 
 	/* Thread infos like mutex must be freed after FreeBlendFile function.
 	   Because it needs to lock the mutex, even if there's no active task when it's
@@ -144,15 +147,9 @@ BL_Converter::~BL_Converter()
 
 Scene *BL_Converter::GetBlenderSceneForName(const std::string &name)
 {
-	Scene *sce;
-
-	// Find the specified scene by name, or nullptr if nothing matches.
-	if ((sce = (Scene *)BLI_findstring(&m_maggie->scene, name.c_str(), offsetof(ID, name) + 2))) {
-		return sce;
-	}
-
-	for (Main *main : m_DynamicMaggie) {
-		if ((sce = (Scene *)BLI_findstring(&main->scene, name.c_str(), offsetof(ID, name) + 2))) {
+	for (Main *maggie : m_maggies) {
+		Scene *sce = (Scene *)BLI_findstring(&maggie->scene, name.c_str(), offsetof(ID, name) + 2);
+		if (sce) {
 			return sce;
 		}
 	}
@@ -176,7 +173,16 @@ EXP_ListValue<EXP_StringValue> *BL_Converter::GetInactiveSceneNames()
 	return list;
 }
 
-void BL_Converter::ConvertScene(BL_SceneConverter& converter, bool libloading)
+void BL_Converter::ConvertScene(KX_Scene *scene)
+{
+	BL_SceneConverter converter(scene, BL_Resource::Library(m_maggie));
+	ConvertScene(converter, false, true);
+	PostConvertScene(converter);
+	m_sceneSlots.emplace(scene, converter);
+	ReloadShaders(scene);
+}
+
+void BL_Converter::ConvertScene(BL_SceneConverter& converter, bool libloading, bool actions)
 {
 	KX_Scene *scene = converter.GetScene();
 
@@ -190,6 +196,11 @@ void BL_Converter::ConvertScene(BL_SceneConverter& converter, bool libloading)
 		m_alwaysUseExpandFraming,
 		m_camZoom,
 		libloading);
+
+	// Handle actions.
+	if (actions) {
+		BL_ConvertActions(scene, m_maggie, converter);
+	}
 }
 
 void BL_Converter::PostConvertScene(const BL_SceneConverter& converter)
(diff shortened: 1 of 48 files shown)


lordloki:
It gives me syntax error when I use LibLoad() function
image

source/gameengine/Converter/BL_Converter.cpp
```diff
@@ -374,270 +382,175 @@ KX_LibLoadStatus *BL_Converter::LinkBlendFile(BlendHandle *blendlib, const char
...
-		ID *action;
+	KX_LibLoadStatus *status = new KX_LibLoadStatus(this, m_ketsjiEngine, scene_merge, path);
 
-		for (action = (ID *)main_newlib->action.first; action; action = (ID *)action->next) {
-			if (options & LIB_LOAD_VERBOSE) {
-				CM_Debug("action name: " << action->name + 2);
+	switch (idcode) {
+		case ID_ME:
+		{
+			BL_SceneConverter sceneConverter(scene_merge, (uintptr_t)main_newlib);
```
panzergame: Note: create a `libraryId` instead repeating `(uintptr_t)main_newlib`.

source/gameengine/Converter/BL_Converter.cpp
```diff
@@ -646,36 +559,68 @@ bool BL_Converter::FreeBlendFile(Main *maggie)
...
+			else {
+				++it;
+			}
+		}
+
+		// Reload materials cause they used lamps removed now.
+		scene->GetBucketManager()->ReloadMaterials();
 	}
 
+	// Actual free of the lender library.
```
panzergame: lender -> blender

source/gameengine/Converter/BL_Converter.cpp
```diff
@@ -792,13 +731,12 @@ KX_Mesh *BL_Converter::ConvertMeshSpecial(KX_Scene *kx_scene, Main *maggie, cons
...
 
-	BL_SceneConverter sceneConverter(kx_scene);
+	BL_SceneConverter sceneConverter(kx_scene, (uintptr_t)maggie);
 
 	KX_Mesh *meshobj = BL_ConvertMesh((Mesh *)me, nullptr, kx_scene, sceneConverter);
-	kx_scene->GetLogicManager()->RegisterMeshName(meshobj->GetName(), meshobj);
 
 	MergeSceneData(kx_scene, sceneConverter);
-	FinalizeSceneData(sceneConverter);
+	ReloadShaders(kx_scene);
```
panzergame: Optimize to avoid material recompilation when using LibNew.

source/gameengine/Converter/BL_Converter.h
```diff
@@ -95,17 +95,31 @@ class BL_Converter
 		CM_ThreadMutex m_mutex;
 	} m_threadinfo;
 
-	// Saved KX_LibLoadStatus objects
-	std::map<std::string, KX_LibLoadStatus *> m_status_map;
 	std::vector<KX_LibLoadStatus *> m_mergequeue;
+	/// List of library name to free.
```
panzergame: name -> names

source/gameengine/Converter/BL_Converter.h
```diff
@@ -121,21 +135,28 @@ class BL_Converter
 	 */
 	void MergeScene(KX_Scene *to, const BL_SceneConverter& converter);
 
+	/** Regenerate material shader after a converting or merging a scene
+	 * depending on all the lights into the destination scene.
+	 */
+	void ReloadShaders(KX_Scene *scene);
+
+	void AddScenesToMergeQueue(KX_LibLoadStatus *status);
```
panzergame: Add comments.

source/gameengine/Converter/BL_Converter.h
```diff
@@ -146,30 +167,25 @@ class BL_Converter
...
 	/// Register a mesh object copy.
 	void RegisterMesh(KX_Scene *scene, KX_Mesh *mesh);
 
 	Scene *GetBlenderSceneForName(const std::string& name);
 	EXP_ListValue<EXP_StringValue> *GetInactiveSceneNames();
 
-	Main *CreateMainDynamic(const std::string& path);
-	Main *GetMainDynamicPath(const std::string& path) const;
-	const std::vector<Main *> &GetMainDynamic() const;
+	Main *CreateLibrary(const std::string& path);
```
panzergame: comments.

source/gameengine/Converter/BL_SceneConverter.cpp
```diff
@@ -91,6 +116,8 @@ KX_GameObject *BL_SceneConverter::FindGameObject(Object *for_blenderobject)
 
 void BL_SceneConverter::RegisterGameMesh(KX_Mesh *gamemesh, Mesh *for_blendermesh)
 {
+	gamemesh->m_libraryId = m_libraryId;
```
panzergame: Use a function: resource->SetLibraryId

source/gameengine/Converter/BL_SceneConverter.cpp
```diff
@@ -139,7 +174,10 @@ BL_ConvertObjectInfo *BL_SceneConverter::GetObjectInfo(Object *blenderobj)
 {
 	const auto& it = m_blenderToObjectInfos.find(blenderobj);
 	if (it == m_blenderToObjectInfos.end()) {
-		BL_ConvertObjectInfo *info = m_blenderToObjectInfos[blenderobj] = new BL_ConvertObjectInfo{blenderobj, {}};
+		BL_ConvertObjectInfo *info = m_blenderToObjectInfos[blenderobj] = new BL_ConvertObjectInfo();
```
panzergame: Back to a constructor.

source/gameengine/Converter/BL_SceneConverter.h
```diff
@@ -61,11 +63,16 @@ class BL_SceneConverter
 
 private:
 	KX_Scene *m_scene;
+	uintptr_t m_libraryId;
```
panzergame: Use const.

source/gameengine/GameLogic/SCA_LogicManager.cpp
```diff
@@ -220,40 +220,63 @@ void SCA_LogicManager::UpdateFrame(double curtime)
 
 void *SCA_LogicManager::GetActionByName(const std::string& actname)
 {
-	std::string an = actname;
-	return m_mapStringToActions[an];
-}
-
+	const auto it = m_mapStringToActions.find(actname);
```
panzergame: create a CM_MapFindNoInsert()

source/gameengine/GameLogic/SCA_LogicManager.cpp
```diff
@@ -220,40 +220,63 @@ void SCA_LogicManager::UpdateFrame(double curtime)
...
 void SCA_LogicManager::RegisterActionName(const std::string& actname, void *action)
 {
 	std::string an = actname;
 	m_mapStringToActions[an] = action;
 }
 
-
+void SCA_LogicManager::UnregisterAction(void *action)
+{
+	for (auto it = m_mapStringToActions.begin(); it != m_mapStringToActions.end();) {
```
panzergame: CM_MapRemove.

source/gameengine/Ketsji/KX_GameObject.cpp
```diff
@@ -593,6 +589,28 @@ EXP_Value *KX_GameObject::GetReplica()
 	return replica;
 }
 
+void KX_GameObject::RemoveRessources(uintptr_t libraryId)
+{
```
panzergame: Add comments.

source/gameengine/Ketsji/KX_GameObject.h
```diff
@@ -368,6 +363,8 @@ class KX_GameObject : public SCA_IObject, public mt::SimdClassAllocator
 	 * object belongs with the caller.
 	 */
 	virtual EXP_Value *GetReplica();
+
```
panzergame: Add comments.

source/gameengine/Rasterizer/RAS_BucketManager.cpp
```diff
@@ -378,22 +378,21 @@ void RAS_BucketManager::ReloadMaterials(RAS_IMaterial *mat)
...
-				if (i == ALL_BUCKET) {
-					delete bucket;
-				}
-			}
-			else {
-				++it;
+	BucketList& allBuckets = m_buckets[ALL_BUCKET];
+	for (BucketList::iterator it = allBuckets.begin(); it != allBuckets.end();) {
+		RAS_MaterialBucket *bucket = *it;
+		if (mat == bucket->GetMaterial()) {
```
panzergame: comment

panzergame:
TODO: Replace intptr_t by an alias and update commit message.

lordloki:
Checking libload blends from Test repository:

- Load.blend:
I have crash at mainptr->curlib.

image

With this patch it works:
```
--- source/blender/blenloader/intern/readfile.c
+++ source/blender/blenloader/intern/readfile.c
@@ -10737,7 +10737,7 @@ static void read_libraries(FileData *basefd, ListBase *mainlist)
 			/* We need to split out IDs already existing, or they will go again through do_versions - bad, very bad! */
 			split_main_newid(mainptr, &main_newid);
 
-			if (mainptr->curlib->filedata) // can be zero... with shift+f1 append
+			if (mainptr->curlib && mainptr->curlib->filedata) // can be zero... with shift+f1 append
 				do_versions(mainptr->curlib->filedata, mainptr->curlib, &main_newid);
 			else
 				do_versions(basefd, NULL, &main_newid);
@@ -10745,12 +10745,14 @@ static void read_libraries(FileData *basefd, ListBase *mainlist)
 			add_main_to_main(mainptr, &main_newid);
 		}
 		
+		if (mainptr->curlib) {
 			if (mainptr->curlib->filedata)
 				lib_link_all(mainptr->curlib->filedata, mainptr);
 	
 			if (mainptr->curlib->filedata) blo_freefiledata(mainptr->curlib->filedata);
 			mainptr->curlib->filedata = NULL;
 		}
+	}
 }
```
But then I have crash at exit

image

- loadDestScene.blend: Same crash than Load.blend

- ComponentLibLoad.blend: It works (although it seems that the first time is slower to load the objects that the next few times)

- MirrorLibLoad.blend: It partially works. Load mirror.blend, when you press Q -> LibFree but after we have crash at exit

image

- MainUvTest.blend: It works

- animbugLoad.blend: It works

- Lighting: It works

- FontLinkedGroupExample: It works

lordloki:
I updated the above comentary with new results after last commits.

lordloki:
With last commits all the libload blend test files works like a charm. Good Work!!!

lordloki:
New thing:
- animbugLoad.blend:
Only pressing 'Z' and 'V' alternatively. Something you get correct lighting and sometimes you get the cube shadeless
image
image

panzergame:
I cannot reproduce the bug in animbugLoad.blend...

panzergame:
I did a test with valgrind without error report...

panzergame:
The second issue is fixed and commited, thanks for its report.

PR #827, opened 2018-09-09
UPBGE: Rename async parameter from LibLoad as is a reserved word

### LINKED ISSUE #826: LibLoad async parameter issue with Python 3.7
async is a reserved word in python 3.7 and it can not passed to LibLoad function without error:
image

Related info: https://docs.python.org/3.7/whatsnew/3.7.html#summary-release-highlights

lordloki:
async is a reserved word in Pyhon 3.7
Check: https://docs.python.org/3.7/whatsnew/3.7.html#summary-release-highlights

This Fix #826 .

The python rst modification should be commited here and UPBGE-Docs repository or only here or only there?

### DIFF
--- a/doc/python_api/rst/bge.logic.rst
+++ b/doc/python_api/rst/bge.logic.rst
@@ -185,7 +185,7 @@ General functions
 
    Restarts the current game by reloading the .blend file (the last saved version, not what is currently running).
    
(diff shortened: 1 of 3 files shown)


panzergame:
The UPBGE-Docs rst (for API) is a copy of the one in main repo.

panzergame:
I think that's `asynchronous` instead of `asyncronous`.

Issue #820, opened 2018-09-02
Vsync python


gionnigio:
i dont really know if it is bug or its intentional but:

bge.render.setVsync(bge.render.VSYNC_ON)  
bge.render.getVsync() 
returns 1

bge.render.setVsync(bge.render.VSYNC_OFF)
bge.render.getVsync() 
returns 0

bge.render.setVsync(bge.render.VSYNC_ADAPTIVE)
bge.render.getVsync()
returns 0 as vsyncoff

upbge 0.2.3

youle31:
hi, I don't really know what is swap interval but if there is a possibility to display a warning when "adaptative vsync" is not supported, it can be useful i guess too. Else, to be honest, i'm not familiar with vsync...

PR #834, opened 2018-09-15
UPBGE: Fix bge.render.[get/set]Vsync.

panzergame:
Previously bge.render.getVsync wasn't returning the value set by setVsync when
VSYNC_ADAPTIVE was used. It can be justified by the fact that adaptive is set
by using a negative swap interval value but the GLX get function still returns
a positive swap interval value and the flag GLX_LATE_SWAPS_TEAR_EXT meaning
adaptive is enabled.

In any case GLX could refuse a swap interval and the user should be able to do
setVsync(getVsync()) with no change on the swap control. To achieve this behaviour
RAS_ICanvas store the current swap control in a enum variable m_controlSwap,
this variable is used in getter GetSwapControl used by bge.render.getVsync().
The setter is virtual and overloaded in each canvas implementation, they use
a common table to find the swap interval associated to the swap control mode:
RAS_ICanvas::swapInterval.

Finally KX_PythonInit is using swap control enum value from RAS_ICanvas instead
of the one from DNA to avoid conversions.

Fix issue #820.

### DIFF
--- a/source/blender/makesdna/DNA_scene_types.h
+++ b/source/blender/makesdna/DNA_scene_types.h
@@ -900,9 +900,11 @@ typedef struct GameData {
 #define OBSTSIMULATION_TOI_cells	2
 
 /* GameData.vsync */
-#define VSYNC_ON	0
-#define VSYNC_OFF	1
-#define VSYNC_ADAPTIVE	2
+enum {
+	VSYNC_ON = 0,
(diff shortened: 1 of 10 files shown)


panzergame:
Example file: 
[ge_820.zip](https://github.com/UPBGE/blender/files/2385592/ge_820.zip)

Look at console prints.

PR #840, opened 2018-09-21
UPBGE: Set frame step to elapsed time in case of external clock.

panzergame:
Previously the time step (base of frame step) was set to the
fixed frame step (1 / FPS * scale) when the user was using external
clock (setting the clock manually).

This commit introduces a different behaviour: the timestep is the
elapsed time bewteen two clock set of the user.


marechal-p:
This change seems to be a nice idea, but I was wondering if there is any documentation on how to setup our own clock for the BGE?

I saw it mentionned in a few places but couldn't figure out how to set it up...

Issue #822, opened 2018-09-05
problem passing array of uniforms to custom shaders


youle31:
Since 18e5a965308929ff14218b969b2c01aed16c5d95 there is an issue to pass array of uniforms to custom shaders

example file: http://pasteall.org/blend/index.php?id=50286

When we revert 32a6eaa7ace458b660ae8f9a72cf8be9ef25b62e then 18e5a965308929ff14218b969b2c01aed16c5d95 it works again

(was trying to help Rocky Madio to explain him how to pass arrays to shaders)

youle31:
There is some reports about that on the net. glGetActiveUniform doesn't seems handle the case. We can use the old method GPU_shader_get_uniform to get the location if the name ends with "]" for example...
(if (name.compare(name.size() - 1, 1, "]") == 0) { return GPU_shader_get_uniform(m_shader, name.c_str()); } )
or maybe there is a better way...

PR #844, opened 2018-09-24
UPBGE: Fix uniform caching of arrays.

panzergame:
Since uniform caching the user was unable to precise the name of
an uniform array item such as colors[42].

This is caused by an ambiguity in the OpenGL function glGetActiveUniform
regarding arrays. If the arrays is of struct type then all the items are
returned by glGetActiveUniform. Else for simple array in main program
of in struct but of primitive type, then only the first element is
returned e.g colors[0].

Two choices were proposed:
- Detecting the array name format and discard caching mechanism while
asking for a location.
- Simulating all item names and store all theirs locations.

The second option is selected in this commit for multiple advantages
over the first:
- Detecting the format of an array name could be complicated as the
program have to manage the case the user is provinding syntaxically
erronous name.
- The current cache is currently getting the location manually for
active uniforms, doing it for arrays is in average not a big amount
of extra work.
- GLSL is limiting big native arrays (rarely up to 100) which limit
the number of locations to retrieve.
- GLSL allow getting the location for "colors" which is an ignored
alias for "colors[0]" by glGetActiveUniform, so this case needed
to be taken in account to ignore caching.

In consideration the caching look up it optimized by using a sorted
vector and a binary search. std::unordered_map is doing an similar
work but do more like caching the uniform name where for searching
only the name hash matter.

This is applied in RAS_Shader::ExtractUniformInfos. If an active
uniform is suppoed array from OpenGL information, the base name is
generated (e.g "colors") and then all the names from indices
(e.g "colors[1]", "colors[2]"). Non-array uniforms are proceeded
as before.
UniformInfo is constructed using the name and shader, the constructor
hash the name and obtain the uniform location.
Then the uniforms are sorted using the name hash value in UniformInfo.

RAS_Shader::GetUniformLocation now uses std::lower_bound to process
a binary search.

Fix issue: #822.

### DIFF
--- a/source/blender/gpu/GPU_shader.h
+++ b/source/blender/gpu/GPU_shader.h
@@ -75,7 +75,6 @@ int GPU_shader_program(GPUShader *shader);
 
 typedef struct GPUUniformInfo
 {
-	unsigned int location;
 	unsigned int size;
 	unsigned int type;
 	char name[255];
--- a/source/blender/gpu/intern/gpu_shader.c
+++ b/source/blender/gpu/intern/gpu_shader.c
@@ -636,7 +636,6 @@ int GPU_shader_get_uniform_infos(GPUShader *shader, GPUUniformInfo **infos)
 	for (unsigned int i = 0; i < count; ++i) {
 		GPUUniformInfo *info = &((*infos)[i]);
 		glGetActiveUniform(shader->program, i, 255, NULL, (int *)&info->size, &info->type, info->name);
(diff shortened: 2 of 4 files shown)


youle31:
hi, I didn't tested yet but had a quick look at the code and it sounds good to me

youle31:
It is working in both of these files:

http://pasteall.org/blend/index.php?id=50286
http://pasteall.org/blend/index.php?id=50400

so it seems allright

PR #852, opened 2018-09-30
UPBGE: Fix and simplify frame time management.

panzergame:
Previously a commit introduce for external clock the targeting of one frame
at every call to NextFrame, not more.
The original behaviour is if the user enabled fixed framerate then compute
as many as frame possible respecting the FPS value (m_ticrate).
This behaviour is restored in the refactor.

The refactor is simplyfing the time calculation and isolating it.
The function KX_KetsjiEngine::GetFrameTimes is returning a struct FrameTimes
with the number of frames to proceed and the time/frame step.
Inside this function we first update the clock time if the user isn't (case
of external clock) then the elapsed time is computed. Next the timestep and
frame number are computed using fixed value or elapsed time. The number of frame
is clamped by max logic/physics frame using the simple method: If the
number of frame is higher than the maximum for logic or physics, then clamp
it and set timestep to the elasped time (which is always greater than fixed
time step 1/FPS) divided by the maximum number of frames.
Finally if the user is not using fixed framerate update the previous clock
else if the number of frames to proceed is zero sleep.
Issue were noticed using a sleep, they were caused by the resolution of the
sleep which is about some ~10us, to fix this a part of the sleep is consumed
with std::this_thread::sleep_for and the rest is busy wait.

CM_Clock has now a function Reset to ensure it starts from 0, this is useful
for external clock starting at 0.

### DIFF
--- a/source/gameengine/Ketsji/KX_KetsjiEngine.cpp
+++ b/source/gameengine/Ketsji/KX_KetsjiEngine.cpp
@@ -245,7 +245,8 @@ void KX_KetsjiEngine::SetConverter(BL_Converter *converter)
 
 void KX_KetsjiEngine::StartEngine()
 {
-	m_previousRealTime = m_clock.GetTimeSecond();
+	// Reset the clock to start at 0.0.
+	m_clock.Reset();
 
 	m_bInitialized = true;
 }
@@ -320,10 +321,8 @@ void KX_KetsjiEngine::EndFrame()
 	m_canvas->EndDraw();
 }
 
-bool KX_KetsjiEngine::NextFrame()
+KX_KetsjiEngine::FrameTimes KX_KetsjiEngine::GetFrameTimes()
 {
-	m_logger.StartLog(tc_services);
-
 	/*
 	 * Clock advancement. There is basically two case:
 	 *   - USE_EXTERNAL_CLOCK is true, the user is responsible to advance the time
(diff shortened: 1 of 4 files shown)


source/gameengine/Ketsji/KX_KetsjiEngine.cpp
```diff
@@ -341,73 +340,82 @@ bool KX_KetsjiEngine::NextFrame()
...
-	for (unsigned short i = 0; i < frames; ++i) {
-		m_frameTime += framestep;
+	// If the number of frame is non-zero, update previous time.
+	if (frames > 0) {
+		m_previousRealTime = m_clockTime;
+	}
+	// Else in case of fixed framerate, try to sleep until the next frame.
+	else if (m_flags & FIXED_FRAMERATE) {
+		const double sleeptime = timestep - dt - 1.0e-3;
+		/* If the remaining time is greather than 100us (sleep resolution) sleep this thread.
```
marechal-p: Should the comment be updated accordingly? 100us => 1ms

youle31:
hi, it's weird I did screenshots in master and in ge_frame_time and this doesn't seems to work as expected (CPU usage should be < in ge_frame_time because of the thread? (when use framerate and vsync are enabled))
I took screenshots:

ge_frame_time:
ge_frame_time
Master:
master

Else, in ge_frame_time, the framerate seems a bit more stable whereas in Master it seems to stutter a bit (but the gameplay does't stutter in this file, just the framerate display)

youle31:
EDIT!!! Ah no it's ok I have the same low CPU usage (wrong merge with git) and a framerate at 59,4, or 5 or 6 or 7

PR #861, opened 2018-10-03
UPBGE: Fix object render settings for batching groups.

panzergame:
Previously object render settings such as light layer or object color
were not managed for batching groups.

For attribute like layers we can choose to make an union of all objects
layers, but in case of color there's no meaning of union.

To solve this issue one object should be used as reference for the group.
The attribute KX_BatchGroup.referenceObject is defined and initialized
to the first merged object when creating a batch group.
Later the user can change it to an other object of the same group.

Internally this object is set in KX_BatchGroup and RAS_BatchGroup,
KX_BatchGroup see a KX_GameObject and RAS_BatchGroup only the RAS_MeshUser
of this object. This mesh user is retrieved in RAS_DisplayArrayBucket::RunBatchingNode
from the first mesh slot mesh user and passed to RAS_IMaerial::ActivateMeshUser
which aims the update of material uniforms from object settings.

### DIFF
--- a/doc/python_api/rst/bge_types/bge.types.KX_BatchGroup.rst
+++ b/doc/python_api/rst/bge_types/bge.types.KX_BatchGroup.rst
@@ -19,11 +19,20 @@ base class --- :class:`EXP_Value`
 
       batchGroup = types.KX_BatchGroup([scene.objects["Cube"], scene.objects["Plane"]])
 
+   .. warning::
+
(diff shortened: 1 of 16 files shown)


panzergame:
Do we target this patch for the beta release ?
It is a bug fix but by introducing a new concept.

BluePrintRandom:
I think so - it's important and the default behavior is kinda broken
(if master is 100% stable before beta)

lordloki:
Yes, we can add to this version

Issue #838, opened 2018-09-17
crash with a game (maybe related to bullet)


youle31:
![bububullet](https://user-images.githubusercontent.com/10052509/45649180-e3a1ec00-baca-11e8-8e36-8daff90a51a1.jpg)

test file:

https://www.dropbox.com/s/w2usg4m4wos42ra/Colladiration.blend?dl=0

steps: Press P and wait 0.5 seconds

lordloki:
Maybe #define BT_THREADSAFE 1 
Will help with the crash.
See
https://pybullet.org/Bullet/phpBB3/viewtopic.php?t=10232&start=105#p38432

Akira1San:
It dont crash with branch UPBGE_v0.2.3_ge_physics_Windows64.zip

youle31:
list of files causing issues related to bullet:

- On discord: https://cdn.discordapp.com/attachments/290595893931933696/495430775987109898/dum_game.blend
(go to level 1 scene, play, crash when you have all coins - simple file)
- The Akira san's file: https://www.dropbox.com/s/w2usg4m4wos42ra/Colladiration.blend?dl=0 (same as the 1 shared at the top of the thread)
- BGMC 22:
Horsey adventures -> crash at exit (press ESC)
Terminus north: crash at exit (press ESC)

I'll edit if I find others

panzergame:
@youle31 Can you please reconfirm that my commit fixed the issues please ?

youle31:
Fixed in https://github.com/UPBGE/blender/commit/8d97babd888749a6490c3aabdbcdd57f5f80919e Thanks @panzergame !

PR #866, opened 2018-10-04
UPBGE: Fix character collision detection.

panzergame:
Chracter controller is using a ghost object and in particular
a ghost object caching collision pairs named btPairCachingGhostObject.
But with the recent update of bullet, we realized after some bug
reported around character that the caching of collision is not
anymore supported in a safe way.

Instead of fixing the caching mechanism, it was choose to replace
btPairCachingGhostObject by btGhostObject.

The btGhostObject is already maintaining a cache of object overlapping
the only step added is asking the pair cache in the world dispatcher
for the character ghost and one of its overlapping objects.
By this way the character is just reading and not managing duplication
or deletion of manifold in its own cache which is the safest solution.

In the same time the function btSequentialImpulseConstraintSolver::getOrInitSolverBody
is not handling ghost object but only soft, rigid, kinematic and multibody.
An assert is disabled with the adjonction of ghost type check.

Fix issue #838.

### DIFF
--- a/extern/bullet/src/BulletDynamics/Character/btKinematicCharacterController.cpp
+++ b/extern/bullet/src/BulletDynamics/Character/btKinematicCharacterController.cpp
@@ -132,7 +132,7 @@ btVector3 btKinematicCharacterController::perpindicularComponent (const btVector
 	return direction - parallelComponent(direction, normal);
 }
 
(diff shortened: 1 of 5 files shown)


lordloki:
Changes seem OK to me. Additionally, you committed changes for unrelated files.

Issue #797, opened 2018-08-19
Windows x86 UPBGE crashing at start


joelgomes1994:
Hello! Windows build process is successful at both x86 and x64, but the x86 build won't start, printing only `EXCEPTION_ACCESS_VIOLATION` on the console when Blender starts. I tried to rebuild from scratch, but the problem persists.

I will try to revert some commits to see where this problem starts. In the meantime, could someone confirm the issue? Thanks.

lordloki:
Confirmed.
Crash inside of gpu_nodes_tag function.
![image](https://user-images.githubusercontent.com/1015605/44320650-9edf5280-a443-11e8-84c0-de27946f6eee.png)

panzergame:
@lordloki : What kind of file are you using ?

lordloki:
Default cube

lordloki:
Joel, can you test this build? https://mega.nz/#!cttHlRiY!NNf4b8NO9k7tFUySVdP4K39Zd6iGUmupDG8qokfNrG4

It is from branch ge_bug_fix_797.
For my it is working

joelgomes1994:
@lordloki I just built the branch you said, it's working for me as well.

PR #869, opened 2018-10-04
UPBGE: Fix for issue #797

lordloki:
Set GPU_INT to 5 avoid buffer overrun

This is due to the following:
in gpu_codegen.c in gpu_node_input_link we have the following catch all

```c
else {
		/* uniform vector */
		input->type = type;
		input->source = GPU_SOURCE_VEC_UNIFORM;

		memcpy(input->vec, link->ptr1, type * sizeof(float)); <<---- for int = 17 * float
		if (link->dynamic) {
			input->dynamicvec = link->ptr1;
			input->dynamictype = link->dynamictype;
			input->dynamicdata = link->ptr2;
		}
		MEM_freeN(link);
	}
```

and in the typedef for GPUInput we have this for vec

```c
float vec[16];               /* vector data */
GPUNodeLink *link;
```

As the type for INT was 17, this will copy 17 floats from link->ptr1 overwriting input->link with garbage.


source/blender/gpu/intern/gpu_codegen.c
```diff
@@ -1297,7 +1297,13 @@ static void gpu_node_input_link(GPUNode *node, GPUNodeLink *link, const GPUType
 		input->type = type;
 		input->source = GPU_SOURCE_VEC_UNIFORM;
 
-		memcpy(input->vec, link->ptr1, type * sizeof(float));
+		if (type == GPU_INT) {
```
lordloki: We can even make this GPU_INT check only and left all above changes as they are in current master

panzergame: What about using an array for size of types indexed by GPUType ? So you can change the values and avoid exceptions in the code when adding conditions.

lordloki: I'm not convinced to do that before final release. I would prefer fix memcpy and make the tiny refactor after

panzergame: As you want, but the refactor is just defining an array of 18 elements.

source/blender/gpu/intern/gpu_codegen.c
```diff
@@ -55,6 +55,11 @@
 #include <string.h>
 #include <stdarg.h>
 
+#ifndef ARRAYSIZE
```
panzergame: Is it possible to include `BLI_utildefines.h` for this macro ?

lordloki: Yes, done

PR #887, opened 2018-10-13
UPBGE: Expose the number of blur passes for variance shadow maps

lordloki:
The default value is 1 (as current default) but the recommended value for quality/real shadows is 2 or 3.
Max 4 passes. To use with care because it is expensive. 
lamp_blur_passes

### DIFF
--- a/source/blender/makesrna/intern/rna_lamp.c
+++ b/source/blender/makesrna/intern/rna_lamp.c
@@ -653,6 +653,12 @@ static void rna_def_lamp_shadow(StructRNA *srna, int spot, int area)
 	RNA_def_property_ui_text(prop, "Samples", "Number of shadow buffer samples");
 	RNA_def_property_update(prop, 0, "rna_Lamp_update");
 
+	prop = RNA_def_property(srna, "shadow_blur_passes", PROP_INT, PROP_NONE);
+	RNA_def_property_int_sdna(prop, NULL, "blurpass");
+	RNA_def_property_range(prop, 0, 4);
+	RNA_def_property_ui_text(prop, "Shadow Buffer Size", "Number of shadow buffer blur pass (is expensive)");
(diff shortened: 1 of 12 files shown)


source/blender/makesrna/intern/rna_lamp.c
```diff
@@ -653,6 +653,12 @@ static void rna_def_lamp_shadow(StructRNA *srna, int spot, int area)
 	RNA_def_property_ui_text(prop, "Samples", "Number of shadow buffer samples");
 	RNA_def_property_update(prop, 0, "rna_Lamp_update");
 
+	prop = RNA_def_property(srna, "shadow_blur_passes", PROP_INT, PROP_NONE);
+	RNA_def_property_int_sdna(prop, NULL, "blurpass");
+	RNA_def_property_range(prop, 0, 4);
+	RNA_def_property_ui_text(prop, "Shadow Buffer Size", "Number of shadow buffer blur pass (is expensive)");
```
lordloki: Shadow buffer size --> Blur Passes

panzergame:
What do you think about passing the number of passes to GPU_framebuffer_blur and factorize some operations such as shader binding ?


[diff.txt](https://github.com/UPBGE/blender/files/2479481/diff.txt)

lordloki:
I was thinking about it. I tested your patch and it gives slightly more quality (2 passes same quality than older 3 passes) but it cost a lot more (2 passes cost equal than older 4 passes).

panzergame:
I tested recently my patch and didn't see improvement of the quality neither a higher performance cost.

I checked the number of passes (vertical blur and horizontal blur) in apitrace and found the wanted number of passes.

lordloki:
I will test again as your solution is better. I also found my results very strange

lordloki:
Now it works OK with your code (pushed). Maybe I pasted wrong your code the last time.

PR #891, opened 2018-10-13
UPBGE: Fix color space of ImageRender textures.

panzergame:
With the introduction of color management, every scene render is
in linear color space and the final draw to the screen (potentially
after filters) might be in sRGB space.

This behvaiour has the side effect of leaving ImageRender/ImageViewport
images in linear spaces as they result from a scene render. But all
blender textures were before supposed to be sRGB space and converted
back into linear in material shaders. Now it is partially the case.

The way of fixing this issue is to leave an option in material texture
panel to select color space of the image. The default is sRGB and
texture sapped with ImageRender in game must select Linear space.

Realtime envmap always ignore this setting as they are in linear
space in any cases.

### DIFF
--- a/release/scripts/startup/bl_ui/properties_data_empty.py
+++ b/release/scripts/startup/bl_ui/properties_data_empty.py
@@ -43,7 +43,7 @@ class DATA_PT_empty(DataButtonsPanel, Panel):
 
         if ob.empty_draw_type == 'IMAGE':
             layout.template_ID(ob, "data", open="image.open", unlink="object.unlink_data")
-            layout.template_image(ob, "data", ob.image_user, compact=True)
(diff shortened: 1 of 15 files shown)


panzergame:
Test file:
[ge_texture_color_space.zip](https://github.com/UPBGE/blender/files/2475765/ge_texture_color_space.zip)

The texture is using linear space. If you select sRGB which was the default, you can see the color error.

panzergame:
The option "color space" in image panel in texture tab is disabled but not in UV editor.

screenshot_20181018_134357

DaedalusMDW:
i think this broke the texture node?  setting the image to "Raw" appears to fix the color space, but then breaks again when launched from an exe.

DaedalusMDW:
colors_blend_u024
colors_exe_u024

PR #907, opened 2018-10-27
UPBGE: Fix for mouse actuator return value 

lordloki:
This occurred under certain situation (i.e fps < 30). Now it run at
first time to relocate the mouse only.


lordloki:
@panzergame the problem was that when the fps is low (due to high game load) we have a lot Negative events and therefore the mouse cursor is always reseting avoiding any possible movement.

I have removed any action under negative event. This solves mouse movement under any fps.

Additionally, I fixed the return value. This avoided the use of mouse actuator with always sensor.

lordloki:
I changed the pull request. Now it fixes the return value and it works perfectly again when activating from an always sensor.

Additionally, it introduces a boolean check for initial skipping instead the modifying the oldposition.

PR #919, opened 2018-11-03
UPBGE: Python controllers module loading fix

### LINKED ISSUE #918: Binary Python modules are not re-imported
Hello.

After you stop game engine and start it again, all text Python modules are re-loaded.
So if I made some changes to code, new actual code will be launched.

But this doesn't apply on binary Python modules, they aren't reloaded.
Even if you compiled a new version, you need to close and re-open Blender itself to see actual changes.

Only if you import that module from some python code, you can make `imp.reload(binary_module_name)`, 
but if you specified binary module name in Python controller directly - you can't reload it until Blender restart.

Here video with steps to reproduce:
https://youtu.be/EYFQExAyjLE

(I wrote binary module in Nim there, but same applies if you'll write it in Cython, for example)

Hope, it can be fixed easily =)

Best regards,
Vladislav

marechal-p:
When doing `pyctrl->Import()` we should reload a module if it was already
loaded in the Python runtime. The call to `Import()` should only be done
to update the used script/module anyway, and shouldn't be called
repeatedly, unless you want to refresh the Python program continuously.
This is already managed by `SCA_PythonController::Trigger()` by watching
the `m_bModified` boolean.

Fix #918.

Signed-off-by: marechal-p <marechap.info@gmail.com>

### DIFF
--- a/source/gameengine/GameLogic/SCA_PythonController.cpp
+++ b/source/gameengine/GameLogic/SCA_PythonController.cpp
@@ -281,21 +281,27 @@ bool SCA_PythonController::Import()
 		return false;
 	}
 
-	// Import the module and print an error if it's not found
-	PyObject *mod = PyImport_ImportModule(mod_path.c_str());
+	// Try to get the module by name
+	PyObject *mod = PyImport_GetModule(mod_path.c_str());
 
 	if (mod == nullptr) {
-		ErrorPrint("Python module can't be imported");
-		return false;
-	}
 
-	if (m_debug) {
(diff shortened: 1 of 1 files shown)


marechal-p:
@panzergame made me realize that while this will work for the first level of imported module (the one directly called by the GE), it will work, but not for any nested import.

It seems to come from the way Python itself manages such libraries.

One lead to follow would be to not use Python's default import system but to write and use a function that would do the import/reload correctly, automatically. See https://github.com/UPBGE/blender/issues/918#issuecomment-437241351.

PR #956, opened 2018-12-01
UPBGE: Code refactor to work with opengl compatibility mode in MacOs

lordloki:
It includes:
- Only set gl_ClipDistance for ATI cards
- Using compatible functions and input/outputs for glsl 1.2
- Use APPLE_vertex_array_object extension for MacosX
- Unset OpenColorIO and OpenSubDiv as they provoque freezes
- New GPU functions to manage VAOs
- Use #ifdef instead of calling function continuosly

After commit  it would only left to merge current BF master (they update mac build system) to work OK.


source/blender/gpu/GPU_vertex.h
```diff
@@ -0,0 +1,46 @@
...
+#define __GPU_VERTEX_H__
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/* GPU Vertex
+ * - only to make accesible VAOs for MacOS when we are in compatibility profile */
+
+void GPU_gen_vertex_arrays(int n, unsigned int *arrays);
```
panzergame: If have a small question about the name, should we follow the opengl name as you did here or mimic the name of other function such as in GPU_texture.h which use the name "create" instead of "gen" ?

source/blender/gpu/CMakeLists.txt
```diff
@@ -105,6 +106,7 @@ set(SRC
 	GPU_select.h
 	GPU_shader.h
 	GPU_texture.h
+	GPU_vertex.h
```
panzergame: Is gpu_vertex not ambiguous with vertex while it targets vertex array object ?

intern/opensubdiv/gpu_shader_opensubdiv_geometry.glsl
```diff
@@ -47,6 +47,10 @@ layout(triangle_strip, max_vertices = 4) out;
   #extension GL_EXT_texture_buffer_object: enable
 #endif
 
+#if __VERSION__ < 130
```
panzergame: Is this modification useful if OpenSubdiv is disabled ?

source/blender/gpu/intern/gpu_shader.c
```diff
@@ -232,6 +232,9 @@ static void gpu_shader_standard_extensions(char defines[MAX_EXT_DEFINE_LENGTH],
 			/* TODO: maybe require this? shaders become so much nicer */
 		}
 	}
+	if (GPU_type_matches(GPU_DEVICE_ANY, GPU_OS_MAC, GPU_DRIVER_ANY) && GPU_legacy_support()) {
```
panzergame: Is this extension needed to be explicitly enabled for shaders ?

source/blender/gpu/shaders/gpu_shader_vertex.glsl
```diff
@@ -8,21 +9,24 @@ out block {
...
-in mat3 ininstmatrix;
-in vec3 ininstposition;
-in vec4 ininstcolor;
-in int ininstlayer;
-
-out vec4 varinstcolor;
-out mat4 varinstmat;
-out mat4 varinstinvmat;
-flat out int varinstlayer;
+/* compatibility / in */
```
panzergame: I prefer keep using your nice workaround with macro:
```
#if __VERSION__ < 130
  #define in varying
  #define out varying
#endif
```

But you can move it into gpu_shader_lib.glsl which is included in all shader generate with function GPU_shader_create_ex (so all of them).

The same can apply with workaround of `texelFetch`

source/blender/gpu/shaders/gpu_shader_vertex_world.glsl
```diff
@@ -79,7 +81,14 @@ void main()
 	varposition = gl_Vertex.xyz;
 
 	varnormal = normalize(-varposition);
+
+#ifdef CLIP_WORKAROUND
 	// Always set clip distance to 1 to disable clipping.
 	for (int i = 0; i < 6; ++i) {
 		gl_ClipDistance[i] = 1.0;
```
panzergame: This is to disable clipping for world (background rendering) did you test a scene with planar reflection and a complex world ?

youle31:
Hi, I can't test unfortunately. I have no special remark about the code. If you have the time to provide a build to Claudio and that he gives good feedbacks, it can help. Also, I think @UnidayStudio had friends on mac who could give feedbacks if a build is shared

youle31:
And thanks very much for your work @lordloki !

lordloki:
Thanks @panzergame and @youle for reviewing. It is not necessary to test as i tested before in a macos machine

source/blender/gpu/intern/gpu_shader.c
lordloki: Not necessary. I will remove

source/blender/gpu/GPU_vertex.h
lordloki: Ok. For GPU_vertex is better GPU_vertex_array, GPU_vertex_array_object or maybe GPU_vao?

intern/opensubdiv/gpu_shader_opensubdiv_geometry.glsl
lordloki: I have to test if it compiles ok without it

source/blender/gpu/GPU_vertex.h
panzergame: i'm fine with GPU_vertex_array

source/blender/gpu/shaders/gpu_shader_vertex.glsl
panzergame: It not possible to pass a varying to a local function in GLSL or it's a problem with gpu_codegen ?

lordloki: The issue is that the "out" or "in" parameter type in a function is diferent from from in/out or varying type qualifier. 

```
GLSL functions are declared and defined similarly to C/C++ functions. A function declaration in GLSL looks like this:

void MyFunction(in float inputValue, out int outputValue, inout float inAndOutValue);
Functions in GLSL use a calling convention called "value-return." This means that values passed to functions are copied into parameters when the function is called, and outputs are copied out when the function returns.

The in, out, and inout qualifiers are not the same as type qualifiers, even though some of them are named the same. These are parameter qualifiers, and they have a different meaning here.
```

source/blender/gpu/shaders/gpu_shader_vertex_world.glsl
lordloki: Yes, I just tested it and it works wrong.
image
I'm looking for a solution.

lordloki: Fix commited
image

lordloki:
I need to compile and test in Mac after last changes

source/gameengine/Rasterizer/RAS_OpenGLRasterizer/RAS_OpenGLDebugDraw.cpp
```diff
@@ -186,15 +186,15 @@ RAS_OpenGLDebugDraw::RAS_OpenGLDebugDraw()
 		attribVector(trans, stride, offsetof(RAS_DebugDraw::Box2d, m_trans), 4, 1);
 	}
 
-	glBindVertexArray(0);
+	GPU_bind_vertex_array(0);
```
panzergame: For a goodie function you can declare `GPU_unbind_vertex_array()` To bind the null VAO.

source/blender/gpu/shaders/gpu_shader_vertex.glsl
```diff
@@ -8,23 +9,26 @@ out block {
...
-in int ininstlayer;
-in vec3 ininstinfo;
-
-out vec4 varinstcolor;
-out mat4 varinstmat;
-out mat4 varinstinvmat;
-flat out int varinstlayer;
-out vec3 varinstinfo;
+/* compatibility / in */
+varying mat3 ininstmatrix;
```
panzergame: it's `attribute` instead of `varying` for vertex input.

source/blender/gpu/shaders/gpu_shader_vertex.glsl
```diff
@@ -8,23 +9,26 @@ out block {
...
+varying vec3 ininstposition;
+varying vec4 ininstcolor;
+varying int ininstlayer;
+varying vec3 ininstinfo;
+
+/* compatibility / out */
+varying vec4 varinstcolor;
+varying mat4 varinstmat;
+varying mat4 varinstinvmat;
+flat varying int varinstlayer;
```
panzergame: `out`/`in` specifier need `flat` but `varying` doesn't and can't.

PR #967, opened 2018-12-15
Make selected text symbols visible

thatimster:
Currently symbols with the default theme are invisible when selected in the text editor. Lowering the brightness of the selected text makes them visible without drastically changing theme look.


panzergame:
I tested it, looks good :)
You have to change the startup file to use this new theme :
- Open upbge
- reload the theme in user pref
- file -> Save Startup file
- copy your new startup file in .config/ or the windows configuration path to release/datafiles/startup.blend.
- commit

Issue #964, opened 2018-12-11
Option to change AA during game [request]


Jesusemora:
we need a python function to change AA during game, similar to `bge.render.getAnisotropicFiltering()`.
something like:
`bge.render.getAntiAliasing`()
`bge.render.setAntiAliasing`(_level_)

there is no way to change this option other than setting it from the blender settings.

adriansnetlis:
As far as I know, setting MSAA on-fly is quite challenging. Changing it requires reloading some OpenGL stuff. I think this could be done in menu if it loads a separate .blend file (thus loading different rendering system), but changing it in-game would be challenging to say the least. It could work, but the execution time of your proposed function would be quite slow. I may be wrong, but I am pretty sure I'm not. Anyways, screen-space based AA methods that you can code in 2D filters (e.g. FXAA) can be easily switched on/off.

lordloki:
Closing. Committed in master. Thanks for propose it.

PR #972, opened 2018-12-19
UPBGE: Allow changing anti aliasing in game.

panzergame:
The new python functions getAntiAliasing and setAntiAliasing are
implemented they both permit to read and set the anti aliasing level.
Setting the level implies recreating the off screen which can be slow.

Fix issue: #964.

Example file: 
[ge_anti_aliasing.zip](https://github.com/UPBGE/blender/files/2695012/ge_anti_aliasing.zip)

### DIFF
--- a/doc/python_api/rst/bge.render.rst
+++ b/doc/python_api/rst/bge.render.rst
@@ -307,6 +307,21 @@ Functions
 
    :rtype: integer (one of 1, 2, 4, 8, 16)
 
+.. function:: setAntiAliasing(level)
+
+   Set the anti aliasing level.
+
+   :arg level: The new anisotropic filtering level to use
+   :type level: integer (must be one of 1, 2, 4, 8, 16)
+
+   .. note:: Changing this value cause all off screens to be recreated, which can be slow.
+
+.. function:: getAntiAliasing()
+
+   Get the anti aliasing level setting.
+
+   :rtype: integer (one of 1, 2, 4, 8, 16)
+
 .. function:: setMipmapping(value)
 
    Change how to use mipmapping.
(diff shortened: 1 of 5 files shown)


doc/python_api/rst/bge.render.rst
```diff
@@ -307,6 +307,21 @@ Functions
 
    :rtype: integer (one of 1, 2, 4, 8, 16)
 
+.. function:: setAntiAliasing(level)
+
+   Set the anti aliasing level.
+
+   :arg level: The new anisotropic filtering level to use
```
lordloki: anisotropic filtering -> anti aliasing

source/gameengine/Ketsji/KX_PythonInit.cpp
```diff
@@ -1424,6 +1451,10 @@ static struct PyMethodDef rasterizer_methods[] = {
 	{"setAnisotropicFiltering", (PyCFunction)gPySetAnisotropicFiltering,
 	 METH_VARARGS, "set the anisotropic filtering level (must be one of 1, 2, 4, 8, 16)"},
 	{"getAnisotropicFiltering", (PyCFunction)gPyGetAnisotropicFiltering,
+	 METH_VARARGS, "get the anti aliasing level"},
+	{"setAntiAliasing", (PyCFunction)gPySetAntiAliasing,
+	 METH_VARARGS, "set the anti aliasing level (must be one of 1, 2, 4, 8, 16)"},
+	{"getAntiAliasing", (PyCFunction)gPyGetAntiAliasing,
```
lordloki: The changes are inserted in the middle of getAnisotropicFiltering braces {}, moreover the text points out to set but the function points out to get and viceversa.

panzergame:
@lordloki Thanks for review, changes are done.

youle31:
It seems there is an issue in this file:

http://pasteall.org/blend/index.php?id=50931

youle31:
Maybe this bug: https://github.com/UPBGE/blender/issues/976 should be fixed before merging this PR

adriansnetlis:
> I think the error message should also say 0, 2, 4 ... instead of 1, 2, 4 ... etc.

No, the anti-aliasing level determines how many points rasterizer samples per-pixel. So if you have 0 points, the rasterizer cannot figure out to which object the pixel belongs. That's why it starts at 1.

Momel15:
Yes, you're right, but the code says otherwise. Check the line I commented and the line before. They're different which shouldn't be the case.

PR #984, opened 2019-01-07
UPBGE: Fix Antialiasing setting at game engine start

youle31:
Since 23183936242af, in KX_BlenderCanvas constructor, UpdateOffScreens
was called and and a new offscreen was created, but the number of AA
samples was set to 0, the the offscreen was not created correctly.

To fix that, we can set the samples number in canvas constructor. This
way, the number of AA samples is available when we create the offscreen


youle31:
Fixes https://github.com/UPBGE/blender/issues/976

youle31:
Ok, thanks for your test, I merge :)

PR #999, opened 2019-01-25
UPBGE: Try to fix mipmap for cubemap textures

youle31:
In cubemap/planar rendering process, this:
void RAS_TextureRenderer::EndRender(RAS_Rasterizer *rasty, unsigned
short layer)
{
	m_layers[layer].Unbind(m_useMipmap);
}

was not called. To generate mipmap, we have to call it each frame.
This is an attempt to fix https://github.com/UPBGE/blender/issues/998

This would need to be reviewed by panzergame or lordloki


youle31:
test file: http://pasteall.org/blend/index.php?id=51100

lordloki:
Thanks @youle31

PR #1026, opened 2019-03-02
UPBGE: fix compound child remove parent

BluePrintRandom:
Remove parent was not updating the child shape of compound child
objects,it was checking to see if the root was compound, after it
already removed it's parent.

### DIFF
--- a/source/gameengine/Ketsji/KX_GameObject.cpp
+++ b/source/gameengine/Ketsji/KX_GameObject.cpp
@@ -411,6 +411,7 @@ void KX_GameObject::RemoveParent()
 	m_node->SetLocalScale(m_node->GetWorldScaling());
 	m_node->SetLocalOrientation(m_node->GetWorldOrientation());
 	m_node->SetLocalPosition(m_node->GetWorldPosition());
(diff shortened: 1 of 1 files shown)


youle31:
I guess i would move the code instead like that:

http://pasteall.org/1523424/diff

-> define rootnode, rootobj, and rootctrl before calling DisconnectFromParent() + NodeUpdate() but it is basically the same idea.

I add your test file so that others can test:

https://cdn.discordapp.com/attachments/290595893931933696/551521044750401659/compound_bug.blend

So i'm ok, congrats, good catch!

In the same time, the initial commit: https://github.com/UPBGE/blender/commit/fb1f17288108154a7c1eeed950d9de41cd9c6bf8

which introduced this bug causes other bugs in the tracker and is very complex, then I don't know if we should keep this commit :/ .

By the way, we should check if the mistake we have here is not in other parts of the commit by panzergame

youle31:
Just after having a quick look at the code, I think a mistake is remaining, because SG_Node *rootNode = m_node->GetRootSGParent();

rootNode is still in use just after in RemoveParent whereas it is defined after DisconnectFromParent and NodeUpdate()

marechal-p:
Follow up PR: https://github.com/UPBGE/blender/pull/1027

PR #1027, opened 2019-03-02
fix compound child remove parent try2

BluePrintRandom:
UPBGE: Fix compound child remove parent

previously the fix before worked for the test case but not for my game

youle pointed out something that worked for both

edit *corner case was remove parent -> set parent in same frame

### DIFF
--- a/source/gameengine/Ketsji/KX_GameObject.cpp
+++ b/source/gameengine/Ketsji/KX_GameObject.cpp
@@ -411,16 +411,12 @@ void KX_GameObject::RemoveParent()
 	m_node->SetLocalScale(m_node->GetWorldScaling());
 	m_node->SetLocalOrientation(m_node->GetWorldOrientation());
 	m_node->SetLocalPosition(m_node->GetWorldPosition());
-	KX_GameObject * parent = static_cast<KX_GameObject *>(m_node->GetParent()->GetObject());
-
-	m_node->DisconnectFromParent();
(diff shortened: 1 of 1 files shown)


source/gameengine/Ketsji/KX_GameObject.cpp
```diff
@@ -411,16 +411,14 @@ void KX_GameObject::RemoveParent()
 	m_node->SetLocalScale(m_node->GetWorldScaling());
 	m_node->SetLocalOrientation(m_node->GetWorldOrientation());
 	m_node->SetLocalPosition(m_node->GetWorldPosition());
-	KX_GameObject * parent = static_cast<KX_GameObject *>(m_node->GetParent()->GetObject());
-
-	m_node->DisconnectFromParent();
-	NodeUpdate();
+	
+
```
marechal-p: Remove useless whitespace.

marechal-p:
Also rename the commit to have the prefix and be less weird, maybe something like:
"UPBGE: Follow up fix for compound child remove parent".

marechal-p:
Ok, so you will not get the bonus points today, but the changes look good to me!

BluePrintRandom:
@lordloki  do you mind if I merge?
I tested w/complex and simple files using compound parent

marechal-p:
@BluePrintRandom the message should be the actual commit description.

PR #1067, opened 2019-10-27
UPBGE: UI space logic: Use same bl_category for components than for p…

youle31:
…roperties

In space_logic.py, the bl_category for python components was not
defined.

Then contrarly to upbge 2.7, comonents and game properties were in 2
different "tabs".

I think since 2.8, to avoid this we must define the bl_category each
time.

With this commit we can see that game properties and components belong
to the same UI


lordloki:
Thanks @youle31

PR #1075, opened 2019-11-14
UPBGE: Add a system to do HUDs without overlay scenes

youle31:
There are 2 parts:

1. Show in front option is implemented for eevee render to have the possibility to display HUD objects in front of other scene objects.
2. Dynamic actuator logic is completed. It's not only available for mesh objects OB_MESH but for all types of objects AND we can DISABLE/ENABLE PHYSICS for all children of an object.
It takes example on Visible actuator.
Using Visibility actuator + Dynamic actuator, we can hide/suspend physics for all objects of the HUD OR unhide/restore physics for all objects of the HUD (physics can be needed if we want to have the HUD objects clickable).

The HUD design should look like this:
Capture

Show in front option for HUD objects is here:
Capture

The root Empty of the HUD (main parent for all HUD objects) can be parented to main Camera to follow main Camera position/orientation.

I think this requires a bit more logic than overlay scenes, and may need python (or deal with property sensor/actuator) if for example we want to suspend logic/physics of non HUD objects while the HUD is displayed, but I think it can do the stuff.

I don't really see how we could restore background/overlay scenes with the current depsgraph.

I think it is more handy to deal with 1 scene at a time (We can still use replace scene if we want) and maybe deal with things based on several collections instead? Logic bricks could handle some options for collections too.

The design is not definitive and can be changed...

PS: If we stop to deal with several scenes at the same time, some parts of BGE code could maybe be changed? (SETLOOPER in BlenderDataConversion, m_scenes in KX_KetsjiEngine...)

### DIFF
--- a/release/scripts/startup/bl_ui/properties_object.py
+++ b/release/scripts/startup/bl_ui/properties_object.py
@@ -259,6 +259,7 @@ class OBJECT_PT_display(ObjectButtonsPanel, Panel):
 
         col = flow.column()
         col.prop(obj, "show_in_front", text="In Front")
+        col.prop(obj, "show_in_front_bge", text="In Front bge") #Game engine transition
         # if obj_type == 'MESH' or is_empty_image:
         #    col.prop(obj, "show_transparent", text="Transparency")
 
--- a/source/blender/draw/engines/eevee/eevee_engine.c
+++ b/source/blender/draw/engines/eevee/eevee_engine.c
@@ -254,6 +254,18 @@ static void eevee_draw_background(void *vedata)
     DRW_stats_group_start("Prepass");
     DRW_draw_pass(psl->depth_pass);
     DRW_draw_pass(psl->depth_pass_cull);
+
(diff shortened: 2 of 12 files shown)


BluePrintRandom:
When combined with mouse over any /xray it works amazing.

youle31:
Great! Thanks for review and test. I'll merge that then. I hope there are not too much bad side effects but if there is an issue, we can still change the design/ fix the code.

nestanqueiro:
this is looking good, and solves some issues, but it needs to support alpha to allow transparent menus to overlay on the scene.... do you think this is hard to accomplish? thanks for this great hack!

youle31:
Thanks! I think this just needs to add 2 lines of code in ma_transparent in eevee_materials.c. I'll try to do/test it soon. I didn't tested neither potential bad side effects with post processing...

youle31:
this doesn't seem to work with transparent materials

PR #1078, opened 2019-11-22
UPBGE: Add an API to get KX_GameObject's Object

youle31:
As KX_GameObject is a kindof container for Object, it can be useful to get its Object to perform operations with bpy on it


youle31:
test file: http://pasteall.org/blend/index.php?id=52616

youle31:
I need to change default camera code, no need to review for now. I thought BKE_view_layer_camera_find always returned a camera including viewport camera but it returns NULL when there is no Camera object in scene then i need to change the code.

youle31:
I think it's ok

youle31:
Thanks very much for review! I merge that

PR #1091, opened 2019-12-26
UPBGE: Restore a simplified version of ImageRender

youle31:
Simplified because I don't think we can use Render to Offscreen with this implementation.
But we can render something to a texture and use the rendered texture bindId to do things with in a 2D filter if we want.

Else, contrarly to old bge, the rendered textures are shaded, which is great.

I hope i didn't do too much mistakes, but i did few tests and it seemed ok.

The test files links are in some of the commits I did


youle31:
Thanks very much for reviw! I merge then :)

PR #1092, opened 2019-12-27
UPBGE: Small reorganisation of Render code an restore 2DFilterOffScreens

youle31:
Test file for 2dfilter offscreens:

http://pasteall.org/blend/index.php?id=52710

Notes:

1. As in 2.8 branch we use framebuffers instead of offscreens, i renamed
files and many variables, functions.... frameBuffer instead of
offScreen.

KX_2DFilterOffScreen is now KX_2DFilterFrmaeBuffer
same for RAS_2DFilterOffScreen > RAS_2DFilterFrameBuffer

but on scripting side, I see no issue to keep "offScreen" name for backward
compatibility.

Anyway, i doubt there are many people who use this.

2. KX_2DFilterOffScreen supported several color attachments per
offscreen, but I restored only 1 color attachment because it's simpler

### DIFF
--- a/source/gameengine/Ketsji/KX_2DFilter.cpp
+++ b/source/gameengine/Ketsji/KX_2DFilter.cpp
@@ -118,7 +118,7 @@ PyMethodDef KX_2DFilter::Methods[] = {
 
 PyAttributeDef KX_2DFilter::Attributes[] = {
 	KX_PYATTRIBUTE_RW_FUNCTION("mipmap", KX_2DFilter, pyattr_get_mipmap, pyattr_set_mipmap),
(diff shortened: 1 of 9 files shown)


youle31:
Note: The bug with default camera free is in this branch but not in upbge0.3 as the other pull request about ImageRender was merged

BluePrintRandom:
attachments are really cool.

hypersomniac expressed interest in restoring them later for realtime
compositor

youle31:
Thanks very much for review and test! Idk if the patch covers all corner cases (if someone tries to use several color attachments) but i think it's ok for most cases then I merge that. Thanks again!

PR #1095, opened 2020-01-04
UPBGE: Use blender View3D, RegionView3D, Aregion...

youle31:
As we can use bContext both in embedded and blenderplayer, we can also simplify a bit the code
reusing some variables already created:
1. in embedded: when we open blender
2. in blenderplayer: in blenderplayer init process which is almost the same than when we open blender.

Side notes:
- This is fixing Depth of field which was a bit different than in viewport.
- Later, if we want to tinker, we could try to implement viewport render for blenderplayer (see blenderplayer_viewport branch) and do other things like have a look at grease pencil, use workbench...

### DIFF
--- a/source/blender/draw/intern/DRW_render.h
+++ b/source/blender/draw/intern/DRW_render.h
@@ -692,9 +692,9 @@ typedef struct DRWContextState {
 const DRWContextState *DRW_context_state_get(void);
 
 /*****************************GAME ENGINE***********************************/
-struct GPUTexture *DRW_game_render_loop(struct bContext *C, GPUViewport *viewport, struct Main *bmain, struct Scene *scene, struct Object *maincam,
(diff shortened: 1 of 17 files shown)


youle31:
note: In the code, few lines have been added by Auto-indentation in KX_Scene.cpp which are not related to this patch

youle31:
I'm not sure to "always" get the right ARegion in KX_Scene::InitBlenderContextVariables() if multiples viewports are opened. I'll try to find a solution

youle31:
Another thing i'm wondering is should we replace ARegion when we replace scene...

youle31:
Thanks very much for the review and the test! I'll wait maybe 1 or 2 days to see if I have an idea which comes during night to improve the code. I'll merge after. Thanks again!

youle31:
Ok, I merge it like that. This makes new variables available and opens possibilities to do other changes if we want to tinker. Thanks very much for review

PR #1096, opened 2020-01-08
UPBGE: Support several engines drawing

youle31:
I added gpencil engine to test.

Before, only eevee engine was enabled in game engine render loop.
The render loop returned the final_tex rendered by eevee and was displayed in bge.

But I had another look at render code and they do differently.

A GPUViewport is bound before drawing.
All engines are rendered and at the end of each engine render loop, the returned color is written into the GPUViewport framebuffer (which can be accessed for example with DRW_viewport_framebuffer_list_get()->default_fb)
Then the GPUViewport is unbound.

Then finally we can access the viewport texture which has accumulated the results of all enabled engines with GPU_viewport_color_texture(viewport).

Then I refactored the game engine render loop in this way to have the possibility to use several engines.

Furthermore, I added the possibility to render in lookdev mode.

Side note: There's currently a bug in blender code in the check to enable gpencil engine ot not (gpencil engine is always enabled). I opened a bug report about it: https://developer.blender.org/T72973 then it will work in our branch when the bug will be fixed in blender.

### DIFF
--- a/source/blender/draw/engines/eevee/eevee_engine.c
+++ b/source/blender/draw/engines/eevee/eevee_engine.c
@@ -354,7 +354,16 @@ static void eevee_draw_background(void *vedata)
     bool use_render_settings = stl->g_data->use_color_render_settings;
 
     GPU_framebuffer_bind(dfbl->default_fb);
-    DRW_transform_to_display(stl->effects->final_tx, true, use_render_settings);
+    /* Game engine transition - ImageRender: avoid double tonemapping */
+    const DRWContextState *drw_ctx = DRW_context_state_get();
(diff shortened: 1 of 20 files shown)


youle31:
Thanks for the hearts :) It's not totally ready though. I have to fix issues in Camera view

youle31:
I think this is all good now

youle31:
Thanks very much for review and tests! Ok I merge and If we want to tinker with to fix these 2 bugs (pink background in lookdev in blenderplayer (might come from some userdef/theme not initialized or a function not called in gpg_ghost init to properly initialize background color?) + antialiasing), we still can in master. Thanks!

PR #1098, opened 2020-01-16
UPBGE: Partially Recover ReplaceMesh and Level of Detail

youle31:
Initial Pull Request to recover a simplified version of ReplaceMesh and LOD:

Simplified version of ReplaceMesh:
-> Physics in not handled for now, only rendered mesh is replaced

Simplified version of LOD:
-> The distinction between materials/meshes is not restored. When a new LOD is choosed both mesh and materials change.

Just before rendering, we update level of detail.
We replace the object data with the "final evaluated" LOD object data.
When I say "final evaluated" it means that it is the LOD object data which takes into account modifiers.

We could use a bit more directly blender code to choose level of detail but as there was already bge code which is there and makes use of ReplaceMesh (takes into account RAS_MeshObject (for physics shape eventually for later), then there are 2 versions of the code, 1 in blender viewport, the other in bge code.


youle31:
I need to comment the UI code in ReplaceMesh actuator for physics for now. I'll do that later

youle31:
I think it is ok for tests if you want. In the last commit, I merged last master because i did few commits to repair debug builds in master and i wanted to know if the LOD viewport code worked in debug mode.

I seemed ok with simple tests files.

youle31:
This is still bugged with another test file...

lordloki:
For me was working before last commit. I havent tested with last commit yet.
https://youtu.be/qYmsc_IgCHQ

youle31:
Yes, the issue is in a specific file when I try LOD with more complex objects.

tets file: https://drive.google.com/file/d/1slIiH5dh-EZVF7r1HU92c6jRigkH96O0/view?usp=sharing

youle31:
(the water has a script which causes an issue. The script has to be deleted to avoid more issues)

youle31:
I'm sorry but I think i'll give up. This sounds too hacky and I don't understand enough. I think we'd need to replace ob_eval->data with lord_ob_eval->data data each frame just before render and after BKE_scene_graph_update tagged. But this is just guesses and idk why i have weird behaviour with some test files. This is a too weak design. It would need another approach to do the stuff

lordloki:
I tested your file and i had the following issues:
- When you move the camera to get far or close to objects to provocate the LOD jump the shadows don't update until you stop to move.
- The tree with sleeves is showed without sleeves.

- When you move the objects to get far or close to camera to provocate the LOD jump you shadows are correct but the tree with sleeves continue showing without sleeves.

BluePrintRandom:
perhaps this should be handled in the drawing code?
like the mesh provides 1 or more meshes (lod levels)
and the drawing code does the switching?

maybe open a pull in blender 2.8 just for this little diff*
(they need LOD too eh?)

lordloki:
I was trying changes in the code with no success either. More tomorrow

lordloki:
Applying modifiers before UPBGE start we solve the issue with the shadows.

youle31:
Maybe we could change the design of LOD itself and add automatically a copy of the original object to the beginning of the list of levels of detail. Like that we could change directly "original" object data instead of trying to change "evaluted" object data each frame just before render as in viewport. idk

youle31:
Finally i kept the first idea.

Each frame just before render we replace the data of the object which should be rendered with the data of the evaluated LOD level object. This operation must occur just before render AND after all the objects have been tagged for update AND after the updates have been flushed with BKE_scene_graph_update_tagged. This works with my stupid test file.

As we need to replace the rendered data each frame, when we use ReplaceMesh for an object, we add to this object a dummy LodManager with a KX_LodLevel which has the data we want to render. then UpdateLods will be called for each objects which have used ReplaceMesh.

lordloki:
Wow!!! I tested and it's works like a charm now. Very very good work!

youle31:
Thanks! It's very experimental, but well, it seems to work then : )

Issue #1104, opened 2020-01-28
Navmesh causes crash on Game Exit


IzaZed:
When setting any object to Navmesh in the physics Tab, the game will run, even perform the Navmesh calculation correctly, but then crashes on exit.

lordloki:
The issue comes when we do undo to restore the scene original then we have to do BKE_id_free but Navmesh has no ID set.

lordloki:
Fixed in current master. Thanks for report. Closing now

PR #1107, opened 2020-01-30
UPBGE: Fix for #1104, engine crash on exit due to NavMesh free data

lordloki:
Now, we don't free navmesh id as it is not based in id.
Additionally, we improve undo system including undo based in step name.


youle31:
It works fine for me, tested in debug mode and no error at exit, and it's cool for the undo improvement at exit. Now I have to find the button to approve a pull request

lordloki:
Thanks a lot for testing. Merging into master

PR #1109, opened 2020-01-31
UPBGE: Fix sun shadows in blenderplayer, and reduce mem leaks

youle31:
Temp. Just to have the pull request view on changed files.

All of this is based on experiments.

With the previous version of blenderplayer "initialisation", no
DST.gl_context was created when we called DRW_opengl_context
create_blenderplayer().

If we create a DST.gl_context, we need to adapt the code a bit to have
it working, and this is what I did here following the compiler error
messages.

Why I wanted to create a DST.gl_context?

- Because I noticed that it was fixing an issue with sun shadows
(reported by Nestanquiero) and that I had in some of my files too.
- Because it seems that there are less mem leaks reported at
blenderplayer exit.

"Activate" the gl_context was causing an issue with vsync, but this can be fixed calling canvas->SetSwapInterval each frame. If we don't "Activate the gl_context, there are issues in some files.

The blenderplayer seems slower to start but once it is launched, it sounds ok.

(I pushed a commit where i cleared launch screen in white to have the same behaviour than before, but when we "activate" the drawing gl_context, it is cleared in black, but i personally don't really care about the color of launching screen.)

As this code is not "more experimental"/"more hacky" than the previous one and seems
to work better, I propose to commit it after some tests.
In my test files, it sounds ok. I'll continue to do few other tests.


youle31:
Thanks very much for tests! I'm in family but i'll continue a bit more tests in next days then commit if this sounds right to me

PR #1115, opened 2020-02-06
UPBGE: Add a "Collection actuator"

youle31:
The Collection actuator is meant to replace a part of Scene actuator.

In the Scene actuator we had:
- Overlay Scenes which were used by the ones who use bge to create games
- Suspend/Resume Scene

The Collection actuator can replace this logic.

1. Suspend/Resume have 3 options:
- logic (suspend/resume logic + animations for all objects + children in the collection)
- physics (suspend/resume physics for all objects + children in the collection)
- visibility (suspend/resume visibility/render for all objects + children in the collection)

2. Overlay has 2 modes:
- Add overlay collection (render all objects of the collection with a selected camera in overlay)
- Remove overlay collection

The camera used to render in overlay must not be used neither for ImageRender, neither as main active camera.

The objects of the overlay collection are only influenced by the lights which are in overlay collection.
Anyway, it is recommended to not use lights (only emission shaders) to avoid extra performance cost.

I guess Overlay render pass can have a BIG performance impact on low end GPUs because it requires an ENTIRE extra render pass.

On code side:

I choosed to make GPUViewport a member of KX_Camera. Then each camera can have its GPUViewport with it's own render data. This was few changes done in ImageRender and in the main render loop.

I implemented SuspendLogicRecursive on the model of SuspendPhysicsRecursive and SetVisibleRecursive. This is just the sensors which are disabled. I didn't tested but the internal logic was already implemented in SCA_ISensor.

About the overlay render pass:

All objects inside the "overlay collection" are flagged with OB_OVERLAY_COLLECTION when an overlay collection is added.

When eevee's cache is "populated", if the object is not flagged, it will be rendered in normal render pass, else it will be rendered in overlay render pass.

Main Render camera and others active cameras (ImageRender..) are always rendered before Overlay Render Camera.

When we Render Overlay Camera, we disable background rendering and we choose a special openGL "blend mode" to do the overlay.

quick tuto i did this morning:

voiture


youle31:
Thanks very much!

PR #1118, opened 2020-02-07
UPBGE: Try to fix crash when removing bge.constraint in some cases

youle31:
This is inspired by old bge code which had not this issue. I think this comes from the way constraints are created with NewProxy(true)

### DIFF
--- a/source/gameengine/Ketsji/KX_ConstraintWrapper.h
+++ b/source/gameengine/Ketsji/KX_ConstraintWrapper.h
@@ -40,7 +40,8 @@ class PHY_IConstraint;
 class	KX_ConstraintWrapper : public CValue
 {
 	Py_Header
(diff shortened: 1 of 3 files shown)


source/gameengine/Ketsji/KX_ConstraintWrapper.h
```diff
@@ -33,14 +33,14 @@
...
+#include "PHY_DynamicTypes.h"
 
 class PHY_IConstraint;
 
 class	KX_ConstraintWrapper : public CValue
 {
 	Py_Header
-public:
-	KX_ConstraintWrapper(PHY_IConstraint *constraint);
+public: KX_ConstraintWrapper(PHY_IConstraint *constraint, PHY_ConstraintType ctype, int user_id);
```
lordloki: I think it is better to put the KX_ConstraintWrapper in the next line after public: not in the same line.

youle31:
Thanks very much for review and tests! I fixed, the typo, I merge.

PR #1119, opened 2020-02-08
UPBGE: Try to fix some transformation weirdness (with anims,addObject..)

youle31:
For now, because I understand almost nothing to the depsgraph,
KX_GameObject::TagForUpdate was using "evaluated object" mainly to
update the Object obmat.

I did that because I had issues with AddObject
in some cases.

But when we use the original object, and
BKE_object_apply_mat4 and ID_RECALC_TRANSFORM | ID_RECALC_COPY_ON_WRITE,
it seems to work better, fixes some issues with animations not played
and a glitch when we addObject.

In most of my test files, this seems to work better and this is more
like in the blender code.


youle31:
No, this is giving warning messages for files with armatures...

youle31:
I'll try to find something better

Issue #1111, opened 2020-02-02
problem in bge render and not in viewport render


youle31:
test file:

http://pasteall.org/blend/index.php?id=52858

Steps:

In camera view, test both in viewport render and in bge render -> not the same behaviour

youle31:
Note: when we put TagForUpdate after BKE_depsgraph_update_tagged, it seems that there is no more the shift... but it doesn't seem a solution

PR #1126, opened 2020-02-15
UPBGE: Fix differences between bge and blender cameras/matrices

youle31:
Now we can directly use blender code to setup camera matrices. This is
fixing issues like: https://github.com/UPBGE/upbge/issues/1111
and others issues when we tweak "advanced" camera settings

However, this is causing a small backward compatibility issue with SCA_SceneActuator->SetCamera.
When we start in Viewport view (default camera), we are not abled to SetCamera because rv3d != RV3D_CAMOB. But anyway, even if it was possible before, I think the GPUViewport size was not correct, and this is always better to start from camera view to ensure to have correct GPUViewport size to not loose performance.

This is causing other backward compatibility issues in some files because now the camera settings are exactly as in blender whereas before, there was a difference between viewport and bge.

But this is great to have the good settings for camera.


youle31:
I'll try to improve that later (try to fix the SetCamera on the first frame)

youle31:
Actually, using directly blender camera would possibly involve many changes in the code. I'll close the pull request for now, and do more experiments

youle31:
Thanks very much! I think i've a fix for nestanquiero bug too

PR #1134, opened 2020-03-01
UPBGE: Fix animation transform issue

youle31:
related to parenting.

test file: https://ufile.io/rt17853z

Bug reported here:
https://blenderartists.org/t/upbge-0-3-logic-nodes-for-smoothing-camera/1209823/17

code picked from object_transform after BKE_object_calc_time_where_is (which is used to update animations)


youle31:
Thanks! I merge then

PR #1139, opened 2020-03-07
UPBGE: Cleanup overlay/background/suspend/resume scenes

youle31:
I propose to do the API cleanup in another commit.

I left RemoveScene in the code because it is used in FreeBlendFile but
removed it from scene actuator UI


youle31:
Thanks very much for review! I merge then

PR #1140, opened 2020-03-08
UPBGE: Fix loading new .blends with start game from file actuator

youle31:
I don't understand why blenderplayer was not like in BL_KetsjiEmbedStart.

In BL_KetsjiEmbedStart, when we start game from a new .blend file, this is called:
```
if (bfd) {
  BLO_blendfiledata_free(bfd);
 }
```
As start game from file was not working in blenderplayer, i did like in embedded, except that due to special gpu/opengl/window context, I had to close the previous window and open a new one. It seems to work.

Else, there is also code in KX_PythonInit to not try to reinitialize bpy because it was causing an assert error.

This is not really ready for now because i'd need to check if in the new loaded .blend, we should free bpy and initialize it again to have the right bpy.data in the new .blend.

As usual, I create the pull request to have the code in pull request view but I'd need to review it myself before


youle31:
It works! Magic.

Test files:

launcher: http://pasteall.org/blend/index.php?id=53013
new .blend: http://pasteall.org/blend/index.php?id=53014

youle31:
I close this waiting we find a solution which works in embedded and blenderplayer

PR #1141, opened 2020-03-10
UPBGE: Cleanup lights code + API

youle31:
test file: http://pasteall.org/blend/index.php?id=53023

I removed RAS_ILightObject, RAS_OpenGLLight.

There's just KX_LightObject class which is kept to have LightsIpoController working.

Else, all can be done with bpy. (light = scene.objects["point"].blenderObject.data)

About lights frustum drawing, idk if it is possible, but it should be done in eevee and not in bge.


youle31:
Thanks very much for review! I merge

PR #1161, opened 2020-03-31
UPBGE: Make format after sorted includes update

youle31:
Some time ago, we already make format command for the whole sources, but the make format script now "sort includes by blocks". I didn't see documentation about that but I did experiments in this branch: https://github.com/UPBGE/upbge/tree/sorted_includes to see how it was working.

As I think I ~understood the way it was modifying the sources, I prepared the includes, then applied the make format command progressively here: #1161

following the rules I decided arbitrarly:

1. Include the main include (For BL_Action.cpp for example, the first include "block" will be BL_Action.h)
2. The second include "block" is the includes between quotes < >
3. The third include "block" is the blender C files includes (BKE, GPU, BLI....)
4. The 4th include "block" is composed with gameengine includes (KX_ SCA, RAS...)

It's possible that there are few mistakes as i tried to go fast, but it compiles and this clean up a bit the sources.


youle31:
I'll reboot on linux to test compilation

youle31:
hehe Thanks! It was my "quality code day try" yesterday. I merge that then.

PR #1168, opened 2020-04-06
UPBGE: Recover use of DRW_engines_free

lordloki:
previously we had crashes related to basic engine depth conservative shader.

The reason is that when we finalized the game engine the basic engine was freed (freeing the depth_conservative shader) but the depth shader wasn't freed as it is a built-in shader and it uses another procedure.

Then when we are initializing again the basic engine for Blender we only were checking for depth shader inexistence to create both shaders but we could have the case that depth shader exists but depth_conservative shader not.

To avoid that we separate the creation of both shaders checking for its existence indepently.


youle31:
Yes, I'm not really talented to free things, but i guess this is better to free more than to free less when we exit the KX_Scene. Then I approve.

I'm wondering if we should call DRW_engines_free even before running game engine but we can see that later.

youle31:
and I can't crash the scene clicking everywhere with your fix

lordloki:
Thanks for reviewing!

PR #1177, opened 2020-04-12
UPBGE: Fix volumetric banding

lordloki:
When the TAA was used, the current code avoided to use the history buffer for the temporal volumetric integration. It used a some kind of simplification.

Instead of it, we use the history buffer to "mix" the volumetric from one frame to another (as it was recommended by https://es.slideshare.net/DICEStudio/physically-based-and-unified-volumetric-rendering-in-frostbite and only used for viewport offline rendering until now). 

Regarding performance i didn't see negative effects.


youle31:
I tested with several files and this gives amazing results compared to before! That's awesome. Thanks!

PR #1178, opened 2020-04-13
UPBGE: Replace LibLoad with the possibility to convert blender objects during runtime

youle31:
It is using BL_ConvertObjects modified a bit to accept a single object conversion.

We can create an object with bpy or append an object from another file with bpy then call:

bge.logic.getCurrentScene().convertBlenderObject(bpy.data.objects["name"]) to convert it into a gameobject during runtime. To remove this object, we just have to call bge endObject method (I guess)

voiture

I didn't tested deeply but here is are my test files:

launcher: https://pasteall.org/blend/a7c9eb35dcb5410fa8cd1a8d3288dd8f (an empty scene)
external .blend: https://pasteall.org/blend/6ab4bcd4018f4eb1ac62a32310992efc

The cube which we are appending into the launcher scene has motion logic bricks which seem correctly converted, it has a LodManager which seems converted, and it has physics which seems converted.

I can't guarantee that there are 0 issues but it seemed ok in my test file.
The advantage is that we are reusing BL_ConvertObjects code and that we are using bpy.ops.wm.append which is blender code.


lordloki:
I was testing the pull with big (2.100.000 triangles) and multimaterial objects (https://youtu.be/A087IylXYHM) and it works like a charm!!!!

Big kudos to @youle31

youle31:
Cool! Thanks for the test in "extreme" conditions : )
I merge then.

PR #1179, opened 2020-04-14
UPBGE: Adding python convertBlenderCollection function

lordloki:
taking advantage of the previous work on convertBlenderObject function we make
this new function to simplify the loading of all objects inside a
collection.

We will see if we can parallelize the function later.


youle31:
no this seems ok about decref comparing to other similat functions.

PR #1181, opened 2020-04-15
UPBGE: Restore post_draw callbacks

youle31:
I inserted the code after RenderDebugProperties.

We are in ortho projection matrix so if we want to use shaders in 3D, I guess we need to pass camera matrix as uniform for gpu shaders.

I tested with 3 files:

1 is using blf module
1 is using gpu module
1 is just to ensure bge.render.drawLine was still working

https://drive.google.com/drive/folders/1O667_fy9_vmRZBmo68i2s0A89d192kCr?usp=sharing

The code here: https://docs.blender.org/api/master/gpu.html?highlight=gpu#module-gpu can be used, but I guess it needs to be adapted in some cases to pass the camera matrix (cam.projection_matrix @ cam.modelview_matrix)


youle31:
Thanks very much for tests and review!

PR #1189, opened 2020-04-23
UPBGE: Prevent draw request when DRW_state_image_render

youle31:
Actually, i noticed that in wanderer scene there was a little flickering
which didn't happen in blender official.

This seems to fix the issue. (in bge render loop, we never set the
DRW_state to DRW_state_image_render (we could to avoid to have infinite
taa maybe, but this would need to check))


youle31:
Does this patch gives bad results for your scenes with volumes? (Sorry if this changed the behaviour)

youle31:
lordloki found an alternative solution which sounds to work fine commenting the line which set "vol_history_alpha" to 0.0

PR #1199, opened 2020-04-29
Render Template

Shohrukh1999:


youle31:
First, i want to apologize to have closed your bug reports a bit brutally last time. I'm a stressed person and i can really be a jerk sometimes (often) when there is something added to the bug tracker (we are only 2 devs, lordloki and me, we work for free, and when something is added in the tracker (a bug report, a request...) it takes time/work for us to answer.

About this code, I've nothing against, but I don't think it should be integrated to upbge sources for several reasons:

- It is not a python component template. Python components are some kindof custom logic bricks that you add to an object following the format which you can see here: https://github.com/UPBGE/upbge/blob/master/release/scripts/templates_py_components/third_person.py
It is a class inherited from bge.types.KX_PythonComponent. The purpose of a python component is to be a kindof logic brick which can do something which is not handled by existing logic system.

- The code you use is already part 1. of the UI (So you can set it before runtime), and 2. of the python API. You can use it for your own goals in your own files, but as we already have an UI and an API for this, this is not useful enough to be integrated into upbge sources code.

(a pull request means you want to add a piece of code into the sources code)

About use_shadow_high_bitdepth, I think it isn't working properly in realtime when objects are not moving because of a small optimisation I tried to do in KX_GameObject::TagForUpdate to not update shadows when objects are not moving because shadow rendering pass have a performance cost.

This:

Capture

prevents shadows to be updated if objects are not moving, because in most cases, shadows don't need to be updated if the objects are not moving and this is why I think this setting change is not directly taken into account when the objects are not moving (And this is generating this weird behaviour of the scene which becomes dark)

I hope i was clear enough about the reasons we can't accept this pull request.

PR #1202, opened 2020-05-02
UPBGE: Ignore Python user site-packages directory by default

lordloki:
We copy the solution from https://developer.blender.org/
rB79a58eef059ffc3f12d11bd68938cfb1b4cd2462 to solve issues that we have
with Audaspace python

Thanks to @youle31 for find the solution

### DIFF
--- a/source/blender/python/intern/bpy_interface.c
+++ b/source/blender/python/intern/bpy_interface.c
@@ -425,6 +425,14 @@ void BPY_python_use_system_env(void)
   py_use_system_env = true;
 }
 
+/* Game Engine transition */
+bool BPY_python_get_use_system_env(void)
+{
+  BLI_assert(!Py_IsInitialized());
+  return py_use_system_env;
+}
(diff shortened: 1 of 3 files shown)


source/blender/python/intern/bpy_interface.c
```diff
@@ -425,6 +425,14 @@ void BPY_python_use_system_env(void)
   py_use_system_env = true;
 }
 
+/* Game Engine transition */
+bool BPY_python_get_use_system_env(void)
+{
+  BLI_assert(!Py_IsInitialized());
```
youle31, line 431: I tested in Debug mode on my computer and this assert error is always raised. Else audio and python work normally (but is was already working for me before). But if this is solving the issue for Daniel Er Bauer, I approve of course the pull request, but without the BLI_assert

youle31:
Thanks very much lordloki

lordloki:
Thanks for reviewing it

Issue #1188, opened 2020-04-23
Volumetrics Toggle


IzaZed:
Feature Request
The recently implemented Volumetric frame-to-frame interpolation by @lordloki is incredibly great, to have that said
But in certain cases it causes some artifacts, especially then using a higher anisotropy value on the volumetric shader.

It would be nice to have a Toggle for the interpolation feature under the render settings

youle31:
lordloki pushed a new commit about volumes to fix an artifacts, could you test again in the next builds in the next days? Thanks!

lordloki:
A build with the fix was uploaded

IzaZed:
The fix for flickering seems to work 👍  
The artifact that makes using anisotropy problematic (as far as my layman brain suggensts) comes from mixing in the previous calculations though

Here's a setup showing the effect, it should be noticable as soon as you move the camera around in the viewport:
[vol_anisotropy.zip](https://github.com/UPBGE/upbge/files/4533008/vol_anisotropy.zip)

lordloki:
Great!. Thanks for reporting.
Closing now.

IzaZed:
Did you have time to check the file? The Issue with anisotropy is still there :/

lordloki:
Oh, i checked rotating the cube but not moving the camera. I see the trail now.
Surely we will have to do a check to change between both behaviours

BluePrintRandom:
http://twvideo01.ubm-us.net/o1/vault/gdc2016/Presentations/Pedersen_LasseJonFuglsang_TemporalReprojectionAntiAliasing.pdf


here is a really good paper on the subject

youle31:
Fixed in https://github.com/UPBGE/upbge/commit/008aa804be52e90d9fdd9883ee374674074a4c23

A note has been added in release notes:
https://github.com/UPBGE/upbge/wiki/Release-notes-version-0.3.0

Thanks to lordloki for original patch for volumes, and for review of this patch!

PR #1210, opened 2020-05-14
UPBGE: Add a button to enable/disable volumetric blending (lordloki

youle31:
volumes patch)

This is a patch to answer IzaZed request here:
https://github.com/UPBGE/upbge/issues/1188

Notes:
When we are working in the viewport, the volumes are updated quicker
when volumes blending is not enabled then it's ok for me to have the
choice to toggle between both behaviours

A test file: https://pasteall.org/blend/c5629cf90d1a402897ab6fca3f7fc384

### DIFF
--- a/release/scripts/startup/bl_ui/properties_render.py
+++ b/release/scripts/startup/bl_ui/properties_render.py
@@ -363,6 +363,20 @@ class RENDER_PT_eevee_volumetric_shadows(RenderButtonsPanel, Panel):
         layout.active = props.use_volumetric_shadows
         layout.prop(props, "volumetric_shadow_samples", text="Samples")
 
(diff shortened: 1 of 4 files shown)


lordloki:
Works great! Do you thing that we should activate blending by default?

youle31:
As you want lordloki. Let's say that when we are in viewport, the eevee's original behaviour is often more handy, but when we are in runtime, with the option enabled is far better.
I can add a note in the release notes maybe to add visibility to this option?

lordloki:
OK, It is a good solution. Thanks

youle31:
Ok, thanks for review! I merge then i'll add a note in release notes.

PR #1211, opened 2020-05-22
Old shadows

youle31:
I give up on that but i wanted to have pull request view to see the number of lines. I'll close this just after


BluePrintRandom:
what was the end result?

youle31:
- problems with shadows disappearing when the light is not well placed
- sun shadows not working
- problem with compilation of some materials in some scenes

Issue #1172, opened 2020-04-10
Armature Jittering when parented


IzaZed:
Hi, I can't seem to get myself working with armatures, they start to jitter once the armature object is parented to a moving one

Here's the test file:

[test.zip](https://github.com/UPBGE/upbge/files/4461344/test.zip)

move around with WASD and the hammer of the gun should start to jitter, I had the same thing when working with an animated character

Is this a bug or am I missing something?

youle31:
Hi, do you have a test file not using node logic addon?

(I fixed a bug related to play armatures actions when the armatures are parented to a camera. Blender code is using a function called ignore_parent_tx iirc to avoid this jittering and I picked this code to solve the issue)

https://github.com/UPBGE/upbge/commit/131e83e37571b91a4bbd9492c638708105bbda68

IzaZed:
A to move forward, Left Mouse to trigger the shooting animation

Shoot at least once, apparently the jittering starts when some bone is not in its original position

[test.zip](https://github.com/UPBGE/upbge/files/4502912/test.zip)

RafaelNasci:
Same problem here.. Can't use any armature parented to capsule in UPBGE 0.3. This was done with suzanne and 3 bones.   When I move the capsule with WASD keys, the mesh jumps out of place and the armature to another place. I have no colision on mesh nor on the armature and I tested the capsule with every physics types(box,charater, etc.) and colision boxes. Even without colision box the armature jumps. I've parented with every type possible(with automatics weights, envelops, no vertex groups etc. and still the same. Even if I apply the armature to the mesh leaving it without modifiers, it still jumps. if I press P and move to left and press ESQ to get out of game, the dashed parent line moves the same amount in the opposite direction, and the mesh and armature goes to different "random" places. Initially I've tried with a makehuman mesh with armature and everything worked very nice, I can walk, trigger different animations, etc. But when I parent the armature to the capsule all hell brakes loose.

youle31:
Fixed

PR #1213, opened 2020-05-26
UPBGE: Fix another issue in parent-children relation transformation

youle31:
This is fixing https://github.com/UPBGE/upbge/issues/1172 but can have a
performances cost

### DIFF
--- a/source/gameengine/Ketsji/BL_Action.cpp
+++ b/source/gameengine/Ketsji/BL_Action.cpp
@@ -378,6 +378,7 @@ void BL_Action::Update(float curtime, bool applyToObject)
     // Extract the pose from the action
     obj->SetPoseByAction(m_action, m_localframe);
 
+    /* Armatures have children then fix transformation */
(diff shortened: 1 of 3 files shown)


youle31:
The old test file is no longer available...

was this commit: c9380510ab8177bef3678f3a5a33a15e

lordloki:
I also tested with these files:
[parentRelation.zip](https://github.com/UPBGE/upbge/files/4683590/parentRelation.zip)
[parenting bug.zip](https://github.com/UPBGE/upbge/files/4683591/parenting.bug.zip)

youle31:
Great! Thanks very much! I had no longer the test file with armature parented to camera and this works ok.

A potential optimisation would be to iterate on children instead of scene->GetObjectList() but I don't know why at this moment of the pipeline, children SG nodes have no "clientobject" then I left it like that but it can be improved.

I merge like that and we can improve later.

PR #1216, opened 2020-06-02
Rename blender & blenderplayer to upbge & upbgeplayer

lordloki:
Windows works but i have to test it in other operating systems and other things as addons, etc.
WIP yet.


BluePrintRandom:
🥳 🍾 👯‍♀️

youle31:
There's no problem for me, but the master is broken is several of my test files, after eevee materials refactor... Maybe it would be better to wait that the master is fixed before merging this commit. I'll see in the next days if I find motivation, but I saw there was a problem with InitMaterialGPUViewport, RenderToTexture...

youle31:
I did another merge and most issues are gone even if ImageRender doesn't work (I spent some time to try to know why but didn't find). It's ok if you want to merge. Thanks!

lordloki:
Better leave this pull for other moment

Issue #1225, opened 2020-06-18
Add dispose callback for components and game objects


mysticfall:
Currently, we can either use the constructor or `start()` callback to initialize a `KX_PythonComponent`, but it does not provide a method to be notified when the component is destroyed (e.g. a character dies, or loses a trait which is implemented as a component, and so on).

In certain cases, it may be required to perform clean up tasks before the game object to which such a component is attached gets disposed, like when it was monitoring outside events.

Other game engines like Godot or Unity provides a callback like `GameObject.dispose()` or `MonoBehaviour.OnDestroy()` to support such a use case.

So, it would be useful if UPBGE can also provide a similar callback in `KX_PythonComponent` and `KX_GameObject`.

youle31:
@marechal-p  implemented callbacks at scene exit here: f62dca6f3c8028dc69fff5be439fcc5d90a87124
(Comment:
Would it be possible to call RunOnRemoveCallbacks from KX_Scene destructor?)

As upbge0.3 is based on an old version of upbge, these callbacks have not been restored yet.

We could restore these callbacks for KX_Scene and add callbacks for KX_PythonComponents in the destructor.

@marechal-p  @lordloki : Do you have comments or thoughts about that?

lordloki:
Yes, i will go over all python components commits to recover them

youle31:
I think python components callbacks did not exist, but we can implement it easily taking example on KX_Scene

marechal-p:
@youle31 how would such a PythonComponent callback look like?

IMO it should be an optional method people should implement in their component, like `start` and `update`. Maybe call it `dispose` like in godot?

mysticfall:
I think `dispose()` could be a reasonable choice for consistency, but I'm a bit concerned if it would cause confusion for new users like myself, since it appears to be a method one can invoke to destory the object in question.

For reference, Unity provides the same methods as UPBGE for `start()` and `update()`, however, it prefixes all other callbacks with `on`, while Godot adopts a similar approach with `_` as a prefix.

youle31:
With the above code, it would give something like that:

![Capture](https://user-images.githubusercontent.com/10052509/85055184-776f8300-b19d-11ea-8072-18e916cc6037.PNG)

It's not good @marechal-p  ?

youle31:
Ok after discussion with @marechal-p , i'll do differently and then i'll send a pull request when i'll be motivated

youle31:
Requests doone here: 

[...]

PR #1227, opened 2020-06-19
UPBGE: Add onRemove callbacks for KX_GameObject (copied from KX_Scene

youle31:
ones)

This is to answer the second part of this request:
https://github.com/UPBGE/upbge/issues/1225

voiture

test file: https://pasteall.org/blend/473877dc9df54cf58c2635a89fdaf7f3

### DIFF
--- a/doc/python_api/rst/bge_types/bge.types.KX_GameObject.rst
+++ b/doc/python_api/rst/bge_types/bge.types.KX_GameObject.rst
@@ -537,6 +537,27 @@ base class --- :class:`SCA_IObject`
 
       :type: :class:`KX_LodManager`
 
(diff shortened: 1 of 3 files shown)


youle31:
I did like that because other callbacks in KX_Scene, and KX_GameObject are on this model.

KX_PythonComponent were a special case because there were already existing builtin method with names that we need to follow (start, update, dispose or end if you want to change)

But for KX_Scene, we have post_draw callbacks which look like that, we had onRemove in KX_Scene implemented by wkk on this model, and in KX_GameObject we have also collisionCallbacks which are using this model.

youle31:
Thanks! I merge that then.

PR #1235, opened 2020-06-23
UPBGE: Try to fix camera matrix when restart game

youle31:
Random fix

Idk why cam matrix doesn't like RestoreObmat code.

Calling RestoreObmat cause issues when we restart a scene with a camera
parented to an object like in this file:

https://pasteall.org/blend/6bab5870d9ef4c9d91ea9a6f80321c0e


youle31:
Thanks @lordloki to have accepted my random fix

PR #1246, opened 2020-07-04
UPBGE: Fix for update animation in KX_WorldIpoController

youle31:
WorldIpoController was not using the right scene

Using KX_GetActiveScene seems dangerous.


lordloki:
Test it and it works like a charm. Great job

youle31:
Thanks! I merge into upbge0.2.5 then

Issue #1255, opened 2020-07-17
Fix the camera fov problem in reflection map


MarllonxD3:
When i change the camera fov/lens in game, this causes a strange problem in the reflection material...
it's like the reflection does not identify the change in camera fov. 
It's possible to fix it? I'm using UPBGE v2.5
.blend file:
https://www.mediafire.com/file/qcrbgsb71v15q3d/FovProblem.blend/file

youle31:
Fixed here: https://github.com/UPBGE/upbge/commit/26b5c05d0932f6e993b80f92d9f9cf225abd6372

Thanks for report

PR #1256, opened 2020-07-17
UPBGE: Fix wrong projection matrix used by texture renderers when main

youle31:
cam projmat settings change

KX_TextureRenderers projection matrix must be invalidated if sceneCamera
used change of projection matrix (if camera lens changes for example)

It is a fix proposal for this bug report:
https://github.com/UPBGE/upbge/issues/1255


youle31:
Thanks! I merge then : )

Issue #1262, opened 2020-07-31
UPBGE 3.0 Action actuator PingPong playing animation in loop


rollalt:
ping pong is repeating the animation to infinity with out stoping so is opening and then closing and then again. Even if my animation is just the door opening.

youle31:
Fixed in https://github.com/UPBGE/upbge/commit/060a558648c8479c514d87195c60b01530b73f4c

The next builds will include this fix, with the hope it's the behaviour you were expecting.

PR #1263, opened 2020-07-31
UPBGE: Try to fix ping pong action #1262

youle31:
report: https://github.com/UPBGE/upbge/issues/1262

test file: https://pasteall.org/blend/f5301366acda487887a056d96767f309

There's a similar system for ACT_ACTION_LOOP_END and ACT_ACTION_FLIPPER.

- To stop to play the action, we convert the playmode from original playmode BL_Action::ACT_MODE_PINGPONG to BL_Action::ACT_MODE_PLAY.
When an action is in BL_Action::ACT_MODE_PLAY, its ending can be set (m_done is set to true in BL_Action.cpp).
- Furthermore, we set a flag to prevent the action to be played again:
m_flag |= ACT_FLAG_PLAY_END;
- For ping pong action, we need to choose the right moment to convert the pingpong action to BL_Action::ACT_MODE_PLAY. With the help of the debugger I choosed to set BL_Action::ACT_MODE_PLAY when obj->GetActionFrame(m_layer) >= end - 1 if there is no positive event, and if the pingpong action has not "continue" option enabled.

Notes:
- The same bug is in 0.2.5.
- This is a bit tricky to find the right behaviour...


youle31:
This is not the same behaviour in official 2.79...

Official 2.79 file: https://pasteall.org/blend/8c827f45fc4147a58f9976ce22352279

Maybe this is the behaviour we'd want to have... idk

source/gameengine/Converter/BL_ActionActuator.cpp
```diff
@@ -170,6 +170,14 @@ bool BL_ActionActuator::Update(double curtime)
...
             obj->SetActionFrame(m_layer, m_localtime);
           }
+          else {
+            obj->SetPlayMode(m_layer, BL_Action::ACT_MODE_PLAY);
+            m_flag |= ACT_FLAG_PLAY_END;
+            // Swap the start and end frames
+            float temp = m_startframe;
+            m_startframe = m_endframe;
+            m_endframe = temp;
+          }
```
lordloki, line 180: We need to separate from LOOP_STOP and LOOP_END as we get a ping pong behaviour for loop stop/end this way.

Additionally, If we need a LOOP_PINGPONG case we need to add a new case instead of use continue.

This below modifications based in your code makes the work:
image

lordloki: Good work!!!

youle31:
Thanks very much, I merge then

PR #1266, opened 2020-08-03
UPBGE: Change active cam when there's an overlay camera in MouseFocusSensor

youle31:
This way, the events like mouseOver... are computed in the Overlay Camera space.

Test file: https://pasteall.org/blend/b9059337f377478fbbf0634b07a67688

This needs more thinking and checks (about suspended things, best place
to Set Active Camera/ Restore old Active Camera)

but this seems to work in a simple example:

https://gifyu.com/image/Wojh

Finally, I think it is ok like that, as it is the user who controls suspend and restore physics/logic/visibility.
I don't think it needs more checks. If there's an issue, I think there are many chances that the user logic bricks setup is wrong.

Unless I forgot something. I tested with this file, and it seems to work ok for me:

https://pasteall.org/blend/94fe2f71b6414a779578072f5f5d38ed

Here we suspend collection 1 physics and logic (mouselook) and we add the HUD. When we are in the HUD with mouse over Suzanne, Suzanne rotates. If we mouse over Suzanne + left click, we resume collection 1 physics and logic (mouselook) and we remove HUD.

I think with this patch we are very near from overlay scenes behaviour, except that it is to the user to choose what he wants to suspend (logic/physics/visibility) when he adds an overlay collection.

OR we make this behaviour as optional. If (scene->GetOverlayCamera() && scene->GetOverlayCamera()->WantsToTakeControlOfEvents()) -> Set it as active.
I think it would be better to have this behaviour as optional finally, then i'll add an option in RNA_camera


youle31:
Thanks very much!

PR #1269, opened 2020-08-08
 UPBGE: Try to fix freeze when we run files with fluids

youle31:
Fluid: https://pasteall.org/blend/bca2c52db77b49068ac069c694dd38ba
Smoke: https://pasteall.org/blend/b5579b5a9649426b9427f9d297503b24

When we call ID_RECAL_TRANSFORM on orig_objects with a fluid modifier or a
smoke modifier, it enables heavy computations in
BKE_fluid_modifier_do...

When we only change ob_eval obmat, it's ok. Except that smoke doesn't seem to
move, but at least there's no freeze.

There are still some issues with blenderplayer but part of issues can be solved if we indicate the absolute path of the fluid cache.

Else, there's a warning displayed in blenderplayer which says that bpy_import_main_set has not been called. It can be a threading issue but I'm not sure for now. There are maybe some things that we can pick from bpy_interface to try to understand.

But in a first time, I propose to commit this which fixes freezes in embedded.

We could see later if we do something about pathes and this warning in blenderplayer.

### DIFF
--- a/source/gameengine/Ketsji/BL_Action.cpp
+++ b/source/gameengine/Ketsji/BL_Action.cpp
@@ -437,7 +437,7 @@ void BL_Action::Update(float curtime, bool applyToObject)
       if (con) {
         if (ob->adt && ob->adt->action->id.name == m_action->id.name) {
           if (!m_obj->OrigObCanBeTransformedInRealtime(ob)) {
-            return;
+            break;
           }
           DEG_id_tag_update(&ob->id, ID_RECALC_TRANSFORM);
           PointerRNA ptrrna;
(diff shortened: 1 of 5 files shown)


youle31:
No, I'll try to fix some things before asking for review

youle31:
Well, i still have freezes sometimes in blenderplayer, but when we give the absolute path for the cache, most of time it works both for liquid and smoke (we can see smoke).

I don't get why we have a warning in the console.

For me it is ok to commit as experimental thing...

We could eventually add a note about that in release notes (choose an absolute path for the cache)

youle31:
Great! Thanks for review and approval : )

Thanks about the note about the commit for fluid cache pathes.

Nice video!

I merge then, thanks!

PR #1271, opened 2020-08-19
UPBGE: Add an option to display only physics debug lines

youle31:
without to have to enable the framerate.

I also removed 1 DebugDraw per scene as it is not needed anymore.

I didn't understand how it is possible that there is a bridge between
RAS_DebugDraw and ccdPhysEnv debug drawer: (new BlenderDebugDraw());
I'm a bit confused but... it works


lordloki (approved):
I tested and it worked great!!! Good job!

youle31:
Thanks very much for review! I merge that even if master is a bit broken

PR #1275, opened 2020-08-27
UPBGE: Force viewport restrict flag for objects in inactive list to avoid perf issue

youle31:
It is fixing a performance issue in a special case reported by donatello on discord.

test file:

https://pasteall.org/blend/a3a087cb474240529ec3bc06dbb52639

Note the performance difference when we uncheck viewport icon on
Thing_suit object.

Note: Both in this branch and in master, when we run this file in blenderplayer, there are 19 memory leaks reported whereas normally there are only 3. We can check later if something is missing in gpg_ghost.


lordloki (approved):

youle31:
Thanks very much! I merge this

PR #1278, opened 2020-08-30
UPBGE: Try to restore post_draw callbacks

youle31:
With recent eevee's refactor, there were issues with post draw.

I tried to have the render in this depth order:

1) scene 2) physics debug 3) post_draw 4) profile drawing

It needs to switch between matrix and depth tests but seems to work with this file:

https://pasteall.org/blend/4671296344af4604ad36af9845ead5a1

Note: I removed the pre/post draw callbacks from blender because they are now using an overlay-fb which is not in use in bge draw loop.

Picture:

Capture

### DIFF
--- a/source/blender/draw/intern/draw_manager.c
+++ b/source/blender/draw/intern/draw_manager.c
@@ -2976,25 +2976,6 @@ EEVEE_Data *EEVEE_engine_data_get(void)
   return data;
 }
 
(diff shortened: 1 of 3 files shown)


lordloki (approved):
I tested and it works like a charm. Good job!!!

youle31:
Thanks very much for review @lordloki !

PR #1279, opened 2020-08-30
UPBGE: Add an option to override logic transform with depsgraph trans

youle31:
As the depsgraph does many things automatically, there are some cases
where the user (donatello) wants to use depsgraph object transform
instead of logic transform.

Donatello shows a file with drivers where an object transform was
driving another object transform:
https://pasteall.org/blend/aee620a478e84b2594ff6151e5c03a9a

In this case (and surely other cases (constraints...), the user might
want to be abled to use depsgraph transform instead of the scene graph
transform (transformation with logic/physics/bge scripts/bge
constraints...)

This is why we can add an option in object tab to let the user choose if
he wants to let the depsgraph do the transformation instead of bge.

Capture


lordloki (approved):
I tested and it works like a charm. Very good job!!

Code seems me ok too.

youle31:
Thanks very much for review and test! I merge this on master (even if master is a bit broken. blenderplayer not working for now)

PR #1281, opened 2020-09-08
UPBGE: Use directly blender drawing code for embedded and blenderplayer

youle31:
FOR LATER:

This will be for the day when we'll need to say goodbye to 2D filters and VideoTextures and overlay collections (because of Vulkan transition), Then it is not to review right now.

However, some things can be done waiting for this day if we want:

- Cleanup (Remove not needed code: commented code, remove overlay collections actuator, check load .blend actuator and restart scene actuator)
- Profile drawing and physics line: Find a way to insert the code in eevee's main render loop (physics lines before blender post_draw callbacks, profile after post_draw callbacks in DRW_draw_callbacks_post_scene

... Don't hesitate to fill if you have other ideas (reserved to developers)


youle31:
I close this for now, i'm not sure it will be needed

PR #1284, opened 2020-09-15
UPBGE: Few tweaks

youle31:
I was reviewing a bit the code and I changed few things.

1. Previously, in some scenes, when taa_tot_samples was not set to 0, there were some small artifacts/pixels moving, then I choosed to force this setting at ge start. It seems that now we can keep viewport taa_tot_samples setting.
It has several advantages:
- The image looks more like in the viewport.
- When taa_tot_samples was set to 0, there were some weird behaviours sometimes (on a pine tree for example, the spikes were more and more thick when the image was not moving)
- The more we were waiting on a static image, the more the framerate was slowing whereas now, the GPU seems to reduce its activity when we reach tot_samples.

2. I noticed that some parts of the drawing code were called whereas it seems not needed with bge render loop then I did an early return for this code when scene->flag & SCE_INTERACTIVE.

3. I noticed that in embedded, calling ED_area_tag_redraw at ge start was avoiding some UI flickering, then I added that in the embedded ge start operator.

4. I noticed that in WM_init_game, the screen area chosen to start ge was choosed with BKE_screen_find_big_area(CTX_wm_screen(C), SPACE_VIEW3D, 0). Then I modified the ReinitContextVariables code to always choose the "big" area too to init context variables.

I tested in few files and it sounds to work well for me

### DIFF
--- a/source/blender/draw/intern/draw_manager.c
+++ b/source/blender/draw/intern/draw_manager.c
@@ -1299,6 +1299,10 @@ void DRW_notify_view_update(const DRWUpdateContext *update_ctx)
   Scene *scene = update_ctx->scene;
   ViewLayer *view_layer = update_ctx->view_layer;
 
+  if (scene->flag & SCE_INTERACTIVE) {
+    return;
+  }
+
   const bool gpencil_engine_needed = drw_gpencil_engine_needed(depsgraph, v3d);
 
   /* Separate update for each stereo view. */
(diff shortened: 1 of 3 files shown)


youle31:
When you'll have a bit of time to test, can you tell me if it sounds safe enough / and if it's working for you lordloki?

lordloki (approved):
The changes work great!! Now, the "pulse" when you stop almost is unoticiable. Additionally, when you set a viewport samples not equal to 0 you have much better performance when you are not moving.

+1000

youle31:
Great! Thanks very much for review and tests!

PR #1288, opened 2020-09-27
UPBGE: Restore bge softbodies (experimental)

youle31:
It also includes few fixes related to physics but not to softbodies.

I tested with few files, we need to apply rotation and scale before pressing P, but it was like that too in 0.2.5.

This seems to work ~as in 0.2.5 but I didn't spend enough time on the code or to test.

But this can be commited as an experimental thing though.


youle31:
arf, it works with my old test files but not anymore with new test files...

lordloki (approved):
Great work!!!
I tested in several cases and it worked as in 0.2.5. I only found a case where it seems to freeze the game engine. That case is using a object that is not totally closed or formed by several separated meshes.
You can test it using suzanne with softbodies.

The other code related other physics fixes seems ok to me.

As resume, we can commit the code and try to solve the freeze pointed above later (I have to make debug builds to see where the freeze or the infinite loop is).

youle31:
Ah yes indeed I can reproduce the freeze too with Suzanne. Thanks very much for review!
I'll try to put breakpoints too to see if I find something.

youle31:
Thanks very much! I merge like that and we can eventually investigate on collision flags later.

PR #1290, opened 2020-10-02
UPBGE: Fix Vertex vs face soft vs soft handling collision

lordloki:
We use type VF_DD now to avoid crashes with objects like Suzanne with several separate parts.

Additionally, we recover bending distance slide option (default to 2).

2 samples:
[Softbody_simple.zip](https://github.com/UPBGE/upbge/files/5319088/Softbody_simple.zip)
[SoftBody_flag.zip](https://github.com/UPBGE/upbge/files/5319089/SoftBody_flag.zip)

### DIFF
--- a/release/scripts/startup/bl_ui/properties_game.py
+++ b/release/scripts/startup/bl_ui/properties_game.py
@@ -159,7 +159,7 @@ class PHYSICS_PT_game_physics(PhysicsButtonsPanel, Panel):
 
             sub = col.column()
             sub.active = soft.use_bending_constraints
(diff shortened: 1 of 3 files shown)


youle31:
I tested both files and it works fine. The code seems ok. Thanks very much!

(I removed bending_distance because I copied the settings from upbge 0.2.5 and it wasn't in RNA so it must be a 0.2.5 bug, but I saw the effect of this setting when I tested your file. Great!)

youle31 (approved):

lordloki:
Great! Thanks for reviewing

Issue #1291, opened 2020-10-06
Transformations for scaled and parented objects 


MechatronixX:
Try this to reproduce the issue. Add two cubes and parent them. Use a motion actuator in UPBGE game logics to rotate the parent around the x-axis. Everything seems fine. Now re-scale the parent around the x-axis and re-run. The child now makes an elliptical, or at least non-circular, motion around the parent cube. 

As this is something that doesn't happen in Blender when rotating the parent, I guess it is either a bug or should at least something to be considered undesirable. 

I have observed this for UPGE 0.3, both in Windows and Linux.

youle31:
Could you follow this guide: https://github.com/UPBGE/upbge/wiki/Report-a-bug-guide to report an issue.

- In your case, could you join an upbge 0.2.5 test file and an upbge 0.3 test file to compare?
- Could you describe the exact steps to reproduce the issue? (you can do a gif with screenToGif for example and give us a link to it here).

Thanks.

MechatronixX:
So this is an example of what I mean. See how the motion is skewed when re-scaling the objects. Even if this for some reason is an intended behavior, at least IMO it shouldn't be. It seems that the object scale factors are involved in some way they shouldn't, re-scaling the relative position vector between the objects. If you can give some pointers about where to start looking in the code,  I can see if I can do anything about it, would be fun :) 

I am using Ubuntu 20.04 with UPGE 0.3 alpha from 2020-10-07. 

Here is a file you could use to reproduce the issue. I guess the gif is of more interest though as it outlines all the steps I took. 
[transform_issue_alpha_03.zip](https://github.com/UPBGE/upbge/files/5366852/transform_issue_alpha_03.zip)

![transform_issue](https://user-images.githubusercontent.com/37342798/95777872-504a3d80-0cc7-11eb-9c57-ec95393a0bb1.gif)

My current workaround to this is to manually recalculate the positions and orientations in a script.

lordloki:
To fix the rotation after scale, you have to apply rotation & scale on central object
![image](https://user-images.githubusercontent.com/1015605/95782625-77593d00-0cd0-11eb-88f1-aecd07d49c89.png)

MechatronixX:
Awsome that a pull request is up already! :)

PR #1306, opened 2020-10-12
UPBGE: Try to fix a weird transformation behaviour with children

youle31:
Note:

I picked what was in upbge 0.2.5 and tried to transpose to Moto with the
help of that:

https://math.stackexchange.com/questions/237369/given-this-transformation-matrix-how-do-i-decompose-it-into-translation-rotati

This is to fix
https://github.com/UPBGE/upbge/issues/1291#issuecomment-707280616

my test file is: https://pasteall.org/blend/0d887e6101fc405e9569d6ca480f9967

I didn't tested elsewhere than in this test file for now, and i didn't look how it was in upbge 0.2.5 but it seems something was done in 0.2.5

### DIFF
--- a/source/gameengine/Ketsji/KX_SG_NodeRelationships.cpp
+++ b/source/gameengine/Ketsji/KX_SG_NodeRelationships.cpp
@@ -31,6 +31,9 @@
 
 #include "KX_SG_NodeRelationships.h"
 
+#include "MT_Matrix4x4.h"
+#include "MT_Transform.h"
+
 
 /**
  * Implementation of classes defined in KX_SG_NodeRelationships.h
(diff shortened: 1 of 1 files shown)


source/gameengine/Ketsji/KX_SG_NodeRelationships.cpp
```diff
@@ -62,15 +65,20 @@ bool KX_NormalParentRelation::UpdateChildCoordinates(SG_Node *child,
...
-
-    child->SetWorldScale(p_world_scale * child->GetLocalScale());
-    child->SetWorldOrientation(p_world_rotation * child->GetLocalOrientation());
-    child->SetWorldPosition(p_world_pos +
-                            p_world_scale * (p_world_rotation * child->GetLocalPosition()));
+    const MT_Transform trans(parent->GetWorldTransform() * child->GetLocalTransform());
+    MT_Matrix4x4 tmat(trans.toMatrix());
+    float sx = MT_Vector3(tmat.getElement(0, 0), tmat.getElement(1, 0), tmat.getElement(2, 0)).length();
+    float sy = MT_Vector3(tmat.getElement(0, 1), tmat.getElement(1, 1), tmat.getElement(2, 1)).length();
+    float sz = MT_Vector3(tmat.getElement(0, 2), tmat.getElement(1, 2), tmat.getElement(2, 2)).length();
```
MechatronixX, line 72: Cannot use tmat.GetWorldScaling() instead?

source/gameengine/Ketsji/KX_SG_NodeRelationships.cpp
```diff
@@ -62,15 +65,20 @@ bool KX_NormalParentRelation::UpdateChildCoordinates(SG_Node *child,
...
   else {
-    // the childs world locations which we will update.
-    const MT_Vector3 &p_world_scale = parent->GetWorldScaling();
-    const MT_Vector3 &p_world_pos = parent->GetWorldPosition();
-    const MT_Matrix3x3 &p_world_rotation = parent->GetWorldOrientation();
-
-    child->SetWorldScale(p_world_scale * child->GetLocalScale());
-    child->SetWorldOrientation(p_world_rotation * child->GetLocalOrientation());
-    child->SetWorldPosition(p_world_pos +
-                            p_world_scale * (p_world_rotation * child->GetLocalPosition()));
```
MechatronixX, line 73: Isn't the p_world_scale multiplication here the issue? If I am not wrong removing that multiplication would make this more or less equivalent of the pull request, because the GetXXXTransform() returns a scaled transform, while the GetXXXPosition and GetXXXPosition return un-scaled data.

MechatronixX (commented):
So nice that this already got looked into! :) 

What about a simple unittest also? Verifying that we can rotate the cube around the origin and that it would go in a circular pattern? 


In the long run I think it would be nice with a new attribute 'pose' which would return the unscaled transformation matrix. This could be used for considering the 6DOF pose only.

source/gameengine/Ketsji/KX_SG_NodeRelationships.cpp
youle31: I did tests removing the p_world_scale but it didn't fix totally the issue.

source/gameengine/Ketsji/KX_SG_NodeRelationships.cpp
youle31: tmat has no function GetWorldScaling.
I did as in upbge 0.2.5: get a transform matrix and extract position, rotation, and scale. In 0.2.5, there was a method to extract directly the scale but in 0.3 the maths lib is Moto, and i had to do it following the article i mentionned in the pull request.

lordloki:
Hi,

I tested using all the Parent relations files and it worked like a charm. Good job!!!
[ParentRelation-2.8.zip](https://github.com/UPBGE/upbge/files/5374048/ParentRelation-2.8.zip)

lordloki (approved):

youle31:
Great! Thanks very much for tests!

I merge that then.

PR #1308, opened 2020-10-17
UPBGE: Restore custom bge viewports

youle31:
minimap test file: https://pasteall.org/blend/97f46d6903ab42cda589ba552b2b635c

I tested with few files to see if i didn't break something else, but it sounded ok. Few more tests are welcome.


lordloki (approved):
I have tested with this file and it worked ok. Also, I tested with normal files to see possible issues and everything was ok.

Good job!!

youle31:
Great! Thanks very much for tests! I had no other test file for viewports but we'll see later if there are issues. Thanks!

PR #1310, opened 2020-10-21
UPBGE: ReplaceMesh Actuator: Restore replace physics shape option

youle31:
test file: https://pasteall.org/blend/cd5635c360c64f7f86a9502e3be82a5a

It is using ReplacePhysicsShape whereas before I think it was using ReinstancePhysicsShape.

I think it is better because it just replace the shape with an already existing shape without having to do computation.

### DIFF
--- a/source/gameengine/Ketsji/KX_Scene.cpp
+++ b/source/gameengine/Ketsji/KX_Scene.cpp
@@ -1673,6 +1673,42 @@ void KX_Scene::ReplaceMesh(KX_GameObject *gameobj,
     return;
   }
 
+  /* Update physics mesh. Do it before use_gfx, else it won't work.
+   * (It has to be done before we call gameobj->removeMeshes() + gameobj->AddMesh(mesh)).
+   */
+  if (use_phys) {
+    if (gameobj->GetPhysicsController()) {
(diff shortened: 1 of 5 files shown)


source/gameengine/Ketsji/KX_Scene.cpp
```diff
@@ -1691,13 +1729,6 @@ void KX_Scene::ReplaceMesh(KX_GameObject *gameobj,
     DEG_id_tag_update(&gameobj->GetBlenderObject()->id, ID_RECALC_GEOMETRY);
```
lordloki, line 1729: When you use several cubes and mix "only gfx" and "only phys" depend on the cube the it is necessary to modifiy the tag_update function and move it to the end:

```
  if (use_gfx || use_phys) {
    DEG_id_tag_update(&gameobj->GetBlenderObject()->id, ID_RECALC_GEOMETRY);
  }
```

lordloki (approved):
Great work!!!!!
Replace mesh work very well. 

I tested using your test file and when you do several cube duplicated and mix "use_gfx" and "use_phys" on them there is an issue (i think due to comment on code i pointed)

youle31:
Thanks very much!

PR #1312, opened 2020-10-25
UPBGE: Implement viewport render in blenderplayer + profile in viewport render (experimental)

youle31:
For now, I allowed only one scene because:
- Replace scene in blenderplayer viewport render generates mem leaks
- There are other few issues related to shading...

I didn't change the behaviour of viewport render in embedded.

In blenderplayer, we can use lookdev and eevee but not workbench (Because if we do nothing to change the shading type at ge start, workbench and lookdev work but not eevee...)

In blenderplayer, I hide overlays (grid, outlines...) drawing.

2D filters, ImageRender, overlay scenes, custom viewports... are not available in viewport render mode.

But stereo works and I guess virtual reality too and this is the main advantages.

Note: Why do we "need" 2 render implementations:

- Because old bge renders are done with cameras (for ImageRender, custom bge viewports, overlay collections...)
Then we choosed to use our own GPUViewports based on cameras settings.
- Because we need our own render loop to have things like overlay collections... working.

Blender render loop has its own GPUViewports but they are not based on Cameras, they are attached to ARegion (ARegion->draw_buffer->viewport[view] iirc). So this is not the same system than bge.
When we use "viewport render" option, we reuse these GPUViewports and we haven't our own GPUViewports.
It is easier to do like that but we can't have the old bge specificities (at least without work).

### DIFF
--- a/source/blender/draw/intern/DRW_render.h
+++ b/source/blender/draw/intern/DRW_render.h
@@ -789,6 +789,7 @@ void DRW_game_render_loop(struct bContext *C,
 
 void DRW_game_render_loop_end(void);
 void DRW_game_python_loop_end(struct ViewLayer *view_layer);
+void DRW_game_viewport_render_loop_end(void);
 void DRW_transform_to_display(struct GPUTexture *tex, struct View3D *v3d, bool do_dithering);
 void DRW_transform_to_display_image_render(struct GPUTexture *tex);
 /**************************END OF GAME ENGINE*******************************/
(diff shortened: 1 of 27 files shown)


lordloki (approved):
I reviewed the code and it seems ok to me. I did the regular tests with my files and it worked like a charm.
As I said before in discord I think we can commit it even with the shortcomings because the great improvements. Later we will see if we can add the rest of features or not.

Great Work!!!!

youle31:
Thanks very much! Let's merge then!

PR #1316, opened 2020-11-01
Rigidbody/Dynamics continuous collision detection

lordloki:
A new option is available per object to set-up the continuous collision detection. There are 2 new properties:
- MotionThreshold: Sets the delta of movement that has to happen in one physics tick to trigger the continuous motion detection.
- SweptSphereRadius: The radius of the sphere that is used to check for possible collisions when ccd is actived.

image

[ccd_bullet.zip](https://github.com/UPBGE/upbge/files/5471905/ccd_bullet.zip)


BluePrintRandom:
\0/

IzaZed:
Very nice! 🥳

youle31:
Hello, I tested and it is true that in your test file, these options improve collision detection of the red bullet and it is great.

In the same time, I can't figure out how to have the same behaviour than in 0.2.5 using these options in this 0.2.5 file:
https://pasteall.org/blend/d16cb0db45a14826893e03d831c93317

Concerning the code, I only have minors/optionals remarks:

- In CCDPhysicsController.h, to avoid a warning during compilation we can move this after float m_radius; 
Capture

- Maybe it would be possible to have the correct defaults for both these options (1.0f, and 0.9f) in the following cases:

1) For the default existing objects (the default cube) (here idk exactly how to do - I thought about versioning-default but it seems it is not working))
2) For the files saved with upbge when these options were not existing (versioning_upbge)
3) For the files saved in Blender 2.7/2.8/2.9 official (in versioning_2.80 there is a hack for this case)
4) When we add a new object (object_init_data in object.c)

- Optionally, errors messages can be added in python methods (even if there are not error messages everywhere for similar methods): like: PyErr_SetString(PyExc_TypeError, "gameOb.setCcdMotionThreshold: KX_GameObject, expected a float in range 0.0 - 100.0");

Conclusion: I'm ok with this patch because it can improve the current collision detection, but in the same time I'm still wondering if there would be something to do to have the same behaviour than before bullet update in my test file.

Anyway, Thanks very much for your work!

youle31 (approved):

lordloki:
Thanks for reviewing, youle!!!! I pushed a fix for all.

Regarding to achieve same behaviour than before i think it is a problem of bullet git master vs bullet oficial releases. I think there is something in git master that makes bullet to have worst behaviour. 

I'm going to replace bullet library in my git folder with the last release of bullet to check it.

youle31:
Awesome Fixes commit! Thanks!

PR #1317, opened 2020-11-06
UPBGE: Restore old collision behaviour

lordloki:
We restore old m_erp2 value.
Thanks to @youle31 for investigating and pointing the solution. I checked the value for bullet 2.83.7 and older and it was 0.8. If you see that the behaviour is better with 1.0 with can change it to that value.

Additionally, we move back to float precision instead of double in Bullet library

Double precision is slower than float one because the float one can use SSE improvements. WE have around 5-15% better performance with float precision.

[stress_test.zip](https://github.com/UPBGE/upbge/files/5498100/stress_test.zip)

Double precision can be interesting in open world with more than 10Km from center.

### DIFF
--- a/doc/python_api/rst/bge.constraints.rst
+++ b/doc/python_api/rst/bge.constraints.rst
@@ -162,6 +162,40 @@ Functions
    :arg time: The deactivation time.
    :type time: float
 
+.. function:: setERPNonContact(erp)
+
+   Sets the Error Reduction Parameter (ERP) for non-contact constraints.
+   The Error Reduction Parameter (ERP) specifies what proportion of the joint error will be fixed during the next simulation step.
+   If ERP = 0.0 then no correcting force is applied and the bodies will eventually drift apart as the simulation proceeds.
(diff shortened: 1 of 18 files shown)


BluePrintRandom:
also double is probably good for stuff with very small parts / simulations

lordloki:
I found this:
The ERP specifies what proportion of the joint error will be fixed during the next simulation step.
If ERP=0 then no correcting force is applied and the bodies will eventually drift apart as the simulation proceeds.
If ERP=1 then the simulation will attempt to fix all joint error during the next time step.
However, setting ERP=1 is not recommended, as the joint error will not be completely fixed due to various internal approximations.
A value of ERP=0.1 to 0.8 is recommended (0.2 is the default).

I'm going to expose ERP and CFM values through bullet scene settings and python functions. I will update this pull request

lordloki:
Capsule height fixed for new bullet version
image

youle31:
I approve all the changes, if you want to merge. Unless you have other things to add in this pull request. Thanks for your work!

youle31:
+10000000

PR #1325, opened 2020-11-19
UPBGE: Fix for XR start in master

youle31:
test file: https://pasteall.org/blend/7d317f4eab18465b8e9d4ca0c3f69dfa

(It is just to not revert
https://github.com/UPBGE/upbge/commit/2dd60e6c2cbdf3e39ed3451ee0f67a5b00a7a5bc
but avoid crash at ge start)


lordloki (approved):
I tested and everything was OK!
Much better solution that my proposed fix. :-)
Good work!

youle31:
Thanks very much for review and test! I merge

PR #1329, opened 2020-11-23
UPBGE: Fix endObject() didn't free blenderObject for converted objects during runtime

youle31:
convertBlenderObject() function didn't tag the gameobject as a replica object. As only replica blenderObjects are freed, when we were calling endObject() for a converted object during runtime, the blenderObject was not freed.

This patch is an attempt to fix this issue.

I had to deal with other specificities related to objects converted during runtime and this result with something a bit experimental.

This is the reason why before merge, it would be nice if you @BluePrintRandom , @mysticfall , @lordloki could test it with your files using convertBlenderObject/convertBlenderCollection/convertObjectList and tell me if it works for you, when you will have a moment to test.

Also don't hesitate if you have remarks about the code.

Thanks!

### DIFF
--- a/source/gameengine/Converter/BL_BlenderDataConversion.cpp
+++ b/source/gameengine/Converter/BL_BlenderDataConversion.cpp
@@ -1169,7 +1169,6 @@ void BL_ConvertBlenderObjects(struct Main *maggie,
         blenderobject, kxscene, rendertools, converter, libloading, converting_during_runtime);
 
     if (gameobj && converting_during_runtime) {
-      gameobj->SetIsConvertedDuringRuntime();
       gameobj->SetIsReplicaObject();
     }
 
--- a/source/gameengine/Converter/BL_BlenderDataConversion.cpp
+++ b/source/gameengine/Converter/BL_BlenderDataConversion.cpp
@@ -1168,6 +1168,11 @@ void BL_ConvertBlenderObjects(struct Main *maggie,
     KX_GameObject *gameobj = BL_gameobject_from_blenderobject(
         blenderobject, kxscene, rendertools, converter, libloading, converting_during_runtime);
 
+    if (gameobj && converting_during_runtime) {
+      gameobj->SetIsConvertedDuringRuntime();
+      gameobj->SetIsReplicaObject();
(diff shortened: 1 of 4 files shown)


mysticfall:
I just tested it with the old version of my project which had caused this problem. Unfortunately, it looks like it's still crashing with the same error message when `endObject()` is called on a created object:
```
AttributeError: 'PyObjectPlus' object has no attribute 'update'
```

The relevant code snippet reads as follows:
```python
prototype: Object = bpy.data.objects[asset.title]
instance: Object = prototype.copy()

instance.hide_viewport = False
instance.hide_render = False
instance.data = instance.data.copy()

assets_root.objects.link(instance)

scene = bge.logic.getCurrentScene()
obj = scene.convertBlenderObject(instance)
```

youle31:
However, my file adapted from this code snippet doesn't crash for me: https://pasteall.org/blend/0051dac56777443baf1dd089b95516db and the blenderObject is now freed contrarly to before.

I could eventually try to debug if I had the whole project (the whole code + .blend). Thanks!

mysticfall:
Maybe we should just merge the PR, if others don't see the problem?

My project no longer contains the code that used to cause the crash and it's in a quite chaotic state now, so it wouldn't be easy to make a minimal reproducible case out of it.

I'll report an issue if I see it happens again in future. Thanks!

lordloki:
Hi,
i didn´t have much time to test but this blend got crash on exit when converted several times (press spacebar several times).
I will test more tomorrow

[Convert_async_cubes_2.8.zip](https://github.com/UPBGE/upbge/files/5594636/Convert_async_cubes_2.8.zip)

youle31:
ok, Thanks for testing and for the .blend! i'll try to debug too when i'll be motivated

lordloki:
Hi, I tested with all my files and everything works perfectly!!.
I think you can commit it without any problem. Good work!!

lordloki (approved):

youle31:
Awesome! Thanks very much for hard testing! I tested too with many files and didn't notice issue. I merge then. Thanks again!

PR #1335, opened 2020-11-27
UPBGE: Try something to reduce leaks count in blenderplayer

youle31:
it reduce reported mem leaks from 3 to 1


lordloki (approved):
No issues observed when i tested my test files.
Good work!!!

youle31:
Great! Thanks very much for testing!

PR #1343, opened 2020-12-04
UPBGE: 4 extra buttons added for Mouse

lordloki:
Current gaming mouses use to have 4 or even 6 buttons normally. Now, we add to support for these buttons. 
Support is added for logic bricks and python


youle31:
I tested, buttons 6 and 7 don't work for me, but I think it is because the driver reserve these buttons to other things. Else it works fine, and the code seems ok.
Thanks for bringing us to the 21th century!

youle31 (approved):

lordloki:
Yeah, buttons 6 and 7 are for programmable buttons. I will commit now and if i dont find a way to use both buttons, i will remove them later.

lordloki:
Thanks for testing.

youle31:
I think it is not needed to remove it even if you don't manage to have it working. I saw that blender events were implemented in the same way than you did then it should be ok. If there's an issue, it comes from user side who has to deal with his own mouse software/drivers

PR #1346, opened 2020-12-08
Fix #1341: Call Py_Initialize() before using Python API

### LINKED ISSUE #1341: Blenderplayer crashes with Python 3.9, invoking _PyObject_GC_New
After my Manjaro unstable installation upgraded Python to 3.9, I can no longer run UPBGE project as it segfaults immediately with the following trace:
```
#0  0x00007fd728569927 _PyObject_GC_New (libpython3.9.so.1.0 + 0x10e927)
#1  0x00007fd72856bedc PyDict_New (libpython3.9.so.1.0 + 0x110edc)
#2  0x0000564722071c7b main (blenderplayer + 0xf40c7b)
#3  0x00007fd727f97152 __libc_start_main (libc.so.6 + 0x28152)
#4  0x00005647220cacde _start (blenderplayer + 0xf99cde)
```

mysticfall:
It fixes #1341 on my end but I'm not entirely confident that it doesn't contain any error because I'm not familiar with the language or the environment. So, it'd be advisable to merge this after sufficient code review and testing.

### DIFF
--- a/source/gameengine/GamePlayer/GPG_ghost.cpp
+++ b/source/gameengine/GamePlayer/GPG_ghost.cpp
@@ -117,6 +117,7 @@
 
 #include "CM_Message.h"
 #include "KX_Globals.h"
+#include "KX_PythonInit.h"
 #include "LA_PlayerLauncher.h"
 #include "LA_SystemCommandLine.h"
 
@@ -1262,6 +1263,10 @@ int main(int argc,
 #endif  // WITH_PYTHON
 
         DRW_engines_register();
+
+#ifdef WITH_PYTHON
+        initGamePlayerPythonScripting(argc, argv, C);
+#endif
 
         bool first_time_window = true;
 
--- a/source/gameengine/Ketsji/KX_PythonInit.cpp
+++ b/source/gameengine/Ketsji/KX_PythonInit.cpp
@@ -2131,16 +2131,8 @@ static struct _inittab bpy_internal_modules[] = {
  * Python is not initialized.
  * see bpy_interface.c's BPY_python_start() which shares the same functionality in blender.
  */
-void initGamePlayerPythonScripting(Main *maggie, int argc, char **argv, bContext *C, bool audioDeviveIsInitialized)
+void initGamePlayerPythonScripting(int argc, char **argv, bContext *C)
 {
-  /* Yet another gotcha in the py api
-   * Cant run PySys_SetArgv more than once because this adds the
-   * binary dir to the sys.path each time.
-   * Id have thought python being totally restarted would make this ok but
-   * somehow it remembers the sys.path - Campbell
-   */
-
-  static bool first_time = true;
   /* Needed for Python's initialization for portable Python installations.
    * We could use #Py_SetPath, but this overrides Python's internal logic
    * for calculating it's own module search paths.
@@ -2192,6 +2184,17 @@ void initGamePlayerPythonScripting(Main *maggie, int argc, char **argv, bContext
 
   /* Initialize Python (also acquires lock). */
   Py_Initialize();
+}
+
+void postInitGamePlayerPythonScripting(Main *maggie, int argc, char **argv, bContext *C, bool audioDeviveIsInitialized)
+{
+  /* Yet another gotcha in the py api
+   * Cant run PySys_SetArgv more than once because this adds the
+   * binary dir to the sys.path each time.
+   * Id have thought python being totally restarted would make this ok but
+   * somehow it remembers the sys.path - Campbell
+   */
+  static bool first_time = true;
 
   if (argv && first_time) { /* browser plugins don't currently set this */
     // Until python support ascii again, we use our own.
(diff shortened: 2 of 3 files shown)


youle31:
It is hard to review and sounds risky at first glance, but I think it is ok, even if I can't 100% guarantee there won't be bad side effect.

I share the patch (because I had to clone your github to make the diff): https://pastebin.com/E65U44sn

And my test file for globalDict: https://pasteall.org/blend/3fc378c6b6df48a09139bea18446f343

I tested with several other files and it seemed ok. Then I approve, but i'd like lordloki review too.

Thanks very much for your work!

youle31 (approved):

youle31:
On a side note @mysticfall , you can directly create and push branches in upbge github to do pull requests if you don't mind, I find it easier to review.

Anyway, thanks again for your work!

youle31:
Ah, actually, I think it needs to check the exitRequest code before freeing the globalDict because it must persist if you load another .blend

source/gameengine/Launcher/LA_Launcher.cpp
```diff
@@ -287,9 +288,11 @@ void LA_Launcher::ExitEngine()
    * which allows the scene to safely delete them :)
    * see: (space.c)->start_game
    */
-
   PyDict_Clear(PyModule_GetDict(m_gameLogic));
 
+  PyDict_Clear(m_profileDict);
+  PyDict_Clear(m_globalDict);
```
youle31, line 294: The globalDict must only be freed if exit request is to quit game, not if you start another .blend.

I propose this diff with your changes + mine: https://pastebin.com/6MpmxRXi

youle31 (changes requested):

youle31:
My diff is not good neither because each time we load a new .blend, a new launcher is created but the globalDict must persist. Then it has to be done differently.

mysticfall:
> Ah, actually, I think it needs to check the exitRequest code before freeing the globalDict because it must persist if you load another .blend
I will do that next time. It'll be easier for me that way too :)

Thanks much for the swift test and review, both of you. After spending a bit more time on reading the existing code, now I think I have a better understanding of the intention regarding the exit code check.

I may try to butcher the code further to move the necessary initialisation block to `GPG_ghost` to restore the old behaviour, but I'm not entirely certain that it's the best approach. So, I'd like to hear your opinions before starting the additional work.

As I understand it, the part that I removed is needed because the old code allowed opening another `.blend` file and share Python object instances between them.

But it also reinitialize everything in the Python world, including reimporting all the modules (e.g. `bpy`, `bge`, etc.). The problem, however, is that it must happen before `Py_Initialize()` is invoked, according to what is commented on `KX_PythonInit.cpp` (it actually throws an error about `_bpy_path` if the order is reversed).

As far as I see it, this requirement is not compatible with the Python embedding guideline which mandates calling `Py_Initialize()` before doing anything with Python API, and that is actually what was causing #1341.

It looks like Python 3.8 and earlier was a bit more tolerant when you violate the guideline but it is no longer the case with the new version.

To me, it feels rather unnatural or even dangerous to keep live Python objects around when the whole Python world gets recreated between the .blend file loads. And from what I've searched over Google to find out the valid use case for switching .blend file during the gameplay, it looks like very few people actually use the feature (using different scenes and library overrides could be a better approach for the others, I suppose).

So, maybe we can just remove this feature and refactor the code to contain all the Python initialization processes in one place instead (as currently they are sprawled all over the place)?

If the engine itself doesn't rely on the shared dictionary instance, I think it will only cause minimal inconveniences for our users since there aren't not many people who rely on it and they still can switch between .blend files if they want, but implement some custom persistence code to share the game data instead.

I can try hacking the current code further to make things work as it is, but it will probably make the codebase regarding the initialisation process even messier than it already is, so I'd like to hear your opinions on this matter before I start.

source/gameengine/Launcher/LA_Launcher.cpp
mysticfall: > Also I'd add m_profileDict = nullptr; after PyDict_Clear(m_profileDict)
I was just trying to follow what the current code does. So, should I reset other global members like `m_profileDict` in the same way?

youle31:
i think your current patch is good for if we don't find another way to keep persistent data with globalDict. Sorry, i was a bit irritated because i don't understand why you want to use python 3.9. But yeah, if we have to do it for python 3.9, i'd prefer your current patch (even if it is breaking persistent data) than moving InitGamePlayerPythonScripting in gpg_ghost

mysticfall:
After discussing it with youle, I reverted most of the changes so that the problem can be fixed with absolute minimal modification of the existing codebase. It should restore the original behaviour with the globalDict persistence between .blend file loads also,

youle31:
I'm ok with the changes 👍  Thanks!

lordloki (approved):
I tested a lot python files and they work ok all of them.
This seems a much better solution than the first one. Good work!!

youle31 (approved):

mysticfall:
Thanks guys for the review and tests! :)

PR #1348, opened 2020-12-09
Fixes #1240: Add fake types for bge.types.*

### LINKED ISSUE #1240: Python component fails to recognize some BGE type names (e.g. KX_Camera)
When I create a simple Python module and add a reference to some of the builtin types like `bge.types.KX_Camera`, it fails to register with an error saying that it cannot find the given type under `types`.

It can be reproduced with the following source:
```python
import bge


class SimpleComponent(bge.types.KX_PythonComponent):

    camera: bge.types.KX_Camera

    def __init__(self, base_object):
        pass

    def start():
        pass

    def update():
        pass
```

mysticfall:
I wish I knew enough C/C++ to auto-generate the list from the source files, but at least now we can register components without resorting to such an ugly hack like this:
```python
try:
    from bge.types import KX_GameObject
except ImportError:
    KX_GameObject = None
```

### DIFF
--- a/source/blender/blenkernel/intern/python_component.c
+++ b/source/blender/blenkernel/intern/python_component.c
@@ -51,61 +51,144 @@
 
 #ifdef WITH_PYTHON
 
-PyDoc_STRVAR(class_documentation,
-             "This is the fake BGE class KX_PythonComponent from fake BGE module bge.types");
+#define FAKE_TYPES \
+    FT_DEF(BL_ArmatureBone) \
+    FT_DEF(BL_ArmatureChannel) \
+    FT_DEF(BL_ArmatureConstraint) \
+    FT_DEF(BL_ArmatureObject) \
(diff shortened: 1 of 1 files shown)


youle31 (approved):

youle31:
About code, I can't review (technically speaking) but it sounds good. If it works for you, then it's ok for me. I just note that we should think to update the types list if we add or remove one type.

Else, I tested with a component file and it worked, registration worked, then it's ok to me.

mysticfall:
> I just note that we should think to update the types list if we add or remove one type.

Yeah, I think we need to make the list automatically generated by the build process someday. The build tools I've been using all have such a feature so I'm positive that CMake/Make have one too.

lordloki (approved):
Tested and worked ok. Great work +10

mysticfall:
Thanks much for the quick review and approval :)

PR #1351, opened 2020-12-14
UPBGE: Add CEGUI GUI actuator

lordloki:
This work is based in a former work from @IMDagger that I found in the net (this work was from a 2.4x era). Thanks.
(http://bazaar.launchpad.net/~imdagger/+junk/blender-2.4x-imd/changes)

I reworked the code heavily because the compiling system, the bge code and CEGUI code are very different from the code that existed 9 years ago.

At this moment the code doesn't compile yet. I just created a pull request to can see all the commits joined and work out from home.


youle31:
I had a quick look at the files changed and I noticed that in Object.c there was BLO_write_struct but not the counterpart for reading. I think it is possible that there is missing code here. Maybe looking at other actuators "model" in object.c can help

PR #1363, opened 2020-12-28
UPBGE: Add "use soft shadow" per light switch

lordloki:
Based in https://developer.blender.org/D4166 by Alexander Court

image

Blend example: [test_lights.zip](https://github.com/UPBGE/upbge/files/5746018/test_lights.zip)

### DIFF
--- a/release/scripts/startup/bl_ui/properties_data_light.py
+++ b/release/scripts/startup/bl_ui/properties_data_light.py
@@ -178,6 +178,9 @@ class DATA_PT_EEVEE_shadow(DataButtonsPanel, Panel):
 
         col.prop(light, "shadow_buffer_bias", text="Bias")
 
+        col.separator()
+        col.prop(light, "use_soft_shadows", text="Soft Shadows")
+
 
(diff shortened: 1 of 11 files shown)


lordloki:
Maybe I have to put the checkbox at Contact Shadows level

youle31:
I think all these settings should be per lights, but as it is not us who decide for render things, if it is just this patch, it's not invasive and doesn't cause too much potential issues for merges, but if we include cascade size and cube size per light, and if we do even more experiments on shadows, we should have our own shadows pipeline in separate files to avoid conflicts with blender code.

If it is just for this patch, i'm ok to merge as it is.

youle31 (approved):

youle31:
also, should the global setting in render tab modify the scene lights options?

lordloki:
Yeah, i see it strange too but i think they focused all the related "render" setting in a closer place to improve the workflow. 

Well, i commit this that has little source impact. After, if we change more things we will see how to create our own pipeline if necessary.

lordloki:
Arg, i forgot in the commit log. Thanks for the review youle.

youle31:
No problem, Thanks to you! "Soft shadow override" makes it more clear too 👍

PR #1365, opened 2020-12-29
UPBGE: Allow to change .blend in embedded in "viewport render mode"

youle31:
it contains several changes:

- Make useViewportRender option persistent during the whole runtime
(because changing of mode when we load another scene or another .blend
can disturb the pipeline) (both embedded and blenderplayer)

- Make shading type persistent during the whole runtime (both embedded and blenderplayer). I think it's better as default and if people want to change shading type they can do it with bpy during runtime.

- We call ED_area_tag_redraw for all ScrAreas when we change .blend to avoid to have not drawn (black) areas and also at game engine exit to refresh the screen. For now, I didn't notice the need to redraw areas when we change scene (only when we change .blend)

- To limit memory leaks count when we change .blend, we use BKE_screen_exit which does a good enough cleaning job. In counterpart, we have to refresh screen (ED_screen_change + ED_screen_refresh) when we launch a new .blend.

- I tried to fix profile drawing alignment but there are still cases where the alignment is not good in viewport render mode.

- I fixed an issue related to window size when we launch a new .blend and when blender window size is minimized.

I tested mainly with this:

- launcher: https://pasteall.org/blend/2eb75c8916c7484f89a22d27c3032d07
- loaded .blend: https://pasteall.org/blend/639e3f6610e34a22ab70a30fac062ce9
- scenes switch test: https://pasteall.org/blend/84ed9f02ce744409a47e2cd660d7eacd

Remaining noticed issues:

- profile alignment not always correct
- little memory usage increase when switching .blend (it was existing before, but i didn't notice). I didn't found for now how to get rid of it (maybe we have to get code from wm_window_free, maybe we can track totblocks count too when we switch scene/.blend)

I think that even it is not perfect, it can be commited, but more tests are welcome.

### DIFF
--- a/source/blender/windowmanager/intern/wm_window.c
+++ b/source/blender/windowmanager/intern/wm_window.c
@@ -2586,6 +2586,21 @@ void wm_window_ghostwindow_embedded_ensure(wmWindowManager *wm, wmWindow *win)
 {
   wm_window_clear_drawable(wm);
   wm_window_set_drawable(wm, win, true);
+
+  GHOST_RectangleHandle bounds;
+  /* store actual window size in blender window */
+  bounds = GHOST_GetClientBounds(win->ghostwin);
+  /* win32: gives undefined window size when minimized */
+  if (GHOST_GetWindowState(win->ghostwin) != GHOST_kWindowStateMinimized) {
+    win->sizex = GHOST_GetWidthRectangle(bounds);
+    win->sizey = GHOST_GetHeightRectangle(bounds);
+  }
+  GHOST_DisposeRectangle(bounds);
+  /* until screens get drawn, make it black */
+  GPU_clear_color(0.0f, 0.0f, 0.0f, 0.0f);
+
+  /* needed here, because it's used before it reads userdef */
+  WM_window_set_dpi(win);
 }
 /* End of Game engine transition */
 /** \} */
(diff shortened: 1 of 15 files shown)


youle31:
scenes switch test: https://pasteall.org/blend/84ed9f02ce744409a47e2cd660d7eacd

lordloki:
Hi,
the improvement is awesome. Great work!!
I found 2 issues:
- in Scene switch when you use blenderplayer the initial scene shows well but when you switch the scene is black (although the profile is shown)
- in loaded another blend when you start in viewport shading "renderer" and load another blend it pass to "solid". This is logic in embedded but it also occurs in blenderplayer. To test give color to the cube, load suzanne and load again the cube.

youle31:
yes, I noticed these 2 issues too. Thanks very much for review.

I think the solution is to port BackupShadingType code from KX_Scene to both gpg_ghost and bl_ketsjiembedstart and make the shading available for the whole runtime as for useViewportRender.

I'll fix that.

youle31:
So finally, I didn't fix any of the "issues" you mentionned but i think i found why it behaves like that and I think it can't be considered as real issues.

- For black shading in scene_switch, it comes from that I didn't add a background neither a light in scene.001. In workbench, it isn't noticeable because we have a default background drawn but we need to add background and light if we are in eevee.
- About colored cube when switching .blend, it happens because the start blend file is not saved. Then when you reload it, the data is loaded from the last saved .blend and not from your current session. Fortunately, undo at exit restores current session settings.

But I made the shading type available for the whole runtime according to the start blend/scene.
I think it's ok as default behaviour and if we want to have different shading types in the scenes, we can set it with bpy.

lordloki:
It has sense. Thanks for the work. I will test later!!

lordloki:
There is crash at exit when we switch scenes:
image

Other than this, it is marvellous how the blender interface is adapted to new scene/blend. just wow. +1000

lordloki:
I tested a bit more and all went perfectly. I think we commit it now. Great great work so far!!

lordloki (approved):

youle31:
Great! Thanks very much for review and tests!
We'll see with time if we manage to solve remaining issues, for now, i commit : )

PR #1369, opened 2021-01-01
UPBGE: Automatic builds

lordloki:
This pull request is not meant to be committed (yet). Only to recopile every changes i did.

Current issues:
- I think it should be work but the current lib download + compiling takes so much space on disk and I have a "Error: No space left on device". 
- Also so much time to download + compile (over the 58min). A possible solution would be to cache the libs but i didn't found the way.
- At this moment, I prentend only for windows/linux as the macos build takes a factor of 10 in spent minutes (we only have 2000min per month). For MacOS should be: around 70min * 10 = 700min per build.

### DIFF
--- a/.github/workflows/upbge_lib.yml
+++ b/.github/workflows/upbge_lib.yml
@@ -1,4 +1,4 @@
-name: Build master
+name: Download lib
 
 on:
   #push:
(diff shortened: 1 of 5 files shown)


lordloki:
Test are removed to reduce download space and compilation time

lordloki:
Closing this for the moment. I will investigate it a bit more in the future.

PR #1371, opened 2021-01-03
Support data (ID) type properties for Python components

mysticfall:
This commit adds support for Python component properties which can accept Blender data types like Mesh or Object as their values. In short, now you can do something like this:

image

In order to declare an `ID` type property, you can initialise the component `args` property with a proper type object like this:
```python
class Bootstrap(KX_PythonComponent):
    args = OrderedDict((
        ("key", "alleycat"),
        ("config", "//config.json"),
        ("Action", bpy.types.Action),
        ("Armature", bpy.types.Armature),
        ("Camera", bpy.types.Camera),
        ("Collection", bpy.types.Collection),
        ("Curve", bpy.types.Curve),
        ("Image", bpy.types.Image),
        ("Key", bpy.types.Key),
        ("Library", bpy.types.Library),
        ("Light", bpy.types.Light),
        ("Material", bpy.types.Material),
        ("Mesh", bpy.types.Mesh),
        ("Movie Clip", bpy.types.MovieClip),
        ("Node Tree", bpy.types.NodeTree),
        ("Object", bpy.types.Object),
        ("Particle", bpy.types.ParticleSettings),
        ("Sound", bpy.types.Sound),
        ("Speaker", bpy.types.Speaker),
        ("Text", bpy.types.Text),
        ("Texture", bpy.types.Texture),
        ("Vector Font", bpy.types.VectorFont),
        ("Volume", bpy.types.Volume),
        ("World", bpy.types.World),
    ))
```
This is an attempt to fill the gap between UPBGE and other popular engines like Unity or Godot which already support a similar feature.

There's an important caveat, however. If you reference non-object data types like `Action` or `Mesh`, it's strongly recommended to add a fake user to prevent unintentional deletion because clearing an once assigned property of such a type may result in wrong user count (`ID->us`).

This behaviour, however, doesn't seem to be related to my changes as you can reproduce the symptom with `ActionActuator` also. As such, I'll just ignore the problem for now and advise users to prefer `bpy.types.Object` which is safe from this problem, or add a fake user to the target data item.

Even though I've tested it quite rigorously including a scenario involving appending an object with such properties from a different file, more tests might be prudent before merging this PR.

Special thanks to @youle31 for the help! :)

### DIFF
--- a/source/blender/blenkernel/BKE_sca.h
+++ b/source/blender/blenkernel/BKE_sca.h
@@ -76,7 +76,7 @@ void sca_move_sensor(struct bSensor *sens_to_move, struct Object *ob, int move_u
 void sca_move_controller(struct bController *cont_to_move, struct Object *ob, int move_up);
 void sca_move_actuator(struct bActuator *act_to_move, struct Object *ob, int move_up);
 
-/* Callback format for performing operations on ID-pointers for sensors/controllers/actuators/components. */
+/* Callback format for performing operations on ID-pointers for sensors/controllers/actuators. */
 typedef void (*SCASensorIDFunc)(struct bSensor *sensor,
                                 struct ID **idpoin,
                                 void *userdata,
--- a/source/blender/blenkernel/BKE_sca.h
+++ b/source/blender/blenkernel/BKE_sca.h
@@ -89,17 +89,12 @@ typedef void (*SCAActuatorIDFunc)(struct bActuator *actuator,
                                   struct ID **idpoin,
                                   void *userdata,
                                   int cb_flag);
-typedef void (*SCAComponentIDFunc)(struct PythonComponent *comp,
-                                  struct ID **idpoin,
-                                  void *userdata,
-                                  int cb_flag);
 
 void BKE_sca_sensors_id_loop(struct ListBase *senslist, SCASensorIDFunc func, void *userdata);
 void BKE_sca_controllers_id_loop(struct ListBase *contlist,
                                  SCAControllerIDFunc func,
                                  void *userdata);
 void BKE_sca_actuators_id_loop(struct ListBase *atclist, SCAActuatorIDFunc func, void *userdata);
-void BKE_sca_components_id_loop(struct ListBase *complist, SCAComponentIDFunc func, void *userdata);
 
 const char *sca_state_name_get(Object *ob, short bit);
 
--- a/source/blender/blenkernel/BKE_sca.h
+++ b/source/blender/blenkernel/BKE_sca.h
@@ -76,7 +76,7 @@ void sca_move_sensor(struct bSensor *sens_to_move, struct Object *ob, int move_u
 void sca_move_controller(struct bController *cont_to_move, struct Object *ob, int move_up);
 void sca_move_actuator(struct bActuator *act_to_move, struct Object *ob, int move_up);
 
-/* Callback format for performing operations on ID-pointers for sensors/controllers/actuators. */
+/* Callback format for performing operations on ID-pointers for sensors/controllers/actuators/components. */
 typedef void (*SCASensorIDFunc)(struct bSensor *sensor,
                                 struct ID **idpoin,
                                 void *userdata,
@@ -89,12 +89,17 @@ typedef void (*SCAActuatorIDFunc)(struct bActuator *actuator,
                                   struct ID **idpoin,
                                   void *userdata,
                                   int cb_flag);
+typedef void (*SCAComponentIDFunc)(struct PythonComponent *comp,
+                                  struct ID **idpoin,
+                                  void *userdata,
+                                  int cb_flag);
 
 void BKE_sca_sensors_id_loop(struct ListBase *senslist, SCASensorIDFunc func, void *userdata);
 void BKE_sca_controllers_id_loop(struct ListBase *contlist,
                                  SCAControllerIDFunc func,
                                  void *userdata);
 void BKE_sca_actuators_id_loop(struct ListBase *atclist, SCAActuatorIDFunc func, void *userdata);
+void BKE_sca_components_id_loop(struct ListBase *complist, SCAComponentIDFunc func, void *userdata);
 
 const char *sca_state_name_get(Object *ob, short bit);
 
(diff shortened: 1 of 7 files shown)


source/blender/blenkernel/intern/python_component.c
```diff
@@ -795,6 +815,22 @@ void *BKE_python_component_argument_dict_new(PythonComponent *pc)
         PyList_SetItem(value, i, PyFloat_FromDouble(cprop->vec[i]));
       }
     }
+#define PT_DEF(name, lower, upper) \
+    else if (cprop->type == CPROP_TYPE_ ## upper && cprop->lower) { \
+        ID *id = &cprop->lower->id; \
+        if (id) { \
+            if (!id->py_instance) { \
+                id -> py_instance = pyrna_id_CreatePyObject(id); \
```
mysticfall, line 823: I'm not entirely certain that it's a safe approach. It seems to work, however, both in standalone and embedded mode.

source/blender/makesdna/DNA_python_component_types.h
```diff
@@ -39,6 +57,28 @@ typedef struct PythonComponentProperty {
...
+  struct bNodeTree *node_tree;
+  struct Object *object;
+  struct ParticleSettings *particle_settings;
+  struct bSound *sound;
+  struct Speaker *speaker;
+  struct Text *text;
+  struct Tex *texture;
+  struct VFont *vector_font;
+  struct Volume *volume;
+  struct World *world;
```
mysticfall, line 81: I tried to use a single `void *pointer` field instead but failed to make it work that way. I don't really know C/C++, so I suspect there might be a way to do it that I missed.

source/blender/blenkernel/intern/python_component.c
```diff
@@ -378,6 +379,25 @@ static void create_properties(PythonComponent *pycomp, PyObject *cls)
         }
       }
     }
+    else if (PyType_Check(pyvalue)) {
+        const char *tp_name = ((PyTypeObject *) pyvalue)->tp_name;
+
+        free = true;
+
+#define PT_DEF(name, lower, upper) \
+        if (!strcmp(tp_name, STRINGIFY(name))) { \
```
mysticfall, line 388: Not a very robust method to check types. I'd like to use `issubtype` instead, but don't know how to get a reference to each Python type instance here.

source/blender/blenkernel/intern/sca.c
```diff
@@ -1188,6 +1190,23 @@ void BKE_sca_actuators_id_loop(ListBase *actlist, SCAActuatorIDFunc func, void *
...
+{
+  PythonComponent *comp;
+
+  for (comp = complist->first; comp; comp = comp->next) {
+      ListBase *properties = &comp->properties;
+      PythonComponentProperty *prop;
+
+      for (prop = properties->first; prop; prop = prop->next) {
+#define PT_DEF(name, lower, upper) \
+          func(comp, (ID **)&prop->lower, userdata, IDWALK_CB_NOP);
```
youle31, line 1203: I'm not sure IDWALK_CB_NOP should be used in all cases

source/blender/blenkernel/intern/object.c
```diff
@@ -604,6 +613,8 @@ static void object_foreach_id(ID *id, LibraryForeachIDData *data)
   BKE_sca_controllers_id_loop(
       &object->controllers, library_foreach_controllersObjectLooper, data);
   BKE_sca_actuators_id_loop(&object->actuators, library_foreach_actuatorsObjectLooper, data);
+  BKE_sca_components_id_loop(&object->components, library_foreach_componentsObjectLooper, data);
```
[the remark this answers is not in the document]
mysticfall: I always wondered what "sca" means 😅

source/blender/blenkernel/BKE_sca.h
```diff
@@ -89,12 +89,17 @@ typedef void (*SCAActuatorIDFunc)(struct bActuator *actuator,
                                   struct ID **idpoin,
                                   void *userdata,
                                   int cb_flag);
+typedef void (*SCAComponentIDFunc)(struct PythonComponent *comp,
```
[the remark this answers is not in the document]
mysticfall: Fixed it.

source/blender/blenkernel/intern/object.c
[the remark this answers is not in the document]
mysticfall: Fixed it.

source/blender/blenkernel/intern/object.c
```diff
@@ -1481,6 +1492,15 @@ static void object_blend_read_lib(BlendLibReader *reader, ID *id)
     }
   }
 
+  for (PythonComponent *comp = ob->components.first; comp; comp = comp->next) {
```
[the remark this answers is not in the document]
mysticfall: Fixed it.

source/blender/blenkernel/intern/sca.c
mysticfall: Changed it along with a few similar parts regarding actuators. This should fix the erraneous behaviour which was mentioned above. Thanks for pointing this out!

source/blender/blenkernel/intern/sca.c
```diff
@@ -1188,6 +1191,23 @@ void BKE_sca_actuators_id_loop(ListBase *actlist, SCAActuatorIDFunc func, void *
   }
 }
 
+void BKE_python_components_id_loop(ListBase *complist, BKEPyComponentIDFunc func, void *userdata)
+{
+  PythonComponent *comp;
```
lordloki, line 1196: This function and related included should go in source/blender/blenkernel/intern/python_component.c

lordloki:
Great work!!!. I need a little free time to do test and proper review. Maybe this night i will have some free time

source/blender/blenkernel/intern/sca.c
mysticfall: Fixed it. Thanks!

youle31:
Awesome work! Very impressive! But too technical for me to have a "sure" opinion about it. However, I'm ok to commit and see later if there are issues with. Thanks very much for your work!
My test file: https://pasteall.org/blend/86c3559f71854b08b80360698c009d9d (light component)

youle31 (approved):

lordloki (approved):
It seems ok to me. I tested and it worked great. Great work!!!!

Regarding ID changes i did several test too and they work fine. I also found that for speaker->sound that it is a similar case than ours they used IDWALK_CB_USER too.

After the commit I will recover the uncommented id_us_minus/plus as i checked too and i didn't see any fail

BluePrintRandom:
🍾 👯‍♀️

PR #1374, opened 2021-01-06
Ensure Python components of a game object started before those on its ancestors

mysticfall:
### Summary

This PR intends to change the invocation order of `KX_PythonComponent` so that a component on a game object would get invoked before those attached to its ancestors.

### Rationale

Consider a case where a Python component `P` requires other components `C1` and `C2` as its dependency. We can imagine they are attached in a game object tree which looks like below:
```
P -*- C1
     |
     *- C2
```
As a component is often initialised from arguments given in its `start(args)` callback, `C1` and `C2` may still remain in an invalid state when `P`'s `start` method gets called, because UPBGE currently does not guaranteed the invocation order in such cases.

Apparently, the engine invokes game objects in an alphabetical order of their names regardless of their respective depth which is not a very useful behaviour, IMO.

In comparison, Godot ensures that `OnReady()` callback on child nodes get invoked before that of the parent which can be useful to avoid the problem described in the above example.

This change intends to introduce a similar behaviour to UPBGE, although it still doesn't guarantee the invocation order between siblings, which we might consider implementing later.

### DIFF
--- a/source/gameengine/Ketsji/KX_PythonComponentManager.cpp
+++ b/source/gameengine/Ketsji/KX_PythonComponentManager.cpp
@@ -2,6 +2,11 @@
 #include "KX_PythonComponent.h"
 #include "KX_GameObject.h"
 
+static bool compareObjectDepth(KX_GameObject *o1, KX_GameObject *o2)
+{
+  return o1->GetSGNode()->GetDepth() > o2->GetSGNode()->GetDepth();
+}
+
 KX_PythonComponentManager::KX_PythonComponentManager()
 {
 }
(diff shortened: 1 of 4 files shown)


youle31 (approved):

youle31:
As discussed on discord, I think even if execution order of objects in python components manager is modified and that it could "potentially" cause backward compatibility with some existing files, there are few chances that it happens.

Personally, the execution order of python components doesn't really matter for me because 1. I almost never use python components 2. There are few chances that I write a complex coding architectures where python components order execution matter.

But as mysticfall says that in other engines, the norm is to execute children before parents, I'm ok with this commit. Thanks!

lordloki (approved):
Code seems ok to me and the reason for changes makes sense. Great work.

mysticfall:
Thanks much for the swift review, both of you! :)

mysticfall:
Merging it now, as youle told me he's busy playing AOE3 at the moment :P

Issue #1377, opened 2021-01-07
0.3.0 parent losing/strange beahviour on scene restart


ExcaliburGO:
I noticed that in UPBGE 3.0, something strange is happening if you restart the scene and some objects have already moved. Especially the parented ones are affected, they get some strange rotation and location. Adding a parent actuator on start sometimes solves it, but not always. Here is the demo file to show you what's happening:
[bug.zip](https://github.com/UPBGE/upbge/files/5780270/bug.zip)
There is an object with 2 children, controlled by mouse and keyboard (WASD), you can restart the scene by pressing X.

youle31:
Fixed in  69d4c24 . Thanks for report

youle31:
My commit doesn't handle the case where original kxobjects are removed during runtime then I reopen the issue.

test file: 
[bug.zip](https://github.com/UPBGE/upbge/files/5784198/bug.zip)

ExcaliburGO:
I just tested the restore_obmat branch and it is solved, thanks, I will be using it until it's done properly and merged into master.

youle31:
Merged, Thanks again for report, i hope this will fix all potential parent issues at scene exit...

PR #1378, opened 2021-01-08
Try to fix objects obmat at kxscene exit

youle31:
I had to move some code from KX_GameObject to KX_Scene.

It seems to work but I dooubt it is conventional...

I retried undo but without success

It is to try to fix https://github.com/UPBGE/upbge/issues/1377

Maybe there is a simpler way...


lordloki (approved):
Code seems ok to me. I tested with several blends and they worked well. Good job

youle31:
Thaks very much! I thought about 2 little potential improvements during sleeping. I'll test, commit, then merge.

PR #1382, opened 2021-01-15
UPBGE: Various updates for gpg_ghost (from wm_init_exit)

youle31:
Pick few things in wm_init_exit/creator and remove bge specific code (ED_screen_refresh_blenderplayer).

EDIT: It is crashing when using "viewport render", i'll check


youle31:
No, actually, i already tried something similar. We have to keep ED_screen_refresh_blenderplayer... Other pieces of code could be taken from this branch to update gpg_ghost if we have bugs. I close that, sorry

PR #1384, opened 2021-01-23
UPBGE: SMAA Quality presets and include predication

lordloki:
After several tests we found these improvements to SMAA:
- Quality presets: Low, Medium, High and Ultra. Default High.
- Internal use of predication for a better edge detection.

### DIFF
--- a/release/scripts/startup/bl_ui/properties_render.py
+++ b/release/scripts/startup/bl_ui/properties_render.py
@@ -467,19 +467,6 @@ class RENDER_PT_eevee_sampling(RenderButtonsPanel, Panel):
         return (context.engine in cls.COMPAT_ENGINES)
 
     def draw(self, context):
-        pass
-
-class RENDER_PT_eevee_sampling_taa(RenderButtonsPanel, Panel):
-    bl_label = "TAA"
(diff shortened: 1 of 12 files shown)


youle31:
Hi, thanks very much for your work and tests about smaa!

I approve.

Optionally, i propose this patch:

[p.zip](https://github.com/UPBGE/upbge/files/5861010/p.zip)

I'm not sure the line in scene.c is needed if we add the smaa_quality in DNA_scene_defaults

But what i noticed is that adding smaa_quality to DNA_scene_defaults allos to use this:

Sans titre

Previously, these settings were in rna_... with a line like RNA_def_property_float_default but I think now they are using DNA_scene_defaults instead, and isk exactly what else we can remove when we add something in DNA_scene_default (like: should we remove the line in scene.c...?)

Else, in the current patch, as I didn't remove the line in scene.c, i moved it just to put it with other upbge specific options

youle31 (approved):

lordloki:
Ah, thanks!!
I though that the DNA_***_defaults were a refactor to replace BKE_**** initializations. But it has the reset to default feature. I hadn't seen until now.

Thanks for reviewing and testing.

PR #1385, opened 2021-01-24
UPBGE: Rework part of game default settings

youle31:
1. Tried to reorganize the versioning code

2. Tried to change new Object and new Scene initialisation method (It is using DNA_scene_defaults and DNA_object_defaults.
I think like that we can remove some code in int_scene_data and
init_object_data)

EDIT 1: 
- startup.blend
I was wrong in the summary. Now, the only versioning file we have to keep maintained is versioning_upbge, the others don't have to be changed.

EDIT2:
- new scenes added, new objects added...
 Also we normally need to keep maintained DNA_thing_default to handle data initialisation (init_object_data, init_scene_data...).


lordloki (approved):
I tested with old & new files and they load saved values correctly.
Code seems me ok.
After you commit this pull I will remove the code under #if 0 in versioning_upbge

Good work!!!

youle31:
Thanks very much! I merge.

PR #1389, opened 2021-01-28
UPBGE: Use userprefs to load addons in blenderplayer

youle31:
Addons need to read userpref.blend file to know which addons are
registered.

It allows to use registered addons in blendeprlayer.

If no userpref.blend is found, default user preferences are
loaded.

Code picked from wm_files > wm_homefile_read

### DIFF
--- a/source/gameengine/GamePlayer/GPG_ghost.cpp
+++ b/source/gameengine/GamePlayer/GPG_ghost.cpp
@@ -941,7 +941,6 @@ int main(int argc,
   /* Try to load existing user preferences from config folder:
    * Either from exported Game folder, either from blender userprefs file
    */
-  char filepath_startup[FILE_MAX] = "";
   char filepath_userdef[FILE_MAX] = "";
   UserDef *userdef = nullptr;
 
--- a/source/gameengine/GamePlayer/GPG_ghost.cpp
+++ b/source/gameengine/GamePlayer/GPG_ghost.cpp
@@ -938,8 +938,30 @@ int main(int argc,
 #endif
   UI_theme_init_default();
 
-  UserDef *user_def = BKE_blendfile_userdef_from_defaults();
-  BKE_blender_userdef_data_set_and_free(user_def);
+  /* Try to load existing user preferences from config folder:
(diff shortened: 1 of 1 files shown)


source/gameengine/GamePlayer/GPG_ghost.cpp
```diff
@@ -938,8 +938,30 @@ int main(int argc,
...
+   * Either from exported Game folder, either from blender userprefs file
+   */
+  char filepath_startup[FILE_MAX] = "";
+  char filepath_userdef[FILE_MAX] = "";
+  UserDef *userdef = nullptr;
+
+  const char *const cfgdir = BKE_appdir_folder_id(BLENDER_USER_CONFIG, NULL);
+  if (cfgdir) {
+    BLI_path_join(filepath_startup, sizeof(filepath_startup), cfgdir, BLENDER_STARTUP_FILE, NULL);
+    BLI_path_join(filepath_userdef, sizeof(filepath_startup), cfgdir, BLENDER_USERPREF_FILE, NULL);
```
lordloki, line 951: filepath_startup is really used?

Additionally, I think there is typo as it should be:
BLI_path_join(filepath_userdef, sizeof(**filepath_userdef**), cfgdir, BLENDER_USERPREF_FILE, NULL);

lordloki (changes requested):
Other that commented code seems me ok. I will test when i modify the addon

youle31:
Yes sorry it was a copy/paste from wm_files without changes, i did the requested changes, Thanks for the review!

lordloki (approved):
Great!! I just pushed the save_as_runtime addon modification to copy userpref.blend

mysticfall:
It works perfectly for me. Thanks much! :)

mysticfall (approved):

youle31:
Perfect! Thanks very much for review!

I'm not 100% sure that the priority goes to exported game folder but my test protocol was messy, then if there's an issue with priorities, we'll fix it later.

Thanks both for review and tests and for the addon! I merge that

PR #1393, opened 2021-02-03
UPBGE: Try to fix dupligroups rigid boby constraints replication

youle31:
I'm not sure about what i'm doing.

Blue Print and Wuiaeyo tested with their files and it seemed ok with this second version of the patch.

Sorry to ask you again @lordloki , when you have time (no hurry at all), to test this version with your files (and review if you want).

Thanks to all for the tests!


lordloki (approved):
I checked against all these files https://github.com/UPBGE/UPBGE-Testfiles/tree/master/Physics/Constraints and they worked well. Good work!!!

youle31:
Thanks very much for tests! And thanks for your works for upbge test files github too and for your works on documentations etc....!

PR #1394, opened 2021-02-04
UPBGE: Update gpg_ghost code

youle31:
- Few more comments
- kindof markers to know whenever it is creator.c code or wm_init_exit.c code
- Few additions (to follow almost exactly blender launching process)


lordloki:
I tested with several complex files and the player worked well. I don't know if you need that i test something special...

youle31:
Thanks very much for tests! No, actually I also tested it several files and it seemed ok.
I'll test again a bit then merge if I don't notice issues.

I wanted to update the code to have something with "markers" to follow the code in creator.c and wm_init_exit.c, for maintenance.

PR #1397, opened 2021-02-05
UPBGE: Recover panzergame Time refactor (without std::sleep_for)

youle31:
Several users said that they prefered time management as in 0.2.4.

Then I recovered panzergame Time management refactor:

Mainly this:

e9b07807f99d97dde7162b3fbb7439ea70fea0b5

+ that:

dfb83b57e0ca5f4451122e3eadc732f9a784df37

### DIFF
--- a/source/gameengine/Common/CM_Clock.cpp
+++ b/source/gameengine/Common/CM_Clock.cpp
@@ -2,21 +2,21 @@
 
 CM_Clock::CM_Clock()
 {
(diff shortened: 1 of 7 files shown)


lordloki (approved):
I tested and work OK. I also tested with metronomo file and it fixes the logic when the FPS requested is over the real fps. Great!!

youle31:

PR #1405, opened 2021-02-15
UPBGE: Fix MacOS hidpi retina display issues

lordloki:
Basically, take into account the nativepixel size for window creation and mouse adquisition.

Due to we don't have any hardware to test directly I commit this but it can be modified in the future if any issue raises.

Thanks to @youle31 for code and review, and to @ExcaliburGO for testing


youle31 (approved):

youle31:
It sounds good to me. Thanks very much for this hard work! This was complex without being able to reproduce the bug.

PR #1413, opened 2021-02-22
UPBGE: Try to get rid of resetTaaSamples API

youle31:
I noticed a slight difference between viewport drawing and custom bge
loop.

Investigating on it, i realized that the difference comes from
vedata->stl->g_data->view_update was always false contrarly to viewport.

It is because in viewport, the depsgraph detects all changes done in the
scene each time there is a notifier and notifies the drw_engines.
This behaviour comes from a callback set in wm_init_exit:

DEG_editors_set_update_cb(ED_render_id_flush_update,
ED_render_scene_update);

At a lower level, the drw_engine is notified by:

DRW_notify_view_update(const DRWUpdateContext *update_ctx)

When we are using viewport render, ARegion viewports are used, then this
function works correctly. But as we are using our own GPUViewports in
custom bge loop, this function is not working correctly.

This commit is a first attempt to have this function working even if we
are using our own custom GPUViewports.

If we have it working, we can normally get rid of resetTaaSamples API as
the depsgraph automatically triggers the needed functions to set the
taa_current_sample to 1.

When we do this, the TAA flags are the same than in viewport render mode at 1 sample (contrarly to previously):

- Master:
master

- This branch:
this_branch

TODOS:
- Check ImageRender (no notifier is called here) (Simple test done)
- Check Overlay scenes                                      (Simple test done)
- check multiple viewports                                 (Simple test done)

Additional notes:

- Original objects must be tagged for update before BKE_scene_graph_update_tagged if we want that the depsgraph notifies the drw_engine that an object/material... moved/changed. Then BKE_depsgraph_update_tagged must be called (1 time before each render loop).

The problem is that when we do that, the depsgraph overrides game logic transform priority. For example, if an object has a viewport lock constraint it will be taken into account in object transformation.

As we want the logic transform to keep priority we need to transform again the evaluated objects (the ones which will be rendered) assigning them the bge SceneGraph transform result.

So I had to split tag for update in 2 parts: 1 to notify the drw engine that a change happened (before BKE_depsgraph_update_tagged), 1 to assign the good SceneGraph matrix to the object (after BKE_scene_graph_update_tagged).

- For ImageRender, I had to (add a random notifier + call BKE_depsgraph_update_tagged) to tell the depsgraph to notify the drw_engine. BKE_depsgraph update_tagged must be called each time we do a render loop.

- For VideoFFMPEG, I had to add a depsgraph notifier to tell the depsgraph that something changed. As VideoFFMPEG doesn't require a render loop, BKE_depsgraph_update_tagged was not needed (the VideoFFMPEG notifier will be flushed the next time BKE_depsgraph update_tagged will be called).

- For custom viewports, I had to do like for overlay scenes -> We wait for the last render pass to update m_prevObmat because the objects don't move during all render passes (main + custom viewports + overlay) and at each render loop, we need to tell the depsgraph that object have been moving at previous logic step.


lordloki:
I tested with several files and they worked very well. The code seems me ok. Great work

lordloki (approved):

youle31:
I hope there won't be unexpected bugs, but I tested with many files and it seemed ok.

I merge this

PR #1419, opened 2021-02-28
UPBGE: Track python controller scripts

youle31:
Used to avoid to free scripts when we shoould not.
Used also to append script at the same time than object...

Should work both in script mode and module mode.

### DIFF
--- a/source/blender/blenkernel/intern/sca.c
+++ b/source/blender/blenkernel/intern/sca.c
@@ -1072,11 +1072,14 @@ void BKE_sca_controllers_id_loop(ListBase *contlist, SCAControllerIDFunc func, v
     switch (controller->type) {
       case CONT_PYTHON: {
         bPythonCont *pc = controller->data;
-        char *modulename = strdup(pc->module);
-        char *ext = strdup(".py");
-        char *dest = strdup("");
-        if (strlen(modulename)) {
+        if (strlen(pc->module)) {
           if (!pc->module_script) {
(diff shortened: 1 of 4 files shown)


lordloki:
I did several test and works ok with cleaning datablocks.
I found a situation that lead to a crash: when you select python module and write the python file + function name (i.e test.function_name) but you don't have any file loaded or created in the text editor.

youle31:
Ah, great, thanks very much for tests! I'll fix that today or tomorrow. Thanks!

source/blender/blenkernel/intern/sca.c
```diff
@@ -1070,7 +1072,35 @@ void BKE_sca_controllers_id_loop(ListBase *contlist, SCAControllerIDFunc func, v
     switch (controller->type) {
       case CONT_PYTHON: {
         bPythonCont *pc = controller->data;
-        func(controller, (ID **)&pc->text, userdata, IDWALK_CB_NOP);
+        if (strlen(pc->module)) {
+          if (!pc->module_script) {
+            char modulename[FILE_MAX];
+            strcpy(modulename, pc->module);
```
lordloki, line 1078: Here i would use `BLI_strncpy(modulename, pc->module, sizeof(modulename));` instead of `strcpy(modulename, pc->module);` to avoid a possible buffer overflow if the name is very long.

lordloki (approved):
It works well. Great work.

youle31:
Thanks very much!

PR #1422, opened 2021-03-05
UPBGE: Enable support of SoftBody volumes (when we uncheck Shape Match)

youle31:
I don't know what it means but it is just to not have the softbody
inactive when we uncheck CCD_BSB_SHAPE_MATCHING (Shape Match -
use_shape_match) option.

EDIT: In 2.79 official, the simulation was running when we unchecked Shape Match, not in upbge 0.2.5


lordloki:
Yeah, I never used this option before but the code follows 2.79 behaviour now. Great!!

lordloki (approved):

youle31:
Thanks! I merge

PR #1426, opened 2021-03-06
UPBGE: ImageRender pre/post draw callbacks

youle31:
With recent custom bge loop refactor, it was not possible anymore to set evaluated scene render options before ImageRender pass.

It can be interesting to set these options:

- It is better to use evaluated scene before when we set the render options because the depsgraph is not notified to do an extra draw pass for no reason.
- We might want to set background transparent or disable post processing effects before ImageRender pass.

To make this possible again, pre_draw callbacks are implemented and inserted after BKE_scene_graph_update_tagged and before the image render bge draw loop.

I have no usecase for post_draw callbacks yet but it costs nothing to add it.

Test file: https://ufile.io/kx28163c

Python API:

Side note:

ImageRender often requires to hide some objects before the ImageRender pass. I think that for performances, it would be better to use ob_eval.hide_viewport = True than to use kxob.visible = False because KX_GameObject::SetVisible is using depsgraph which can be a bit slow. Furthermore, there if ob_eval visibility state is changed, it is just during ImageRender pass, there should be no need to restore visibility after bge.tex.refresh() contrarly to if we use kxob.visible

### DIFF
--- a/doc/python_api/rst/bge.texture.rst
+++ b/doc/python_api/rst/bge.texture.rst
@@ -779,6 +779,65 @@ Image classes
 
       :type: bool
 
+   .. attribute:: pre_draw
+
+      A list of callables to be run before the render step.
+      These callbacks can be used to make background transparent
+      or disable post processing effects of evaluated scene.
+      Evaluated scene has to be used for performance reasons.
+
+      :type: list
+
+   .. code-block:: python
+
+      import bpy, bge
+      from bge import texture
+
+      cont = bge.logic.getCurrentController() # on main camera
+      scene = bge.logic.getCurrentScene()
+      rendercam = scene.objects["rendercam"]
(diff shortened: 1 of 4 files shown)


lordloki:
I never used this feature before but code seems OK to me. I tested and it worked well. Great work!!!

lordloki (approved):

youle31:
yeah, it's because some ImageRender APIs can't work anymore (background, zenith_color, alpha....). I thought it would be more flexible to set these settings before ImageRender pass in a callback (disable bloom, AO, volumetrics, make the background transparent, use different color manament options...). If you use not evaluated scene to do that it is very slow because there are tons of operations trigger in depsgraph, whereas if you do this in scene_eval just before render, it costs nothing.

Thanks very much for review and test!

PR #1429, opened 2021-03-10
UPBGE: Add the possibility to convert actions at runtime

youle31:
This patch does 2 things:

- When we append a cube which has an action actuator, then convert the cube with convertBlenderObject, the action is registered in logic manager and can be played without doing anything on user side.

- It adds an API to register a bpy.types.Action in the logic manager. Like that we can append only an action and play it.
A counterpart is added to unregister the Action from the logic manager. It doesn't free the Action itself but you need to unregister the action from the logic manager before freeing it with bpy.data.actions.remove(Action, do_unlink=True).
parenthesis: convertBlenderObject and derived methods counterpart is endObject. libLoad had a libFree counterpart which was unregistring loaded actions. So I added this counterpart for Actions only.

test files:

1) launcher to append a cube with an action actuator:
https://ufile.io/e1vpfbnh
2) launcher to append only a bpy.types.Action:
https://ufile.io/6338s79x (EDIT: wrong link)
https://ufile.io/j8iivkbe (this one)
3) Library (to append either the cube, either the action only):
https://ufile.io/esuqx6ud


lordloki (approved):
I tested and it worked like a charm. Code seems ok to me. Great work!!!

youle31:
Thanks very much @lordloki for review and tests! I merge.

Issue #1437, opened 2021-03-16
upbge 0.3 (program crashes and then quits when running game with navigation mesh


dennispetsinis:
System Information
Operating system: windows 10 (version 19042
Graphics card: Amd Ryzen 3200U

Blender Version 2.92 (originally tried as well 2.93)
Can not be sure it has to with blender, adding logic in upbge 0.3
Attached file.
I get the same results on my desktop as well.

I started upbge.

scene name: Main

added a cube as my landscape. (added material)
selected: added build navigation mesh

added a cone as enemy. (added material - sea blue)
Selected: added logic (Always - And - Steering (path following) - target (player) - navigation (navmesh)

added cube for player. (edited the cube, added material)
Selected: added logic (Collision / material- sea blue) - And - (Edit object / end object)
                                                                                                 (scene / set scene / main)
Press "P"

targeting and collision occur.

there is no scene change, program crashes and terminates.

[firing cube_1.zip](https://github.com/UPBGE/upbge/files/6152079/firing.cube_1.zip)

youle31:
Thanks for report! It should be fixed in last master

PR #1438, opened 2021-03-17
UPBGE: Try to fix a crash with navmesh and memory leaks

youle31:
It is to try to fix:
https://github.com/UPBGE/upbge/issues/1437

I see no more memory leaks reported in blenderplayer


source/gameengine/Ketsji/KX_NavMeshObject.cpp
```diff
@@ -810,6 +817,19 @@ bool KX_NavMeshObject::BuildNavMesh()
     if (vertices) {
       delete[] vertices;
     }
+    if (dvertices) {
```
lordloki, line 820: There is another delete [ ] dvertices more down. The same occurs in the other checks.

lordloki (approved):
It worked well. Great fix!!

source/gameengine/Ketsji/KX_NavMeshObject.cpp
youle31: Thanks! I'll fix that.

youle31:

PR #1444, opened 2021-03-24
UPBGE: Try to recover 0.2.5 behaviour for compound things

youle31:
- It recovers something in BlenderDataConversion to take into account compound children. My guess is that as I didn't know what was compound, I disabled it during 0.3 development.
- It recovers a commit just after 0.2.2: 067da32bf00a0917f2deceb7dbe8e88ef6f6f9a1 (I saw you already had recovered another previous commit related to compound just before 0.2.2)

wuaeio told me that it was not working then I did that. He tested and told me that it was working as excpected according to him.

I saw you already pushed a commit related to compound a long time ago in official bge, then maybe you know what it is and can evaluate better than me if it sounds ok to you.

parenthesis: I left dm argument in physenv->ConvertObject(...., DerivedMesh *dm, ....) but it is always nullpr.

### DIFF
--- a/source/gameengine/Converter/BL_BlenderDataConversion.cpp
+++ b/source/gameengine/Converter/BL_BlenderDataConversion.cpp
@@ -544,7 +544,7 @@ RAS_MeshObject *BL_ConvertMesh(Mesh *mesh,
 
 //////////////////////////////////////////////////////
 static void BL_CreatePhysicsObjectNew(KX_GameObject *gameobj,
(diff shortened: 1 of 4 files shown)


lordloki:
I tested the test files and they worked fine. The code seems me ok to me too. Thanks for the work!!

lordloki (approved):

youle31:
Thanks very much for review and tests! I merge

Issue #1445, opened 2021-03-25
Animation of custom Property not executable in logic brick editor


BerglerAndi:
Hey there,
 i have a problem with playing animations of custom properties in the logicbrick actuator. I use the custom properties to control materials. The first object with i made works. But all further objects with animated custom properies i made, don't work. In my example the light with red an green work perfect. But the white light doesn't do.

btw: if i copy the first object, it will work..

![grafik](https://user-images.githubusercontent.com/58920393/112540531-95b08200-8db2-11eb-82e5-d7d0110578ea.png)
[CustomProp_Animation.zip](https://github.com/UPBGE/upbge/files/6207735/CustomProp_Animation.zip)

BluePrintRandom:
personally I would use object color and py or actions with the object color

this is already allocated per object I believe.

youle31:
Actually, i tag it as a bug because it shouldn't be working at all. Idk why for now, the modifier on the cylinder makes it playable.
The BL_Action code needs to be changed i guess...

About animating materials with drivers, it is more a feature request than a bug then a discussion could be opened but wouldn't treated with same priority than the bug.

PR #1446, opened 2021-03-26
UPBGE: BL_Action: Add checks to try to ensure we're playing the right action

youle31:
it is an attempt to fix https://github.com/UPBGE/upbge/issues/1445

There are 2 things:

- We compare directly action pointers instead of action names when we can.
- When it is not possible, for currently playable action "data types", we ensure that m_action (actuator's action) FCurve rna path contains the "data name" (constraint name, modifier name...) (if not the action is not played). The code is derived from BKE_modifier_uses_time and adapted.

BL_Action code can potentially be improved to support more action types but needs more investigations and can be done later.

### DIFF
--- a/source/gameengine/Ketsji/BL_Action.cpp
+++ b/source/gameengine/Ketsji/BL_Action.cpp
@@ -428,8 +428,8 @@ void BL_Action::Update(float curtime, bool applyToObject)
          md = (ModifierData *)md->next) {
       bool isRightAction = false;
       /* action - check for F-Curves with paths containing 'modifiers[' */
-      if (ob->adt && ob->adt->action && ob->adt->action->curves.first) {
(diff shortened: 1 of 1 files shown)


lordloki:
For me it is not working after last commit. If I revert the last commit it works. The code seems OK. I checked the idprop action branch and it worked well too. Great!!

youle31:
Ah, ok, lol, I'll check and test again, and commit without last commit. Thanks very much for review and tests!

youle31:
Ok, Thanks very much for the checks you did. Indeed, the last commit seems to cause an issue. I don't understand why, but it is working as you said. I tried to improve the names checking, left debug infos commented just in case, reverted previous commit. I merge like that and if you see improvements or find a mistake in the code I added, we can fix later. Thanks very much for your eagle eye review!

Issue #1481, opened 2021-05-16
UPBGE 0.3 Object Text don't work in overlay collection mode.


fumangy:
I have tested last version and I have this issue in console , when I make  an accountant with object Text  (Error: failed convert font property "Text").
Attach video and blend file. 
Thanks ,  best regards.
Toni.
UPBGE   2021-05-15 
[Contador.blend.zip](https://github.com/UPBGE/upbge/files/6490224/Contador.blend.zip)
 

https://user-images.githubusercontent.com/29063438/118415658-f0c46e00-b6ab-11eb-8ca9-c05fe33c31df.mov

fumangy:
It seems that for it to work, there has to be some moving object, rotation, or having some kind of physics, rigid body, dynamic, character and this moving.
If the scene is completely static it does not work.
This is correct.
Thank you.
[Contador_1.mov.zip](https://github.com/UPBGE/upbge/files/6493241/Contador_1.mov.zip)
[Contador_1.blend.zip](https://github.com/UPBGE/upbge/files/6493243/Contador_1.blend.zip)

lordloki:
It seems that it is necessary to add a notify for viewport update in overlay

youle31:
Yes, it's because when the logic is updated, a notifier is added but this notifier is flushed during first render pass, and not anymore in overlay pass... This bug can concern all places where notifiers are added during logic pass (BL_Action...). Objects notified/tagged 1 time during logic could be flushed again, or maybe only at last render pass.

lordloki:
The last build hasn't got the issue. Thanks to youle for the fix. Closing the report. Thanks for the feedback.

fumangy:
Thanks devs for made this amazing engine in to blender.

PR #1482, opened 2021-05-18
UPBGE: Fix most of missing depsgraph notifiers for extra render passes

youle31:
There are remaining cases not handled (ImageRender + few other cases)... It would need to find a system for all occurrences of DEG_id_tag_update in ge code.

But I think it is fixing most of common cases. Maybe there is a simpler/better way than this patch to proceed.

It is to fix this: https://github.com/UPBGE/upbge/issues/1481 and other potential issues with extra render passes...

### DIFF
--- a/source/gameengine/Ketsji/BL_Action.cpp
+++ b/source/gameengine/Ketsji/BL_Action.cpp
@@ -34,8 +34,6 @@
 #include "BKE_object.h"
 #include "BLI_listbase.h"
 #include "BLI_string.h"
-#include "DEG_depsgraph_query.h"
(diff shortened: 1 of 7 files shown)


youle31:
I'm not sure the code is correct for now about finding an item in std::vector<std::pair<Object *, IDRecalcFlag>> vec... I need to test

lordloki (approved):
The code seems me ok to me. Also, I tested and it worked ok. Good work!!!

youle31:
Thanks very much! I merge (There can still be cases where we can have not notified things for a drawing pass but we can see later for this)

PR #1484, opened 2021-05-19
UPBGE: improve text properties conversion

lordloki:
Instead of passing an empty string for String property creation we pass the stream data.

Additionally, we remove the stream fail check as the conversion is well done but we mantain the bad check to check the stream integrity. This way we avoid the console error message that it is not very useful for the user (as the conversion is really done).

Previously, the console error message was visible when we use, for example, a font with the text "hello" but we set the string property as integer.


youle31:
I had a quick look at stackoverflow for .fail() and understood why an error was spawned. Also tested with Contador.blend and works fine. Thanks very much!

youle31 (approved):

PR #1488, opened 2021-05-22
Make PythonComponent's properties overridable

mysticfall:
This will allow overriding the properties of a `KX_PythonComponent` which is attached to an object linked as a library.

However, it doesn't make components themselves overridable, or other properties of `GameObjectSettings` like sensors or actuators for that matter. I tried to make it work but haven't been successful so far.

If anyone has an idea what could be missing to make it work, please let me know. Thanks!

### DIFF
--- a/source/blender/blenkernel/BKE_python_component.h
+++ b/source/blender/blenkernel/BKE_python_component.h
@@ -40,6 +40,9 @@ struct PythonComponent *BKE_python_component_create_file(char *import,
 void BKE_python_component_reload(struct PythonComponent *pc,
                                  struct ReportList *reports,
                                  struct bContext *context);
+
+struct PythonComponent *BKE_python_component_copy(PythonComponent *comp);
+
 void BKE_python_component_copy_list(struct ListBase *lbn, const struct ListBase *lbo);
 void BKE_python_component_free(struct PythonComponent *pc);
 void BKE_python_component_free_list(struct ListBase *base);
--- a/source/blender/blenkernel/intern/python_component.c
+++ b/source/blender/blenkernel/intern/python_component.c
@@ -726,7 +726,7 @@ void BKE_python_component_reload(PythonComponent *pc, ReportList *reports, bCont
   load_component(pc, reports, CTX_data_main(context));
 }
 
-static PythonComponent *copy_component(PythonComponent *comp)
+PythonComponent *BKE_python_component_copy(PythonComponent *comp)
 {
   PythonComponent *compn;
   PythonComponentProperty *cprop, *cpropn;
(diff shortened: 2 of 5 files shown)


lordloki:
Do you have a little blend to test it?
Maybe using PROPOVERRIDE_OVERRIDABLE_LIBRARY instead PROPOVERRIDE_LIBRARY_INSERTION?

youle31:
Hello, I quickly tested it with these files:

[main.zip](https://github.com/UPBGE/upbge/files/6527368/main.zip)
[linked.zip](https://github.com/UPBGE/upbge/files/6527369/linked.zip)

I never used library override previously. I just saw that to have it working, one the object is linked we need to call "Make library override" operator in search bar (After that we can write Move speed and Turn speed on the overridable library object in the test file)

It's working for me for component properties. For now, I only tested with this.

mysticfall:
@lordloki // I believe `RNA_define_lib_overridable(true)` is equivalent of declaring every property as `PROPOVERRIDE_OVERRIDABLE_LIBRARY`, while `PROPOVERRIDE_LIBRARY_INSERTION` could be additionally needed for `PROP_COLLECTION` type properties. At least, that is how I understand it :)

@youle // Thanks for testing and the sample files!

I just updated the PR with additional lines to make Actuator/Sensor/GameObject properties overridable as well.

mysticfall:
I had to revert the commit about SCA parts since youle found issues after the change (crashing on exit and unintended links created between logic bricks). Attached a patch that shows the reverted change, in case it may be needed in future:

[patch.zip](https://github.com/UPBGE/upbge/files/6527860/patch.zip)

youle31 (approved):

youle31:
I'm ok, I just said that maybe changes related to sensors/actuators.... in rna_object where not needed anymore

youle31 (changes requested):
I didn't realized that component properties had an issue in current patch state (as discussed on discord, as far i understood, if an override object component is removed, the changes are not saved, then after looking at other PROPEVERRIDE_LIBRARY_INSERTION occurrences, it's maybe needed to use rna_def_property_override_funcs to have it working maybe)

mysticfall:
> I didn't realized that component properties had an issue in current patch state (as discussed on discord, as far i understood, if an override object component is removed, the changes are not saved, then after looking at other PROPEVERRIDE_LIBRARY_INSERTION occurrences, it's maybe needed to use rna_def_property_override_funcs to have it working maybe)

Added `rna_def_property_override_funcs` as suggested by youle which apparently fixed the problem with added components not being persisted when reload the .blend file.

It appears the same code can be applied to SCA/P parts as well without too much difficulty. However, I'm not entirely sure if the polymorphic nature of SCA bricks, or the way they reference each other with links would work.

As such, I just removed everything except for the PythonComponent related changes for now.

youle31 (approved):

youle31:
I tested and it seems to work fine for me as far I can tell. It seems to have the same behaviour than modifiers in these test files:

[main.zip](https://github.com/UPBGE/upbge/files/6565472/main.zip)
[linked.zip](https://github.com/UPBGE/upbge/files/6565473/linked.zip)

lordloki (approved):
I don't have any experience in library overriding but i testedthe changes and it worked. Code seems to follow the same convention than other overrriding code.

Great work!!!!

mysticfall:
Thanks lordloki and youle :)

PR #1494, opened 2021-05-28
UPBGE: Add an option for AddObject to duplicate object data

youle31:
This adds code to duplicate completely a gameobject with its data:

- First the Object is duplicated using USER_DUP_OBDATA flag which enables all duplicate options:

Capturei

- For an OB_MESH: materials, particle systems and mesh will be duplicated:

Capture

- Then the duplicated object is converted as single object in BL_ConvertObjects pipeline.

Limitation:

- It concerns only the object and not it's potential children for now to simplify.
- As there is an existing bug when creating softbody with scale not applied, I didn't took into account scale when spawning a softbody (only position and orientation (not ori.scaled(scale))

parenthesis: I changed my mind about Append from another file then call convertBlenderObject for several reasons:

- Seemed complex
- Append also append children but convertBlenderObject can't handle children
and for other reasons.

2 test files:

[AddSoftBody.zip](https://github.com/UPBGE/upbge/files/6565456/AddSoftBody.zip)
[dup_mat.zip](https://github.com/UPBGE/upbge/files/6565457/dup_mat.zip)


lordloki (approved):
I tested and it worked like a charm. The code seems me ok too.
Very good work!!!

youle31:
Thanks very much!

PR #1506, opened 2021-06-20
Logic editor UI enhancements

mysticfall:
I did a little experiment to revamp the logic editor UI, mainly to enhance the user experience for those who prefer writing components instead of logic bricks to create an UPBGE game.

Previously, the component and properties panels resided inside the logic brick editor so it was necessary to have it open even when you don't use logic bricks at all. And because the editor takes up quite a large space, it's usually necessary to open it in a separate space, thus forcing the users constantly switching back and forth between the spaces.

As such, I moved the components and properties panel into a separate properties tab as shown below to streamline the UI:

image

With the change, now you can manage UPBGE components and properties from anywhere without having to open logic brick editor, while also freeing up space for those who don't use components much.

There is an unfortunate caveat, however, that this change is not entirely backward compatible. If you open a previously saved .blend file after applying this patch, you will see the logic editor showing up like this with a console message _"Error, region type 4 missing in - name:"Logic", id:17"_:

image

You can drag the empty sidebar to minimise it, but in order to properly fix it, you'll need to delete and add the affected space again.

I'd like to ask if you think this would be a trouble worth taking for the suppose enhancements. Personally, I think it is, because the problem doesn't affect newly create files and benefit both groups of our users, those who rely on logic bricks and those who mainly use components.

As logic brick connections can be quite complex if you soley rely on using them, I think it'd be reasonable to maximise the editor space by hiding the region which most of those who use logic bricks won't use on regular basis.

### DIFF
--- a/release/scripts/startup/bl_ui/properties_game.py
+++ b/release/scripts/startup/bl_ui/properties_game.py
@@ -17,7 +17,6 @@
 # ##### END GPL LICENSE BLOCK #####
 
 # <pep8 compliant>
-import bpy
 from bpy.types import Panel, Menu
 
 
@@ -43,16 +42,20 @@ class GAME_PT_game_components(GameButtonsPanel, Panel):
         game = ob.game
 
         row = layout.row()
-        row.operator("logic.python_component_register", text="Register Component", icon="PLUS")
-        row.operator("logic.python_component_create", text="Create Component", icon="PLUS")
+
+        row.operator("logic.python_component_register", icon="PLUS")
+        row.operator("logic.python_component_create", icon="PLUS")
 
         for i, c in enumerate(game.components):
             box = layout.box()
             row = box.row()
+
             row.prop(c, "show_expanded", text="", emboss=False)
             row.label(text=c.name)
-            row.operator("logic.python_component_reload", text="", icon='RECOVER_LAST').index = i
-            row.operator("logic.python_component_remove", text="", icon='X').index = i
+            row.context_pointer_set("component", c)
+            row.menu("GAME_MT_component_context_menu", icon="DOWNARROW_HLT", text="")
+
+            row.operator("logic.python_component_remove", text="", icon="X").index = i
 
             if c.show_expanded and len(c.properties) > 0:
                 box = box.box()
@@ -61,6 +64,30 @@ class GAME_PT_game_components(GameButtonsPanel, Panel):
                     row.label(text=prop.name)
                     col = row.column()
                     col.prop(prop, "value", text="")
+
+
+class GAME_MT_component_context_menu(Menu):
+    bl_label = "Game Component"
+
+
+    @classmethod
+    def poll(cls, context):
+        ob = context.active_object
+
+        return ob and ob.game and ob.game.components and context.component
+
+    def draw(self, context):
+        layout = self.layout
+
+        components = context.active_object.game.components
+        index = components.find(context.component.name)  # FIXME: Should not use component.name as a key.
+
+        layout.operator("logic.python_component_reload", icon="RECOVER_LAST").index = index
+
+        layout.separator()
+
+        layout.operator("logic.python_component_move_up", icon="TRIA_UP").index = index
+        layout.operator("logic.python_component_move_down", icon="TRIA_DOWN").index = index
 
 
 class GAME_PT_game_properties(GameButtonsPanel, Panel):
@@ -713,6 +740,7 @@ class OBJECT_PT_levels_of_detail(ObjectButtonsPanel, Panel):
 classes = (
     GAME_PT_game_components,
     GAME_PT_game_properties,
+    GAME_MT_component_context_menu,
     PHYSICS_PT_game_physics,
     PHYSICS_PT_game_collision_bounds,
     PHYSICS_PT_game_obstacles,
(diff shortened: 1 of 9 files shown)


source/blender/windowmanager/WM_types.h
```diff
@@ -308,6 +308,7 @@ typedef struct wmNotifier {
 /* Changes to asset data in the current .blend. */
 #define NC_ASSET (26 << 24)
 #define NC_LOGIC (27 << 24)
+#define NC_GAME (28 << 24)
```
youle31, line 311: It is not called anywhere in the PR

source/blender/editors/space_buttons/buttons_context.c
```diff
@@ -608,6 +608,7 @@ static bool buttons_context_path(
     case BCONTEXT_OBJECT:
     case BCONTEXT_PHYSICS:
     case BCONTEXT_CONSTRAINT:
+    case BCONTEXT_GAME:
```
youle31, line 611: I'd move it that with the other case: which have a break; to avoid "fallthrough" I think

source/blender/makesrna/intern/rna_space.c
```diff
@@ -480,6 +480,7 @@ static const EnumPropertyItem buttons_context_items[] = {
      "Bone Constraints",
      "Bone Constraint Properties"},
     {BCONTEXT_MATERIAL, "MATERIAL", ICON_MATERIAL, "Material", "Material Properties"},
+    {BCONTEXT_GAME, "GAME", ICON_SCRIPT, "Game", "Game Object Properties"},
```
youle31, line 483: This is not really important, but Game Object properties suggests that there is an existing gameobject in the viewport whereas Gameobjects are only created after bge start. I'd rename this Object Game Components and Properties or something shorter

source/blender/windowmanager/WM_types.h
mysticfall: Yeah, it looks like we don't need to define a new node category for that. I'll remove it when I update the PR.

source/blender/editors/space_buttons/buttons_context.c
mysticfall: Maybe I'm missing something as I haven't looked much into what `buttons_context_path_object` actually does. But the fallthrough was intentional since I thought the excution path for `BCONTEXT_GAME` should be handled in a similar manner for the case with `BCONTEXT_OBJECT` because an `Object` is where `GameObjectSettings` instance is attached to.

Please let me know if that's not a correct behaviour.

source/blender/makesrna/intern/rna_space.c
mysticfall: Personally, I think it wouldn't be too confusing since it's not rare to find such settings in Blender UI that are only taken into effect when the game or animation is running. If it's ok to call it a "component panel" when `KX_PythonComponent` instance is only created in runtime, I suppose it wouldn't be too unreasonable to expect users to understand that a "game object" settings wouldn't imply that there's a live `KX_GameOjbect` somewhere inside the editor.

That being said, I don't object to rename it if there's a better alternative than "Game Object Properties". Maybe "Game Scripts" could be an option? (By the way, if I manage to implement KX_GameObject related stuffs as discussed on Discord, they will likely end up in the same panel too.)

mysticfall:
I updated the PR with component reordering feature and some minor enhancements:
image

release/scripts/startup/bl_ui/properties_game.py
```diff
@@ -17,10 +17,138 @@
 # ##### END GPL LICENSE BLOCK #####
 
 # <pep8 compliant>
-import bpy
 from bpy.types import Panel, Menu
 
 
+class GameButtonsPanel:
+    bl_space_type = 'PROPERTIES'
```
youle31, line 24: I'm wondering what exactly involves changing the bl_spacetype from LOGIC_EDITOR to PROPERTIES for game properties. I fear it can break some things.

mysticfall: Oh, it's a new panel for the properties tab. I restored the old one for the logic editor (`LOGIC_PT_properties`) after our conversation on Discord, so it should work as it did before.

youle31: Ah yes, it's ok sorry.

mysticfall: No trouble :)

source/blender/editors/space_buttons/buttons_context.c
youle31: yes, indeed, looks good after looking at it in visual studio 👍

youle31 (approved):

youle31:
Thanks for the great work @mysticfall !

I sounds good to me.

I tested with this for an existing component:

[sound_component.zip](https://github.com/UPBGE/upbge/files/6719480/sound_component.zip)

and with that for UI behaviour:

[comp.zip](https://github.com/UPBGE/upbge/files/6719481/comp.zip)

lordloki (approved):
I tested with blends from youle and they worked like a charm. Code seems ok to me too.
Great work!!

mysticfall:
Thanks much for the review as always, both of you! :)

I'll merge this PR now.

PR #1513, opened 2021-07-04
Allow extending KX_GameObject from Python scripts

mysticfall:
This pull request adds `Game Object` panel to the properties window, allowing users to extend `KX_GameObject` derivatives (e.g. `KX_Light`, `KX_Camera`) with a custom Python class:

image

It follows the same symantics as what `KX_PythonComponent` provides and the order of initialisation gaurantees that `KX_GameObject.start()` is invoked after all of its dependent `KX_PythonComponent.start()` is called.

It also fixes a memory leak that increased Python reference count of each component every tick and optimises the way callbacks on a Python proxy is invoked.

Because of my limited understanding of C++ and Blender internals, I'd suggest to merge it only after we can be sure that it wouldn't break existing features.

I believe this change would introduce some interesting possibilities regarding how game scripts can be organised and designed in UPBGE, on which I'll write a short article and post it somewhere soon so that our users may have a better understanding of the motivation behind this proposed change.


mysticfall:
Maybe the commit `44b0918` isn't needed any more as I changed my approach after that. But probably it wouldn't hurt to keep it that way, in case we may want to override the object type on per-instance basis in future.

mysticfall:
I just wrote an article to explain the motivation behind this proposed change:
https://github.com/mysticfall/alleycat/wiki/Understanding-Inheritance-And-Composition-In-Game-Scripting

I hope it could help you understand why I regard this to be an important feature for my use case.

source/blender/editors/space_logic/logic_ops.c
```diff
@@ -829,6 +829,182 @@ static void LOGIC_OT_region_flip(wmOperatorType *ot)
...
+  }
+
+  ob->custom_object = pycomp;
+
+  WM_event_add_notifier(C, NC_LOGIC, NULL);
+
+  return OPERATOR_FINISHED;
+}
+
+static int custom_obeject_create_exec(bContext *C, wmOperator *op)
```
lordloki, line 863: custom_obeject_create_exec should be custom_object_create_exec

mysticfall: Fixed it! :)

mysticfall:
I fixed the typo and the crash problem. I also made a separate commit to rename `PythonComponent*` to `PythonProxy*`.

lordloki, youle31 // Could you review the changes again? Thanks!

youle31 (approved):

youle31:
As said on discord, I find it confusing, but I runned tests with this file:

https://ufile.io/txd6nap7

and as I didn't see major issues and as lordloki voted for it, I approve.

youle31:
I didn't see that GetReplica was changed, I need to test

lordloki (approved):
Hi,
The code seems me ok to me. At least the changes seem me reasonables as my coding level is not so high to review this change deeply. I tested with several files and they worked great.

Mysticfall, this is a very great work, congratulations. Additionally, I would like to thank to youle for the great effort reviewing and testing.

mysticfall:
Thanks much for the review, both of you! I'll merge and close this PR now :)

PR #1515, opened 2021-07-13
UPBGE: Fixes for 2D filters on mac + 1 tmp fix on linux

youle31:
I tested with several files:

[2dfilters.zip](https://github.com/UPBGE/upbge/files/6811367/2dfilters.zip)

Can be reverted with git UI if causing issues.


BluePrintRandom:
So this means mac m1 works with upbge right?

youle31:
yes, the remote sessions were done on a Mac with a M1 thing

PR #1524, opened 2021-08-01
Add support for Python's logging API

mysticfall:
This PR adds basic support for Python's builtin [logging API](https://docs.python.org/3/howto/logging.html).

image

For now, it only works with the _Viewport Render_ mode and does not allow using an external configuration file for fine-grained control. Hopefully, those limitations might be lifted once the PR gets merged.


mysticfall:
Here's a sample .blend file to test the PR:
[logging-test.tar.gz](https://github.com/UPBGE/upbge/files/6912017/logging-test.tar.gz)

mysticfall:
I updated the PR to add `KX_GameObject.logger` and `KX_PythonComponent.logger` which configure and return a `logging.Logger` instance so that the users don't have to do it themselves.

youle31 (approved):

lordloki (approved):
Seems ok to me. Great work.

Just a hint: We will have to add a note in the release notes pointing out that this works with "use viewport render" active only.

mysticfall:
Thanks @lordloki and @youle31 for the review, as always! :) I'll merge the PR and delete the branch now.

I agree that we better mention the limitation in the release note. By the way, it still sort of works with non-viewport render mode, just not updating in realtime.

mysticfall:
By the way, for anyone who might need a more fine-grained control, you can add `logging.json` to the project directory:
```json
{
  "version": 1,
  "loggers": {
    "root": {
      "level": "WARNING",
      "handlers": [
        "stdout",
        "console"
      ]
    }
  },
  "formatters": {
    "simple": {
      "format": "[%(levelname)s] %(name)s - %(message)s"
    }
  },
  "handlers": {
    "stdout": {
      "class": "logging.StreamHandler",
      "formatter": "simple",
      "stream": "ext://sys.stdout"
    },
    "console": {
      "class": "alleycat.log.ConsoleLogger",
      "formatter": "simple"
    }
  }
}
```
You can modify the above example to do various things like adjusting logging level for a specific part of the code (e.g. the player object) or add more handlers (e.g. write to a file or to an in-game console/overlay).

PR #1538, opened 2021-08-23
UPBGE: Fix Export as runtime

youle31:


lordloki:
I did several tests with new blends and former blends and they worked well. Great job!!

lordloki (approved):

PR #1540, opened 2021-08-25
CEGUI Actuator take 2

lordloki:
I have updated the cegui branch to rebase current master. The necessary libs to compile are here: https://github.com/UPBGE/lib
The additional dataset is here: https://mega.nz/file/Fl0EhYZB#oWMZOn3eeTX7nStKMqg22AG_vExmU01cEcT80KayyW8

This dataset has to be extracted and it has 2 functions: 
- First, it is the default placed where CEGUI checks for templates. Then you will have to modify the directories at LA_Launcher.ccp at InitEngine( ) in  rp->setResourceGroupDirectory setting functions.
- Scheme and Layouts will have to be called from CEGUI actuator. I will post an image later.

Here is a test blend: [prueba.zip](https://github.com/UPBGE/upbge/files/7043655/prueba.zip)
image

Current state:
The screen is black when start the engine and there is a crash at exit I can't debug yet.

Teorically with the example exposed above would have to show this: https://vimeo.com/27969738

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -210,6 +210,9 @@ endif()
 option(WITH_GAMEENGINE_BPPLAYER "Enable Blend encrypted (from BPPlayer application) reading capabilities" ON)
 mark_as_advanced(WITH_GAMEENGINE_BPPLAYER)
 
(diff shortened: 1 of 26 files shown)


BluePrintRandom:
as BGUI has been implemented this is on the back burner for now**

PR #1548, opened 2021-09-05
UPBGE: Recover Activity culling

youle31:
I think it works as excpected but SuspendLogic doesn't work properly (it is not related to activity_culling patch and there is the same issue in 0.2.5).

Physics and Actions are correctly suspended/restored.

test file:

[activity_culling_test.zip](https://github.com/UPBGE/upbge/files/7114273/activity_culling_test.zip)


lordloki (approved):
Code seems ok to me. I tested with your example and with a blend that i have from old test files. They worked perfectly. Great work!!!

[ActivityLogicAnimation.zip](https://github.com/UPBGE/upbge/files/7124193/ActivityLogicAnimation.zip)

PR #1550, opened 2021-09-10
Suppress KX_GameObject::Update when it's suspended

mysticfall:
As discussed with youle on Discord.

### DIFF
--- a/source/gameengine/Ketsji/KX_GameObject.cpp
+++ b/source/gameengine/Ketsji/KX_GameObject.cpp
@@ -1861,13 +1861,15 @@ void KX_GameObject::SetComponents(EXP_ListValue<KX_PythonComponent> *components)
 void KX_GameObject::Update()
 {
 #ifdef WITH_PYTHON
-  if (m_components) {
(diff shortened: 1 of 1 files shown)


youle31:
Sorry, I was not concentrated on discord, I didn't understand...

About IsActionsSuspended it concerns only actions manager (playAction)

When you suspend logic it should stop both logic bricks and actions currently.

But an API could be implemented later to suspend only actions.

I think instead of IsActionsSuspended, SCA_IObject::m_suspended should be used.

mysticfall:
Sorry, looks like _I_ wasn't concentrated either :) Anyway, changed it as requested. Thanks!

youle31:
Thanks! It's ok for me. Later i'll do few modifs in the global naming/system to see more clear about  LogicAndAction or OnlyAction suspended.

youle31 (approved):

youle31:
I take the liberty to merge this. If some issues are noticed later we can fix later. Thanks very much @mysticfall  !

PR #1556, opened 2021-09-16
UPBGE: Attempt to update ffmpeg bge code to avoid linux warnings

youle31:
No need to review, unless we decide to fix linux warnings.

I just do the PR to have a better code visualisation

Note:

- If we want the code to work with older versions of ffmpeg lib we can add #ifdef FFMPEG_OLD_CODE (See FFMPEG_READER for example)


youle31:
I close this for now

lordloki:
youle, I open again this pull to can see the diff better.

source/gameengine/VideoTexture/VideoFFmpeg.cpp
```diff
@@ -68,6 +68,7 @@ const double defFrameRate = 25.0;
 // constructor
 VideoFFmpeg::VideoFFmpeg(HRESULT *hRslt)
     : VideoBase(),
+      m_codec(nullptr),
```
lordloki, line 71: m_codec variable can be removed. It is declared, initialized and asigned but never used.

lordloki (approved):
I just checked the code and it seems ok to me. I tested with video play and camera stream and it works ok.
I think we can commit it now and whether any issue appears we will fix it. Anyway, we will have to make the change sooner than later due to ffmpeg5.0 is coming to Blender.

PR #1558, opened 2021-09-23
Fix wrong init value for tp_vectorall_offset

mysticfall:
This PR eliminates the following warning message occurning while compiling `python_proxy.c`:
```
warning: initialization of ‘long int’ from ‘void *’ makes integer from pointer without a cast NULL
```

`FakeType` defines a structure to create a new Python type from the native side. And according to the [relevant documentation](https://docs.python.org/3/c-api/typeobj.html), the problematic slot used to be `tp_print` which expected a function pointer (for which `NULL` is a legal value).

However, it looks like they changed it to `tp_vectorall_offset` which expects `Py_ssize_t` that translates to `int` or `long` depending on the target platform.

And because `NULL` is defined as `((void *)0)` in C, so it's not a valid value for non-pointer type such as `long`.

### DIFF
--- a/source/blender/blenkernel/intern/python_proxy.c
+++ b/source/blender/blenkernel/intern/python_proxy.c
@@ -142,7 +142,7 @@
         sizeof(PyObject),                               /* tp_basicsize */ \
         0,                                              /* tp_itemsize */ \
         (destructor)NULL,                               /* tp_dealloc */ \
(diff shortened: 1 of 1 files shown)


youle31 (approved):
Thanks very much!

lordloki (approved):

PR #1561, opened 2021-09-25
UPBGE: SDL Check macro removal

lordloki:
There is no need to maintain this system as we compile with SDL 2.0+ statically always. Additionally, to comment that this system was done during the transition between SDL 1.2 to SDL 2.0. Now this transition is more than over.


youle31 (approved):
Great! Thanks very much!

PR #1566, opened 2021-10-03
UPBGE: System to remap bone constraints targets after armature duplication

youle31:
It is if we decide to not rework armature sensor/actuator.

It is doing an automatic remapping only if "targets" are children of Armature.

EDIT: It is because we know that "targets" are duplicated if they are children of Armature (as children are spawned with parent).

After we did an armature duplication, if one of his children is a "bone constraint target", this child becomes the new armature bone constraint target.

EDIT: Actually, i'm not sure about the logic if there are several constraints with several targets/subtargets... would maybe need to find a more solid system...

Test file:
[constraintRemapTest.zip](https://github.com/UPBGE/upbge/files/7274624/constraintRemapTest.zip)

EDIT: I'll change this later to take into account bone


youle31:
I'll reopen this later when fixed

PR #1612, opened 2021-12-19
Fix build problem with Python 3.10 due to internal API references

mysticfall:
This change will fix build errors on Python 3.10 caused by using its internal API (i.e. functions prefixed by `_`).

List below are issues dealing with similar problems:

* https://developer.blender.org/T85573
* https://bugzilla.redhat.com/show_bug.cgi?id=1912903

### DIFF
--- a/source/gameengine/Ketsji/KX_PythonInit.cpp
+++ b/source/gameengine/Ketsji/KX_PythonInit.cpp
@@ -2543,8 +2543,7 @@ void createPythonConsole()
   BLI_strncpy(filepath, BKE_appdir_folder_id(BLENDER_SYSTEM_SCRIPTS, "bge"), sizeof(filepath));
   BLI_path_append(filepath, sizeof(filepath), "interpreter.py");
 
-  // Use _Py_fopen to make sure we use the same fopen function as python use.
-  FILE *fp = _Py_fopen(filepath, "r+");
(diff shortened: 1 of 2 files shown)


source/gameengine/Ketsji/KX_PythonInit.cpp
```diff
@@ -2544,7 +2544,7 @@ void createPythonConsole()
   BLI_path_append(filepath, sizeof(filepath), "interpreter.py");
 
   // Use _Py_fopen to make sure we use the same fopen function as python use.
```
lordloki, line 2546: Maybe the comment should modify to "// Use fopen as _Py_fopen was removed from Python 3.10" or something like that.

lordloki (approved):
Other than the comment, the changes looks good to me. Good work.

source/gameengine/Ketsji/KX_PythonInit.cpp
mysticfall: Oh, I missed it somehow. Updated the PR after removing the comment. Thanks for the review!

youle31 (approved):
It sounds ok and compiles ok with python 3.9. Thanks very much for the work!

mysticfall:
Thanks both of you! I'll close and merge this now.

PR #1619, opened 2021-12-26
Add base type info to 'fake' BGE type definitions

mysticfall:
This fixes the problem of KX_GameObject subtypes (e.g. KX_Camera) not being recognised as a valid base class for a custom game object.

How to reproduce the error
------------------------------------

* Write a custom class that inherits `KX_Camera`
* Select a camera object and open the _Game Object_ panel.
* Choose _Select_ and try to assign the custom class.

Prior to this patch, it will fail with an error message saying the provide class is not a valid subtype of `KX_GameObject`.

### DIFF
--- a/source/blender/blenkernel/intern/python_proxy.c
+++ b/source/blender/blenkernel/intern/python_proxy.c
@@ -54,89 +54,89 @@
 #ifdef WITH_PYTHON
 
 #  define FAKE_TYPES \
-    FT_DEF(BL_ArmatureBone) \
-    FT_DEF(BL_ArmatureChannel) \
-    FT_DEF(BL_ArmatureConstraint) \
-    FT_DEF(BL_ArmatureObject) \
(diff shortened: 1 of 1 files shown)


youle31 (approved):
I tested with this file:
[comp.zip](https://github.com/UPBGE/upbge/files/7776572/comp.zip)

and it was ok. I didn't notice compil warnings and code looks ok.

Minor thing: alphabetical order not always respected.

Thanks very much for the work!

mysticfall:
> Minor thing: alphabetical order not always respected.

Yeah, that's what annoyed me too. Because the entry that defines `KX_GameObject` needs to appear before the others that depend on it, I had to break the order.

So I chose to move the whole `BL_*` block down instead of just picking out `BL_ArmatureObject`, for example, to keep the alphabetical order as much as possible.

Thanks very much for the review, as always :)

lordloki (approved):

PR #1624, opened 2022-01-02
Fix module path problem on Python 3.10

mysticfall:
Due to a change of behaviour in Python 3.10, running a game in embedded mode once will now make Blender lose all system module paths like `/scripts/modules`, resulting in various problems.

Before 3.10, `Py_InitializeFromConfig` preserved `sys.path`, so backing up the environment after that point didn't cause much of a problem.

However, the behaviour seems to have changed in 3.10 (I couldn't find any reference in documentation or changelog, though) which results in loss of system module paths such as `scripts/modules` after running the game once in embedded mode.

This PR tries to fix the problem by taking a back before `Py_InitializeFromConfig` so that it can be correctly restored after the game finishes.

How to reproduce:

 * Launch UPBGE compiled with Python 3.10.
 * Open scripting console and write `import sys` and `print(sys.path)`.
 * Confirm that the printed output includes system module paths.
 * Run embedded start once and repeat the above process.
 * Confirm that `sys.path` gets reset as shown below:

image

### DIFF
--- a/source/gameengine/Ketsji/KX_PythonInit.cpp
+++ b/source/gameengine/Ketsji/KX_PythonInit.cpp
@@ -2309,6 +2309,8 @@ void initGamePythonScripting(Main *maggie, bContext *C, bool *audioDeviceIsIniti
     PyPreConfig preconfig;
     PyStatus status;
 
+    backupPySysObjects();
+
     if (BPY_python_get_use_system_env()) {
       PyPreConfig_InitPythonConfig(&preconfig);
     }


mysticfall:
By the way, one notable problem that can be caused by the changed behaviour is failing to import `aud` module when running in the embedded mode.

It's because the module depends on `numpy` which is supposed to be loaded from the cleared out module path.

youle31 (approved):
From discussion on discord, it sounds ok. I tested with several files and found no bug related to this patch, then it sounds alright. Thanks for the work!

lordloki (approved):

lordloki:
I tested and it worked well with python 3.9

mysticfall:
Thanks for the testing! I'll merge and close this now.

PR #1628, opened 2022-01-09
Update workflow action for UPBGE 0.3 / Python 3.10

mysticfall:
Update workflow action for UPBGE 0.3 / Python 3.10.

### DIFF
--- a/.github/workflows/publish-api-and-stubs.yml
+++ b/.github/workflows/publish-api-and-stubs.yml
@@ -8,7 +8,7 @@ on:
   workflow_dispatch:
 
 env:
-  PYTHON_VERSION: "3.9"
+  PYTHON_VERSION: "3.10"
 
 jobs:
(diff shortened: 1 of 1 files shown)


youle31:
idk if it matters but i saw this: s/#APP_VERSION#/0.3.1

We choosed to use 0.31.0 version internally because we thought that using 0.3.1 would cause versioning (versioning_upbge.c) issues.

mysticfall:
I think it'd be ok since it's only for determining the version of the Python module. The reason I decided it that way was because I wasn't sure if PyPI or any of Python's dependency management tools would properly recognise `0.4` as newer than `0.31` for instance.

lordloki (approved):
Changes seems ok to me. Good work!!

mysticfall:
Thanks! I'll merge and close this now :)

PR #1630, opened 2022-01-10
UPBGE: Use old undo only at bge exit but new undo in viewport

youle31:
I used a new Global ... idk if it is good


lordloki (approved):
I tested with several files and it worked perfectly. Good work.

Regarding the global variable, I think that this is our first global, it is not a problem.

youle31:
Great! Thanks very much @lordloki for tests! I tested too on my side and it seemed ok then I merge.

PR #1664, opened 2022-02-23
Add OpenImageIO dependency to fix CI build

mysticfall:
This fixes the recently introduced CI build problem.


lordloki (approved):
It seems me ok. 

Have you seen that from several versions ago the python AUD module documentation is not generated? I didn't see where the error is.

mysticfall:
> It seems me ok.
> 
> Have you seen that from several versions ago the python AUD module documentation is not generated? I didn't see where the error is.

Oh, I didn't know that. Another mystery to solve than :) I'll see if I can figure out what went wrong with it this weekend.

I'll merge this one now. Thanks for the review.

PR #1666, opened 2022-02-24
Fix problem with numpy error while building API doc

mysticfall:
This should restore the missing documentation with `aud` module.


lordloki (approved):
Let's go to check it!!! :-)

mysticfall:
I'll merge this now so that it can run a few hours later. Let's see if it works :)

PR #1676, opened 2022-03-12
Additional changes for FFmpeg 5.0 compatibility

mysticfall:
Fixes a build problem on Linux (e.g. AUR upbge-git package).

### DIFF
--- a/source/gameengine/VideoTexture/VideoFFmpeg.cpp
+++ b/source/gameengine/VideoTexture/VideoFFmpeg.cpp
@@ -51,6 +51,7 @@ extern "C" {
 #  include "libswscale/swscale.h"
 #  include <libavcodec/avcodec.h>
 #  include <libavutil/imgutils.h>
+#  include <libavformat/version.h>
 }
 
 // default framerate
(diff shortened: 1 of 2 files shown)


mysticfall:
I managed to make it compile again on Linux but couldn't verify it since it crashed with a seemingly unrelated segfault.

lordloki // Could you see if it's working on your end please? I didn't include youle as a reviewer since he's taking a leave.

lordloki:
Hi,
The main issue that I have to test ffmpeg5 is that the windows ffmpeg libs are compiled in a way that the function `avformat_open_input(AVFormatContext **ps, const char *url, ff_const59 AVInputFormat *fmt, AVDictionary **options);` for ffmpeg < 5 uses a not const AVInputFormat. Then if we define AVInputFormat as const it gives compiling error.

I have done an additional patch https://gist.github.com/lordloki/235e010f936c717cde17e4bc07d5ffce making the correct definition for const AVInputFormat *inputFormat in each case.

Also, I fixed the current master for the python issue.

mysticfall:
Thanks for the heads up about the `txt_to_buf` issue. It indeed fixed the crash for me so now I can test it on my end. I was able to run the attached example which I found on Discord and it appears to run correctly after the patch.

As to the problem with `const`, I assume it'd be a temporary measure until the Blender people update it to depend on FFmpeg 5.0?

If the fact that I was able to compile and run UPBGE against FFmpeg 5.0 means the non-UPBGE part of Blender is already compatible with the new version, I wonder if it'd be ok to just bump `FFMPEG_VERSION` in `versions.cmake` to `5.0`.

I'll just apply the suggested change if that's not an option. By the way, where should I put the declaration for  `FF_API_AVIOFORMAT` in that case?

lordloki:
I just committed in the branch the code to make it pre-ffmpeg5 compatible. I tested with camera and movie files and it works perfectly.

Great work!!

lordloki (approved):

mysticfall:
Thanks much! I'll merge this one now :)

PR #1696, opened 2022-04-28
UPBGE: Fixing crash using volumetrics

lordloki:
After last changes in curves, volumetrics and smoke we have to add new DRW initialization for these elements.
Curves initializes and update the hair methods now.

I have some doubts with DRW_globals_update() to update UBO's. This function was out of our DRW_game_render_loop function due to any specific reason?

I tested with a few files and they worked well.

[volumetric_crash.zip](https://github.com/UPBGE/upbge/files/8586684/volumetric_crash.zip)

### DIFF
--- a/source/blender/draw/intern/draw_manager.c
+++ b/source/blender/draw/intern/draw_manager.c
@@ -3437,10 +3437,14 @@ void DRW_game_render_loop(bContext *C,
 
   const int object_type_exclude_viewport = v3d->object_type_exclude_viewport;
 
-  DRW_hair_init();
+  /* Update UBO's */
+  DRW_globals_update();  
 
   /* Init engines */
   drw_engines_init();
+  DRW_curves_init();
+  DRW_volume_init(DST.vmempool);
+  DRW_smoke_init(DST.vmempool);
 
   drw_engines_cache_init();
   drw_engines_world_update(DST.draw_ctx.scene);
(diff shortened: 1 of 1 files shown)


youle31:
Hello,

I didn't notice these changes, thanks very much!

here in DRW_engine_render_loop_ex:
Capture d’écran 2022-04-28 235351

curves, volumes and smoke_init are before drw_engines_init().

About update global UBO i didn't add it previously because I thought it was not needed but I didn't look since a long time and it can be good to add it. Thanks!

source/blender/draw/intern/draw_manager.c
```diff
@@ -3437,10 +3437,14 @@ void DRW_game_render_loop(bContext *C,
 
   const int object_type_exclude_viewport = v3d->object_type_exclude_viewport;
 
-  DRW_hair_init();
+  /* Update UBO's */
+  DRW_globals_update();  
 
   /* Init engines */
   drw_engines_init();
+  DRW_curves_init();
```
youle31, line 3445: Is i working if we put that before drw_engines_init?

youle31 (approved):

source/blender/draw/intern/draw_manager.c
lordloki: Yes, it works ok. I move them.

lordloki:
I also added it a GPU_BACKEND_OPENGL check when we call GPU_flush due to new METAL backend.

Thanks for reviewing and testing. I apply the pull.

PR #1699, opened 2022-05-14
UPBGE: BGUI library added to modules folder

lordloki:
The BGUI library has been reworked to be OpenGL agnostic. Now it is totally based in blf, gpu, bge and bpy modules. 
If you want to see the modifications done from original BGUI library check this repository: https://github.com/lordloki/bgui
I included the BGUI module in the modules folder similar to other specific modules as 'gpu_extras', etc.

I removed the video widget because i didn't find a way to refactor it.

These are 3 examples that work perfectly: 
[examples.zip](https://github.com/UPBGE/upbge/files/8693534/examples.zip).

I included new rst files to the python api replicating the info shown at old BGUI readthedocs API.

image


BluePrintRandom:
=D

youle31:
Awesome! I tested the 3 examples and they seem to work perfectly!

- Tested in embedded and blenderplayer and was ok.
- Tested if we can still see what is behind the UI and ok.
- No issue noticed after bge exit (gpu blend state none is fine I think)

I didn't tested complex cases and don't really know how it works to handle enable/disable UI and handle UI keyboard/mouse events vs viewport keyboard/mouse events but as this part did not really changed since old bge, I guess it is ok.
I didn't compiled API to see if there are warnings but you seem to have done this very seriously.

For example files, idk where they can be uploaded to be visibled.
In 2 of 3 example files there were minor warnings related to keyboard.events/mouse.events deprecated (keyboard.inputs) but it is not important.

Big congrats for this amazing work!! And Thanks!

youle31 (approved):

lordloki:
Thank you for the extensive review!!. The examples I will upload in the manual as I will do a new chapter under tutorials (I will fix the keyboard/mouse event deprecation warnings too)

mysticfall:
First of all, thanks for the great work! I still think a CEF-based approach would be a more comprehensive solution in the long run. But CefPython being stuck in the past and Blender's new GPU module still lacking needed format support, I can see that BGUI could be our only feasible option to create non-trivial UI assets at the moment.

The only issue I could find in the PR is that the RST files don't follow the rules of Sphinx' [Python domain](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#the-python-domain).

Ideally, they can be changed to look like the existing documentation on classes like `KX_GameObject` to make them correctly parsed by my stub generator.

But I suppose it should be ok since incorrect RST syntax is less problematic for people who read it on API documentation page than a program which parses it to generate stubs.

So, I don't object to merging the PR and revising it later over time.

I'll approve the PR as it is. Thanks again for the great work! :)

mysticfall (approved):

lordloki:
Thanks for the reviewing!!. Yeah, I think that the mock-ups CEF-based that you showed they had more potential that this basic GUI. 
This is something simple to cover the gap until with can reach to that level.

Regarding rst files don't passing the stub generator I didn't take into account (I check API generation only). 

I'm going to commit the pull now. After if you can modify a basic rst as progress bar.rst to see how i have to modify, I will modify the rest of rst files following that guideline.

PR #1706, opened 2022-06-02
UPBGE: Missing errors reports + python controller module reload

youle31:
Since: https://github.com/UPBGE/upbge/commit/ccbf35bf672408e4406548266a352c4227ca508d
some errors which were reported in module mode in some cases were not
reported anymore. (was working in 0.2.5 for embedded Texts at least)

test file: 
[untitled.zip](https://github.com/UPBGE/upbge/files/8821136/untitled.zip)

In previous commit I reverted this and in this one I tried another
fix for the bug https://github.com/UPBGE/upbge/commit/ccbf35bf672408e4406548266a352c4227ca508d was fixing
about reloading scripts during runtime option on python controllers (D)

As I added a line in python_proxy.c and as it concerns 2 bugs, and as I'm not really comfortable with the code, I add all of you as reviewer if you can quickly test and/or review code if you want.

Thanks!

- Not reported errors test file: [untitled.zip](https://git
[untitled.zip](https://github.com/UPBGE/upbge/files/8821174/untitled.zip)

- Python Controller Debug option: 
[001_reloadme.zip](https://github.com/UPBGE/upbge/files/8821169/001_reloadme.zip)

EDIT: After this commit i'll also add an extra debug info for an error which is sometimes spawned for a reason I don't get (both in master and here when calling bpy_text_import_name: This message is sometimes spawned: ERROR: bpy_import_main_set() was not called before running python. this is a bug.)
I'll add the name of the module trying to be imported to try to understand why maggie is not set.

### DIFF
--- a/source/blender/blenkernel/intern/python_proxy.c
+++ b/source/blender/blenkernel/intern/python_proxy.c
@@ -511,6 +511,7 @@ static bool load_class(PythonProxy *pp,
       } \
       Py_DECREF(pypath); \
     } \
+    bpy_import_main_set(NULL); \
     PyGILState_Release(state); \
     return value;
 
--- a/source/blender/python/generic/bpy_internal_import.c
+++ b/source/blender/python/generic/bpy_internal_import.c
@@ -108,6 +108,18 @@ struct Main *bpy_import_main_get(void)
 void bpy_import_main_set(struct Main *maggie)
 {
   bpy_import_main = maggie;
+
+  /* Also restores original blender reload method backed up in bpy_import_init at ge exit */
+  if (maggie == NULL) {
+    if (imp_reload_orig) {
+      PyObject *mod = PyImport_ImportModuleLevel("importlib", NULL, NULL, NULL, 0);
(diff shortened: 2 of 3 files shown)


lordloki (approved):
I tested and it worked ok. Thanks for the work.

mysticfall:
Sorry for being late to test. I just tried to run the test file on both the PR and master branch build but couldn't find any difference.

In both cases, I got the following error without details (i.e. `NameError: name 'a' is not defined`):
```
Error: Python(Cube), Python module can't be imported
ImportError: Failed to import module : 'module'
```
Maybe I'm missing something obvious or not understanding the context. Could you let me know how I should test the files?

youle31:
After more testing, discussion with mysticfall, it seems that I went totally confused while doing/testing this Pull Request which appears to be wrong, then i'll close it and replace with another commit (discussed on discord) with the hope it will fix the issue and that it won't appear again. Thanks very much for tests, and sorry about my mistake.

PR #1730, opened 2022-08-19
Gi

e2002e:
This repository is only used as a mirror of git.blender.org. Blender development happens on
https://developer.blender.org.

To get started with contributing code, please see:
https://wiki.blender.org/wiki/Process/Contributing_Code


BluePrintRandom:
what is this?
SSGI upbge?

e2002e:
Hi, yes this is a merge from upbge and a the ssgi build. Unfortunatelly it works only with blender 3.x.x and the branch doesn't implement the gameengine.

BluePrintRandom:
if it works in the viewport it should work in 'use viewport mode'

PR #1731, opened 2022-08-19
patched 3.0.0 for SSGI in Eevee

e2002e:
Hello, I found out that there is an addon for global illumination that is powerfull (https://n0451.gumroad.com/l/PgyXc).
I tried to find the code to include in upbge because the project turned out to be a build of blender, and ended up emailing the author, who answered with a .patch file... so git apply has saved me painful times of copying. The base branch is 3.0.0 because and I had only to patch manually 3 files by checking out to this branch.
The only thing i'm worried about is that I can't compile, there are missing python libs... I compile fine on the clone of your repository. Must be about the CMakeList.txt... ?
The author did not recommend to patch a blender version before 3.1 but I can't find such version.


lordloki:
hi @e2002e ,

as you can see in the link (https://code.blender.org/2021/06/eevees-future/) this ssgi patch or something based on it, will be include in the eevee rewrite. 
Additionally, it introduces a LOT changes over the Eevee engine and we try to avoid as much as possible (it creates a lot merge issues).

Then the way to get this feature is simply to wait to this feature be included in Blender in the Eevee next rewrite.

e2002e:
Go that, trying my own build for now.

PR #1735, opened 2022-08-21
Update platform_unix.cmake

e2002e:
Doesn't need ${} to test llvm_version nor quotes around 4.0.0
This prevented compilation when using precompiled libraries.


lordloki:
The current code in platform_unix.cmake is correct. It is necessary th ${} because LLVM_VERSION is a CMAKE variable that is set in versions.cmake ```set(LLVM_VERSION 12.0.0)```

e2002e:
I am changing from zorin to kunbuntu, got to much compatibility problems with development (missing newest libraries and softwares).

PR #1747, opened 2022-09-14
Upbge v0.3 release

e2002e:
This is the current release with ssgi as discussed in #1746. This was written a while ago as an addon, and then got included directly into the C code, author made a build dedicated to this module and kindly gave me the patch.
Can be found at: https://n0451.gumroad.com/l/PgyXc

Now, with this patch included, the only thing that would still be needed to compete with other engines at the graphics level, is tesselation.

Screenshot_20220914_223638


e2002e:
Still there is a issue though, when refraction is activated, reflections disappear. I'm going to contact the author today.

e2002e:
I have made my little researches and eevee gets a support for raytrace, they are working with the guy I contacted and will implement hardware support.
I'm closing this.
Will take more time to take care of my cultures of royalkush.

e2002e:
My issue with upbge-v3.0 is that it requires python3.9, which is not so easily compiled in the latest linux.

PR #1748, opened 2022-09-15
Fix compile problem due to missing avcodec header

mysticfall:
This fixes the building problem due to a missing include statement of `avcodec.h`.

I'm not entirely sure why it's only happening when building UPBGE on Linux with system dependencies or if it's the correct way to address the problem.

But it looks to solve the issue on my end so I'd appreciate if someone more knowledgeable (reads, either youle or ExxiL 😉) could verify the change.

Thanks!

### DIFF
--- a/source/gameengine/VideoTexture/VideoFFmpeg.h
+++ b/source/gameengine/VideoTexture/VideoFFmpeg.h
@@ -21,6 +21,7 @@
 
 extern "C" {
 #  include "ffmpeg_compat.h"
+#  include <libavcodec/avcodec.h>
 #  include <pthread.h>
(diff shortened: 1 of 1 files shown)


lordloki (approved):
One more include doesn't hurt to anyone :-)

youle31 (approved):
No pb on MSVC with this. Thanks you!

mysticfall:
Thanks, both of you! I suspect there might be another issue with the build because I also have failed runs with the Blender snapshot pipeline in my bpystubgen repo. I'll make a follow-up PR to deal with it if it proves to be the case.

Issue #1718, opened 2022-07-24
UPBGE .33 27/7/2022 will be closed after removing nodes in logic brick node view (Bricky Nodes addon)


BlackEagle1983:
Hello . when I delete a node in logic brick node view UPBGE .33 will be closed without any error message . 

windows 10 pro
cpu i5 4690
ram 12
graphic GTX 950 2 gig

youle31:
Video shared by report author:
[1.zip](https://github.com/UPBGE/upbge/files/9179360/1.zip)

Thanks for report, i'll investigate later on it

youle31:
I can't reproduce the issue in last upbge master (last upbge builded from sources)

Could you:

1. Share a test file
2. Wait for next build and test again AND if it doesn't crash anymore, close this report

Thanks

BlackEagle1983:
> I can't reproduce the issue in last upbge master (last upbge builded from sources)
> 
> Could you:
> 
> 1. Share a test file
> 2. Wait for next build and test again AND if it doesn't crash anymore, close this report
> 
> Thanks

Yes .of course . Thanks

BlackEagle1983:
[test.zip](https://github.com/UPBGE/upbge/files/9188186/test.zip)
This is the test file . The Issue is in Logic Brick Node View Section .

IzaZed:
This is an issue with the bricky nodes addon, not upbge itself (I think)
I will check it out later this week

youle31:
I can't reproduce the crash in last master. Imo it could be a blender bug which has been fixed like: beb746135dbe0c5ca21c2334e7dc475f201ad71e or similar

IzaZed:
It does seem to be a blender bug as it's working in UPBGE 0.31 and the bricky node addon was last updated before that version was released

youle31:
I'll investigate on this later (at autumn maybe). I managed to reproduce it on my other PC.

(It is either a Blender bug, either maybe a bug related to bpy.ops.logic undo changes done few monthes ago)

youle31:
Hi, I did a build with a potential fix here: https://ufile.io/i8efpi3z for windows 64.

Could you test and see if bricky nodes is still crashing or if there are other anomalies? Thank you

PR #1753, opened 2022-09-27
UPBGE: Attempt to fix issues with Bricky Nodes

youle31:
There were reports here: https://github.com/UPBGE/upbge/issues/1718 and on discord about Bricky Nodes addon crashes.

I saw it was related to changes I did in logic_ops related to undo (to use old undo for logic operators).

It seems this code avoids some crashes, even if I don't know exactly what happens internally.

IzaZed and arch-sketch tested and confirmed it was better after this patch.


lordloki:
It seems ok to me. Good job!!

youle31:
Thanks for review! I merge.

PR #1758, opened 2022-10-07
UPBGE: Try to fix asserts in embedded player loop

youle31:
NOT READY FOR REVIEW. Just to have an overview

About https://developer.blender.org/rBb47a234f98dd7e6e0b53ad96b5b4d278635874d8, there are indeed some asserts when restarting scene or loading other .blend in embedded.

I begun to try to use void BKE_blender_globals_main_replace(struct Main *bmain); inside the embedded loop instead of BLO_blendfiledata_free(bfd);

The idea, when we want to change bmain (at scene restart or blend load) is:
1. We store current wmWindowManager list :
ListBase wmbase;
wm_window_match_init(C, &wmbase);
2. We replace old main with the loaded main from the new .blend:
BKE_blender_globals_main_replace(bfd->main);
3. We use a piece of code to either adapt or replace new wmWindowManager list to new bmain with:
wm_window_match_do(C, &wmbase, &G_MAIN->wm, &G_MAIN->wm);

At exit, if we don't use bfd, we undo what has been done during runtime. Else, we load again original .blend in his"saved" state.


youle31:
It seems it is causing more issues than it solves. Embedded loop "works" as it is but with a new assert in debug mode caused by https://developer.blender.org/rBb47a234f98dd7e6e0b53ad96b5b4d278635874d8. I close this but keep the branch just in case.

PR #1771, opened 2022-12-19
UPBGE: Remove DerivedMesh usage in bge code

youle31:
It removed DerivedMesh code in game engine specific code.

It's using directly ob_eval->data to create/update what we need in bge (in BL_DataConversion to create the RAS_MeshObject, in CCD_PhysicsController to Set and Update the physics shape, in navmesh code to create the navmesh in viewport and the KX_NavMesh in bge).

I did not notice differences with or without derivedmesh.

It will make the merges more handy and if this PR is merged and that all works fine, we could remove some bge specific code related to DerivedMesh later.

Note 1: For SoftBodies update, it updates directly original mesh, which means objects with modifiers are not supported when using softbody (it's the same in master anyway)

Note 1 bis: If we'd like to have softbody acting as the last modifier in the stack, I guess we could call UpdateSoftBody after BKE_scene_graph_update_tagged and deform mesh eval vertices instead of orig mesh vertices.

Note 2: I converted navmesh.c to navmesh.cc to be abled to use new mesh API (mpoly = me->polys().data()......) which is only available in c++


lordloki:
I just tested all my physics, navmesh and softbody files and they worked very well.
Additionally, i tested high poly files and they worked like a charm.
I think you can commit with safe. If any problem arise we can fix it later.
Very good work.

youle31:
Thanks very much!! I merge than

PR #1774, opened 2022-12-31
Upbge0.2.5 - My [ / ] key does not work in the text editor.

juliorodriguesx:
I use the WINDOWS 10 operating system. However, when coding in python my [ / ] key does not work inside the UPBGE. Outside the program it works normally. If I copy a path from outside the UPBGE and paste it inside the python editor it works normally. How do I fix this bug? I can't write any path because of this.


youle31:
Hi, this kindof "issue" must be reported as issue here: https://github.com/UPBGE/upbge/issues with maximum details (OS, Hardware, country, keyboard disposition... But even if you report, there are chances than the issue won't be investigated as 0.2.5 development is stale.

Personally, I can't reproduce your issue on win 10 with a french keyboard. (Also, pathes on windows are generally like that: "C:\\Users\\john\\Desktop\\test.blend")

Anyway I close this as it is not a pull request, thanks for comprehension.

PR #1794, opened 2023-02-15
WIP: Add Ignore Depsgraph option

IzaZed:
This is an attempt at reducing the Depsgraph workload on static, idle objects that don't need to be updated.
It works best if there are only static/no collision objects in the scene, otherwise the depsgraph load is half of what we have now.
I still want to investigate if I can find a way to improve performance when having a physicsobject in the scene, so **please don't merge yet!**

image

It adds a new option to override the depsgraph entirely, tested with 1000 cubes:

image


BluePrintRandom:
can this be brought up to current and merged into master @youle31  ?

youle31:
> can this be brought up to current and merged into master @youle31 ?

On my side, as explained on discord, i don't think this patch will bring major benefits. This is tested with not moving object but scene has many objects and of if only one object is moving, the depsgraph cost will remain high.

youle31:
> can this be brought up to current and merged into master @youle31 ?

On my side, as explained on discord, i don't think this patch will bring major benefits. This is tested with not moving object but scene has many objects and of if only one object is moving, the depsgraph cost will remain high.

IzaZed:
I haven't had a look at this for quite a while as other things are more pressing right now
But I fear that youle's right, at least the current status of the branch is only useful when everything is static

What we'd need would be to ignore an object regardless of other updates in the scene, as currently the depsgraph is completely re-validated whenever something changes - _Then_ an option like this would really make sense

I'd like to investigate this further when I have the time and it's still an issue then, though, after 4.0 is out

Issue #1716, opened 2022-07-12
Converted blender object doesn't keep its parent/children relations


ThiagoMoraesGD:
As stated in this thread: https://blenderartists.org/t/upbge-0-3-append-external-blend/1389754 

Trying to append a rigged model from another file in game, using bpy.ops.wm.append and convertBlenderObjectsList() but after import the object to the actual scene, i lost all objects relations (parent/children)
[blend.zip](https://github.com/UPBGE/upbge/files/9093553/blend.zip)
.

youle31:
Hello, I did a test branch here:

https://github.com/UPBGE/upbge/commit/fc116a8462035347a9ccd164743e1eb965d36d0a named insta_col_runtime_test.

I only tested 1 case (test file at bottom -> it is to add objects with a rigid body constraint)

If you want you can test for your cases (children and track to) USING instance collections to see if it works properly.
As it's complex for me, i have difficulties to know what kindof bugs it could generate.

youle31:
I'll link test files here and edit when i test something:

- rigid body joint:

https://github.com/UPBGE/upbge/files/10789612/insta_col_runtime_test.zip

- parent-child with @ThiagoMoraesGD (current issue .blend modified to have it working in this branch):

[parent_child.zip](https://github.com/UPBGE/upbge/files/10794962/parent_child.zip)

- Track to (#1797):

[track_to.zip](https://github.com/UPBGE/upbge/files/10795047/track_to.zip)

And I share a test build:

https://ufile.io/1ijd73mf

(it is the build of the branch to test to convert instance collections during runtime).  (the ufile.io link expires in 1 month or something like that)

youle31:
Fixed here: https://github.com/UPBGE/upbge/commit/40fffb47d90d3c595919eaa61102ec04d55203eb

It sometimes needs some tweaks with python to have it working properly

ThiagoMoraesGD:
Hi, i test the new master but it seems that is not possible to iterate trough all objects..Do i have to convert one by one?

youle31:
you have to convert instance collection 1 by one but instance collection contains as many objects as you want.

The instance collection can be either converted to active list (scene.objects) or inactive list (scene.objectsInactive).
The instance collection itself in bge is just a gameobject but not visibled and with no physics.
Only instances created from instance collection are complete gameobjects with physics, logic...

maybe print scene.objects and scene.objectsInactive to see what is is active list and in inactive list, you will see what has been added inside.

[...]

PR #1798, opened 2023-02-22
UPBGE: Test converting instance collections at runtime

youle31:
It's to try to anwer this limitation: https://github.com/UPBGE/upbge/issues/1716

There are example files in the issue.

I'm a bit reluctant to do such things because it's complex and idk what bugs it can generate, but we can try anyway to include it in master to experiment I guess.

For now I only tested with the files included in the issue report, and I did not debug anything, I didn't even print scene.objects after conversion to see if something was wrong.

For now, I just put it in PR list if someone wants to test/review it, but there is no ermergency and if someone think it can lead to issues, we can abandon it.

I see 1 case where it could be problematic: When we add an instance collection with Full Duplication option checked (not tested).

EDIT:

- Tested in debug build and no problem noticed
- Tested with print(scene.objects) and no problem noticed
- Fixed issue with Full Duplication option in this commit: https://github.com/UPBGE/upbge/pull/1798/commits/6f49f09429595aa020e628659a1e9d2fb59cfd68
Test file for the Full Duplication bug before this commit:

[insta_col_spawn_duplica_rb_joint.zip](https://github.com/UPBGE/upbge/files/10810635/insta_col_spawn_duplica_rb_joint.zip)

EDIT2: The issue with full duplication option is that each time blenderdataconversion is done, it adds a new dupligroup/groupobj to object list and DupliGroupRecurse is called for ALL objects in objeclist/active gameobjects list.

I guess anyway blenderdataconversion was not meant to be called during runtime, then it was not really an issue.

For now, I think that this commit https://github.com/UPBGE/upbge/pull/1798/commits/6f49f09429595aa020e628659a1e9d2fb59cfd68 is enough but another way to fix this specific issue (if we have a new report one day), would just be to remove dupligroup objects from active gamobj list after DupligroupRecurse has been called (at the very end of blenderdataconversion)

SUB-EDIT: Finally, I tried to arrange this issue in these commits:
https://github.com/UPBGE/upbge/pull/1798/commits/282f292ba9bd8a390bd660b698c22489ae9186e0
https://github.com/UPBGE/upbge/pull/1798/commits/effc031ce4f1621448679c536576617b5297e614
https://github.com/UPBGE/upbge/pull/1798/commits/5764b048aab5c899104a8398429ce7477a888b75

but we can keep "safeguards" to avoid "Full Duplication" of an instance collection because idk in what case it could be useful.

EDIT3: Try adding instance collection several times after this commit https://github.com/UPBGE/upbge/pull/1798/commits/282f292ba9bd8a390bd660b698c22489ae9186e0

Test file for add insta col sevral times:
[insta_col_added_several_times.zip](https://github.com/UPBGE/upbge/files/10821088/insta_col_added_several_times.zip)


lordloki:
Thanks for merging with master. I was having issues reverting to former máster to compile this pull (they were issues with addons folder location due last blender commits). I will test this night. Great work

youle31:
yes, I had issues too. No pb, thanks in advance for your tests!

lordloki:
I just test all the files, the ones added in the pull and the ones in the github issue, and they work perfectly.
The code seems ok to me. Great great work as always.

lordloki (approved):

youle31:
Thanks!! I noticed a bit weird behaviours in some cases early this morning in some cases (instance collection is sometimes visibled whereas only the instances should be visibled), but it can be fixed with python or i'll change the code later after merge to always hide instances collections in active list (if INSTANCE_SPAWN).

Issue #1805, opened 2023-04-20
Assert at scene restart in embedded 0.3+ (debug builds windows)


youle31:
Test file:

[fps-test.zip](https://github.com/UPBGE/upbge/files/11278657/fps-test.zip)

Press F1 several times to restart the scene.

I'll check later.

EDIT: I forgot but actually, i already tried to fix that here: https://github.com/UPBGE/upbge/tree/embed_try

and as it was causing more issues than it solved, I gave up on this, concluding that we can live with this assert...

youle31:
Fixed by Blender developers in this merge https://github.com/UPBGE/upbge/commit/d326880e3671d5e4e0d322b7062c38f8a43c61be or in the previous days. Thanks very much to them!

PR #1848, opened 2023-10-01
Try to improve BL_KetsjiEmbedStart

youle31:
EDIT: Not stable

It is using WM_file_read to load a new .blend or when restart game is called instead of old "load_game_data".

We avoid to free pre-runtime wm->undo_stack during the whole runtime. At the end of runtime, WM_file_read is called again to load the pre-runtime file, and we assign the undo_stack that we saved to the wmWindowManager of the read file.
 
There is a remaining memory leak reported in console, i didn't find the cause.

It is avoiding this assert: https://github.com/UPBGE/upbge/issues/1805

I tested with few test files (I don't have many "restart-load" files) and it sounds to work ok.

Test files used:
[restart-load.zip](https://github.com/UPBGE/upbge/files/12777400/restart-load.zip)


youle31:
I close this for now

PR #1850, opened 2023-10-07
3denpy patch 1

3denpy:
This repository is only used as a mirror. Blender development happens on projects.blender.org.

To get started with contributing code, please see:
https://wiki.blender.org/wiki/Process/Contributing_Code


BluePrintRandom:
what is this @3denpy ?

PR #1875, opened 2024-02-14
add GetColorTexture and GetDepthTexture to KX_FilterFrameBuffer

IzaZed:
This feature allows `KX_FilterFrameBuffer` to provide a `GPUTexture` object when added as offscreen buffer. This texture can then be used in other shaders and the gpu module draw calls.

Usage:
```python
filter = bge.logic.getCurrentScene().filterManager.addFilter(0, 12, glsl)
filter.addOffScreen(0)
```
and then to get the texture (and use it as `uniform sampler custom_framebuffer` in glsl):
```python
texture_object = filter.offScreen.getColorTexture()
shader = GPUShader(...., ....)
shader.uniform_sampler('custom_framebuffer', texture_object)
```

Test File:
[Test04.zip](https://github.com/UPBGE/upbge/files/14284286/Test04.zip)

### DIFF
--- a/doc/python_api/rst/bge_types/bge.types.KX_2DFilterFrameBuffer.rst
+++ b/doc/python_api/rst/bge_types/bge.types.KX_2DFilterFrameBuffer.rst
@@ -36,3 +36,17 @@ base class --- :class:`~bge.types.EXP_Value`
       .. warning:: If the off screen can be resized dynamically (:data:`width` of :data:`height` equal to -1), the bind code may change.
 
       :type: integer
+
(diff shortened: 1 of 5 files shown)


youle31:
Hello, it sounds good, I didn't test your patch but i tested on my side and it was approximatively the same. There is just the doc missing; You can add few lines in bge.types.KX_2DFilterFrameBuffer.rst taking care about indentation/whitespaces and merge. Thanks!!

youle31 (approved):

PR #1885, opened 2024-04-29
Fix build problem due to missing metainfo.xml file

mysticfall:
Fix an invalid reference of a metainfo.xml file (changed by df563ed):

```
CMake Error at source/creator/cmake_install.cmake:270 (file):
  file INSTALL cannot find
  "/home/mysticfall/.cache/yay/upbge-git/src/upbge/release/freedesktop/org.blender.Blender.metainfo.xml":
  No such file or directory.
Call Stack (most recent call first):
  cmake_install.cmake:51 (include)


make: *** [Makefile:130: install] 오류 1
```

### DIFF
--- a/source/creator/CMakeLists.txt
+++ b/source/creator/CMakeLists.txt
@@ -810,7 +810,7 @@ file(REMOVE ${CMAKE_BINARY_DIR}/bin/lib/libglapi.so.0.0.0)\n
       DESTINATION "./share/applications"
     )
     install(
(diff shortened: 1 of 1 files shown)


lordloki:
Just to note that the new file is called org.upbge.UPBGE.metainfo.xml

mysticfall:
> Just to note that the new file is called org.upbge.UPBGE.metainfo.xml

Sorry, I was a bit in a hurry and should've been more careful! Will update the PR soon.

lordloki (approved):

mysticfall:
I'll merge it now. Thanks much! :)

PR #1887, opened 2024-05-03
Fix 1 frame delay when copying transformations using python

opheroth:
Fix 1 frame delay when copying transformations from one object to another using python in game engine.

### DIFF
--- a/source/gameengine/Ketsji/KX_KetsjiEngine.cpp
+++ b/source/gameengine/Ketsji/KX_KetsjiEngine.cpp
@@ -517,15 +517,21 @@ bool KX_KetsjiEngine::NextFrame()
       if (i == 0) {  // No need to UpdateObjectActivity several times
         scene->UpdateObjectActivity();
       }
-
-      m_logger.StartLog(tc_physics);
+      // I'm not sure if this goes here, I think this is for calculating
+      //time computing pysics, but physics are calculated in a lower section
+      //Since I don't know the code as well as I wish, if this is needed don't remove
+      //(Opheroth)
+      //m_logger.StartLog(tc_physics);
 
       // set Python hooks for each scene
       KX_SetActiveScene(scene);
 
+      /*Moved this down below to fix the 1 frame lag when copying transformations
+      from an object to another using python (Opheroth)
       // Process sensors, and controllers
       m_logger.StartLog(tc_logic);
       scene->LogicBeginFrame(m_frameTime, times.framestep);
+      -----------------------------------------------*/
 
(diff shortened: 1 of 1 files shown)


youle31:
Hello,

I wouldn't change the pipeline order for this, it sounds too risky...
Scripts can also need to be executed before UpdateAnimations (a script calling playAction for example).

Instead of changing upbge source code, i'd suggest to execute the script in a KX_Scene pre_draw callback, and this should solve the issue.

I saw that Musikai shared a file on blenderforartists forum:
[sync_test.zip](https://github.com/UPBGE/upbge/files/15205838/sync_test.zip)

This file can be adap
[sync_test2.zip](https://github.com/UPBGE/upbge/files/15205842/sync_test2.zip)

Parenthesis: I know that blender 2.79 behaved differently, but current way of working sounds ok to me, even if in this special case it needs to delay the script execution in a pre_draw callback.
ted like this:

opheroth:
I've tried calling pre_draw callbacks for animated armatures, but it didn't work, I've tried copying bone location but for some reason, the delay is still there, that's why I moved that part, and that fixed it, I haven't found problems in my projects that uses animated armatures. Is there a way to fix that armature problem without editing source code?

opheroth:
> Hello,
> 
> I wouldn't change the pipeline order for this, it sounds too risky... Scripts can also need to be executed before UpdateAnimations (a script calling playAction for example).
> 
> Instead of changing upbge source code, i'd suggest to execute the script in a KX_Scene pre_draw callback, and this should solve the issue.
> 
> I saw that Musikai shared a file on blenderforartists forum: [sync_test.zip](https://github.com/UPBGE/upbge/files/15205838/sync_test.zip)
> 
> This file can be adapted like this: [sync_test2.zip](https://github.com/UPBGE/upbge/files/15205842/sync_test2.zip)
> 
> Parenthesis: I know that blender 2.79 behaved differently, but current way of working sounds ok to me, even if in this special case it needs to delay the script execution in a pre_draw callback.

In this file, the problem is still there for the armature, the moving object is fine, but not the one copying the bone's tail position.

[CopyPosArm.zip](https://github.com/UPBGE/upbge/files/15206631/CopyPosArm.zip)

youle31:
There are other reasons imo why UpdateAnimations is in RenderCamera -> because it needs to wait that camera matrices (view and projection) are set.

About your file, it sounds to work like that:

[CopyPosArm2.zip](https://github.com/UPBGE/upbge/files/15210378/CopyPosArm2.zip)

opheroth:
Thanks for the help, and the explanation. It was very useful.

PR #1891, opened 2024-05-09
Fix wheels getting attached in wrong orientation in vehicle wrapper

opheroth:
I've been looking where the problem could be, and found an inverted rotation, probably the one mentioned here. If btRaycastVehicle.cpp is modified, this fix the orientation problem, and it seems to me it only affects the vehicles, please let me know if I'm wrong.


opheroth:
A flipped axis was not noticed until further tests, so this is not correct yet

PR #1892, opened 2024-05-09
Needed to correct orientation on wheels using vehicle wrapper

opheroth:
This and the KX_VehicleWrapper.cpp changes, fix the wrong orientation on wheels.
Please let me know if this has conflict with other modules.


opheroth:
A flipped axis was not noticed until further tests, so this is not correct yet

PR #1893, opened 2024-05-11
Vehicle wrapper wheels orientation fix

opheroth:
Orientation on wheels was mirrored due this sign. 

The attached blend was made by Musikai and stated the problem here:
https://blenderartists.org/t/vehicle-loose-wheels/1520846
[Vehicle_Loose_Wheels_279.zip](https://github.com/UPBGE/upbge/files/15284356/Vehicle_Loose_Wheels_279.zip)

### DIFF
--- a/extern/bullet2/src/BulletDynamics/Vehicle/btRaycastVehicle.cpp
+++ b/extern/bullet2/src/BulletDynamics/Vehicle/btRaycastVehicle.cpp
@@ -110,9 +110,9 @@ void btRaycastVehicle::updateWheelTransform(int wheelIndex, bool interpolatedTra
 	btMatrix3x3 rotatingMat(rotatingOrn);
 
 	btMatrix3x3 basis2;
-	basis2[0][m_indexRightAxis] = -right[0];
-	basis2[1][m_indexRightAxis] = -right[1];
-	basis2[2][m_indexRightAxis] = -right[2];
+	basis2[0][m_indexRightAxis] = right[0];
+	basis2[1][m_indexRightAxis] = right[1];
+	basis2[2][m_indexRightAxis] = right[2];
 
 	basis2[0][m_indexUpAxis] = up[0];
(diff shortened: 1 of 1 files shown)


youle31:
Hi, I tested both in 2.79b and in 0.42alpha.

- When it is possible, we try to avoid changing directly bullet library.
- I noticed that with your patch, when we press W to go forward in 0.42alpha, it goes in the opposite direction compared to blender 2.79b.

I propose the following changes instead:

[p.zip](https://github.com/UPBGE/upbge/files/15295538/p.zip)

Can you test my patch and tell me if it sounds ok for you? Thanks

opheroth:
Hi again, thanks for your reply. I've tried your changes proposal, but still there is a problem with mirrored axis, I've attached a file, also made by Musikai to explain the problem. 
You can see the image used in the cube, getting mirrored when playing the game, and that orientation error gives a wrong orientation compared with 2.79 when reading it inside the game.

In the change I've proposed, the orientation is now correct, but I forgot to modify in the KX_VehicleWrapper.cpp, the line 195:
195 force *= -1.f;
That line inverts the movement, removing it fix the problem of opposite direction.
Should I add the change?

To me it seems btRaycastVehicle.cpp is only related to vehicles using VehicleWrapper, but I'm really just beginning to understand the source code. Any feedback or help will be very appreciated. Thanks in advance.

[Vehicle_Loose_Wheels_0.42_test.zip](https://github.com/UPBGE/upbge/files/15302086/Vehicle_Loose_Wheels_0.42_test.zip)

youle31:
Ok, I will do the fix as you suggest directly in master, because there is an extra thing to do when we modify bullet code. Thanks very much for the fix and the tests!

youle31:
https://github.com/UPBGE/upbge/commit/6f2fca97d506ec6c686928d8540cb1e8ff1130d8

PR #1916, opened 2024-10-13
Fix build on Apple devices: `py_path_bundle` is not a C string

zakkor:
Build was failing due to compile errors on macOS. It was trying to call strchr on py_path_bundle, but py_path_bundle was not a C string


youle31:
Thanks!

But regular contributors have no mac M1 to test, then we can't fix other bugs that might occur with this platform/hardware

PR #1963, opened 2025-06-13
UPBGE: Add a modifier for softbodies

youle31:
It creates a simple basic OnlyDeform modifier.

This PR has several goals:

- Create an unique deformed shape for softbodies which have several mesh users
- Avoid deforming original mesh as it creates issues for example on scene restart
- Attempt to fix some transform issues (We still need to Apply Scale before bge start)

It doesn't solve all issues but I think it is a progress.
It has been tested with some of my files:

[Nouveau dossier.zip](https://github.com/user-attachments/files/20730340/Nouveau.dossier.zip)

### DIFF
--- a/source/blender/makesdna/DNA_modifier_defaults.h
+++ b/source/blender/makesdna/DNA_modifier_defaults.h
@@ -562,6 +562,11 @@
     .flag = 0, \
   }
 
+#define _DNA_DEFAULT_SimpleDeformModifierDataBGE \
+  { \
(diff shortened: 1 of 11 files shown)


lordloki (approved):
I just reviewed the code and it seems ok to me. Good work as always.

youle31:
Thanks very much for review! I will merge. Currently, we just need to always apply scale before bge start. I'll maybe have a look later at that.

PR #1974, opened 2025-07-25
UPBGE: Improve GLSL uniform parser and replace legacy built-in variables with UBO access

youle31:
This PR improves the GLSL uniform parser in RAS_Shader::GetParsedProgram with the following changes:
•	Ignores user-declared uniforms for bgl_RenderedTexture and bgl_DepthTexture to avoid redeclaration errors (these are now handled internally).
•	Handles uniform arrays and multiple variable declarations on a single line (e.g. uniform float a, b[4], c;).
•	For each user uniform, registers a push constant (except for built-in ones).
•	Replaces legacy built-in variables in shader code:
•	bgl_TextureCoordinateOffset[...] → g_data.coo_offset[...]
•	bgl_RenderedTextureWidth → g_data.width
•	bgl_RenderedTextureHeight → g_data.height
•	Comments out all user uniform declarations in the final shader code to prevent GLSL errors.
•	The parser is robust to extra spaces and tabs between keywords, types, and variable names.
Motivation:
•	Ensures compatibility with the new UBO-based system for built-in variables.
•	Prevents redeclaration errors and makes user shader code cleaner and more future-proof.
•	Improves user experience by allowing more flexible uniform declarations.
Testing:

What is above was generated with IA. I tested with few builtin shaders + 2 custom filters and it sounds ok. As the parser has been improved (with IA), old syntax works.

UBO is here to leave more space for custom uniforms because we are limited to 128 bytes for push_constant

### DIFF
--- a/doc/python_api/rst/bge_types/bge.types.KX_2DFilter.rst
+++ b/doc/python_api/rst/bge_types/bge.types.KX_2DFilter.rst
@@ -11,67 +11,86 @@ base class --- :class:`~bge.types.BL_Shader`
 
    .. note::
 
-     Following builtin uniforms are available:
+      Since version 0.5+, the following builtin uniforms are available via the UBO ``g_data`` (type ``bgl_Data``):
 
-       "bgl_RenderedTexture",        // RENDERED_TEXTURE_UNIFORM.
+         ``g_data.width``
+            Rendered texture width (float)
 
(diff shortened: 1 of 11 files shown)


youle31:
Thanks! I think it's ok. For ubo, I took example on other areas of Blender. For parser, the code is done with IA, I tested and saw no issue. If we see an issue later, i'll fix it.

PR #1980, opened 2025-08-10
Optimizations for GPU Skinning

lordloki:
Little skinning shader optimizations (around 10-15% improvement with my graphic card Nvidia 1660)

- Remove the need to calculate normal matrix = transpose(inverse(mat3(bone_pose_mat))) inside shader. Very cost as it
calculates it for each vertex&bone instead of each bone.

- combine pre/post matrices reducing memory consumption.

- Remove weight bifurcation as weight it is normalized previously out of shader. Also remove bifurcations if (w > 0.0) as it is better to do the multiplications with w = 0.0 instead of a bifurcation.

@youle31 could you test with your hardware? Perhaps the improvement will be minor since your hardware is better.


lordloki:
Nah, wrong matrix normal calculation.

PR #1982, opened 2025-08-13
UPBGE: GPU skinning initial implementation

youle31:
Add experimental GPU skinning support for armature deformation

This patch introduces GPU-accelerated skeletal animation (skinning) as an experimental feature for UPBGE, providing significant performance improvements for animated meshes with many vertices.

**Key Changes:**

- **New GPU skinning pipeline**: Implements compute shader-based vertex deformation that runs entirely on GPU
- **Dynamic VBO format switching**: Automatically switches between float3 and float4 position buffers based on skinning requirements
- **Modifier system integration**: Adds `use_gpu_deform` property to Armature modifiers with appropriate UI controls
- **Version compatibility**: Includes versioning code to ensure backward compatibility with existing files
- **Memory management**: Proper handling of GPU resources and mesh replication for instanced objects

**Technical Implementation:**

- Uses compute shaders to perform bone matrix transformations directly on vertex buffers
- Implements proper normal transformation using inverse transpose matrices
- Maintains compatibility with Blender's deformation pipeline for seamless switching between CPU/GPU modes
- Includes safety mechanisms to restore modifier stacks and mesh data when switching modes

**Limitations:**

- Currently supports only one Armature modifier per mesh
- Requires armature to have a single child mesh object
- Only updates position and normal vertex attributes
- Simplified deformation compared to full CPU pipeline

**File Format Changes:**

- Increments UPBGE file subversion to handle new modifier properties
- Adds versioning logic to set default CPU deformation mode for existing files

What's above was written with AI. I noticed that in last master, there were some crashes with vulkan backend using GPU skinning, but we can try to fix later. There will surely be some cases not well handled by current code, and potentially causing crashes.

EDIT: Vulkan crash happens since: ea83567811aaf66a80a039e0b7c3fa2d640a7290 it can be reverted for tests, waiting a fix is found
EDIT2: Normally it is fixed, i'll test again later to be sure


lordloki (approved):
The code seems ok to me. Anyway, we can fix any issue later.
I tested the code and it improves a lot the performance. Great job!!

youle31:
Thanks very much for review, tests, and support!

PR #1984, opened 2025-08-15
UPBGE: Add default UPBGE addons

lordloki:
We adopt the same system as in Blender to select the default UPBGE addons.

Currently, I have selected Logic nodes, Bricky nodes and Save as Runtime addons. 
We can discuss later if we add some additional ones.


youle31:
Thanks! It sounds to work well, and code sounds ok 💯

PR #1986, opened 2025-08-18
UPBGE: Add an option to use DupliObject system for fast KX_GameObject replication/transform

youle31:
This PR introduces a dedicated iterator for BGE dupli objects in the dependency graph, improving the separation between Blender’s internal duplilist and UPBGE’s runtime instances.
Key changes include:
•	New BGEObjectData structure for temporary BGE objects, managed via unique_ptr to ensure safe memory handling and avoid copy issues.
•	Dedicated BGE object provider API (BGEObjectProvider), allowing the game engine to inject dupli objects directly into the depsgraph iterator without callbacks or global state.
•	Integration in the depsgraph object iterator: BGE dupli objects are now collected and iterated separately, ensuring they do not interfere with Blender’s native duplication and rendering logic.
•	UI improvements: Added a new “UPBGE Dupli Base” property panel for game objects, making it clear when an object is used as a runtime dupli base in UPBGE.
•	Flag management: The new OB_DUPLI_UPBGE flag is used to mark objects as BGE dupli bases, and is respected throughout the conversion, transformation, and visibility code paths.
•	Code cleanup: Removed legacy callback and global vector usage, replaced with direct, thread-safe access to the iterator data.
Benefits:
•	Robust isolation of BGE dupli objects from Blender’s core systems (render, export, etc.).
•	Improved memory safety and maintainability.
•	Clear UI for UPBGE users, with minimal impact on Blender’s standard workflow.
Notes:
•	All changes are contained within UPBGE-specific code paths and do not affect standard Blender behavior.
•	The new system is thread-safe and ready for future extensions (e.g. full BGE transform control).

What is above is written by AI. It adds an option in Object UI to mark an Object as a Base to generate DupliObjects (with ProcessReplcia). This is to try to solve the addObject stutter bottleneck. These DupliObjects inherits their Transform directly from SceneGraph and do not depend on depsgraph (then they will be "unsynchonized" with depgraph.

We could have tried to use existing ob->duplilist, but it would have been more complex and potentially conflicting with blender ob->duplilist management. At first, some variables like bge_objects list in iterator were static but I changed it to make it directly integrated to DEGObjectIterData struct (makes it more thread safe).

With this system, duplication is fast, ans duplis transformation is fast (like for old KX_GameObjects in blender 2.79 or upbge 0.2.5).

If we want to go even more far later, we could add an option to transform all objects (existing + replicas) with the SceneGraph instead of depsgraph. But I think it's fine as a first step to try this approach, see if it works well, fix potential bugs.


lordloki (approved):
The code seems ok to me (although part it is out my knowledge). I tested adding several miles of bullets and it works ok. The performance is better and much more stable than doing the same with the current Blender depsgraph.

Great Job!!!!

PR #1989, opened 2025-09-01
UPBGE: Rework fh_spring

youle31:
This patch reworks the Fh spring logic in CcdPhysicsEnvironment: •	Forces are now applied along the contact normal for mesh shapes. •	Friction and anisotropic friction are only applied when the object is "considered in contact" with the ground (rough approximation: distance < margin), avoiding unwanted effects when airborne. This breaks compatibility with previous behavior, but the old logic was rarely used and produced weird results.


as5633123-bit:
[EpicGamesIntegration.md](https://github.com/user-attachments/files/22079029/EpicGamesIntegration.md)

youle31:
Thanks very much for review!

PR #1992, opened 2025-09-09
Added Fixed Physics Steps feature and Renamed the 'Use Frame Rate' to…

vuaieo:
… 'FPS Limit' because its the most common describtions in any game engine to understand what it do... and moved it to the Game Physics panel in Scene properties editor becuase it affects the physics fundamentaly when using Variable mode


youle31:
Hello, I quickly tested, it sounds  ~ok overhaul but there are several things to fix. I talked with gpt4 about my remarks about this patch and asked it to write a review comment and here it is:

1. External clock support
•	The Python API correctly exposes functions to enable the external clock (setUseExternalClock, setClockTime, etc.).
•	However, it is essential to ensure that the main engine loop (in KX_KetsjiEngine::NextFrame() and GetFrameTimes()) actually uses the external clock value and disables the application of timescale, tickrate, etc. when the external clock is active.
•	To check: the presence of a test on the USE_EXTERNAL_CLOCK flag to bypass internal time calculation.
2. Python API documentation
•	The new Python functions (e.g., setPhysicsTickRate, setUseFixedPhysicsTimestep, etc.) are not documented in bge.logic.rst.
•	It is essential to add Sphinx documentation for each new function, including: signature, description, parameters, return values, and any side effects.
3. Use of std::this_thread::sleep_for
•	Using std::this_thread::sleep_for(std::chrono::nanoseconds(...)) for timing (sleeptime) can cause stutters or imprecise timing depending on the platform (timer resolution, system latency).
•	For a game engine, it is recommended to use a hybrid approach (sleep + busy-wait) to ensure more precise timing, especially for very short sleep values.
4. Incorrect framerate display when changing physics steps
•	Setting physics steps to 30 (instead of 60) results in an abnormal framerate display (e.g., 100,000 fps).
•	This indicates a bug in time management or in the average framerate calculation (GetAverageFrameRate()), likely due to poor synchronization or division by a very small value.
•	The displayed framerate must be based on real elapsed time, and changing the physics tick rate should not disrupt the time accumulator or synchronization logic.
5. Robustness when changing tickrate
•	When changing the physics tick rate, the accumulator (m_physicsAccumulator) is correctly reset.
•	However, it is important to ensure that all time and state variables remain consistent to avoid artifacts when switching modes (fixed/variable timestep, changing tickrate on the fly).
---
Summary and recommendations
•	Fix external clock handling in the main loop if not already done.
•	Add Sphinx documentation for all new Python functions.
•	Review sleep management to avoid stutters, especially on Windows.
•	Fix framerate calculation so it remains consistent when changing physics steps/tickrate.

In my human vocabulary:

- m_timescale should be taken into account
- UseExternalClock should be takend into account
- python API is missing for the new functions in bge.logic.rst
- when we set physics steps to 30 instead of 60 in fixed mode, the displayed framerate by the bge profiler sounds messed -> display values like 500000 fps.
- std::sleep_for can cause stutters dépending on platforms, it's why I commented it. You can use it if you want but use it with "busy wait" -> ask gpt or the AI you use about that. I think it is the reason why when I enable fixed mode with use_fixed_fps_cap set to 60, the displayed FPS on my computer is 57/58.
- Please enable use_fixed_fps_cap by default (both for consistency with variable mode and energy savings)

PR #2042, opened 2026-08-11
Update rna_scene.cc to make FPS setting more understandable

Snicelory:
Before it was "Use Frame Rate" which is the exact name of the Blender "Use Frame Rate" setting. BUT this option actually uses the FPS set in the Game Physics panel. So this should be more helpful.


youle31:
Hello, Thanks for your contribution, there is just a missing space after panel.

```
panel"

instead of

panel "
```

It gives this in UI:

image

Else, I'm ok with this change, Thank you.

Snicelory:
Haha nice, thats my first contribution ever.
Rather small, but I think first time users will be helped with this.
Also yeah I bunked up the space, thanks for noticing that :D 

-Marcel

